PR #6, opened 2016-04-29
Use d2l-ajax to get auth header for request

d2l-joelp:


d2l-joelp:
@ryantmer @apalaniuk

ryantmer:
So much nicer. Can you update the demo to work with this? (Could just use `url=demo.json` like we had before, so the demo can work with or without auth)

bower.json
```diff
@@ -28,7 +28,8 @@
   ],
   "dependencies": {
     "polymer": "Polymer/polymer#^1.4.0",
-    "iron-ajax": "PolymerElements/iron-ajax#^1.2.0"
+    "iron-ajax": "PolymerElements/iron-ajax#^1.2.0",
+    "d2l-ajax-ui": "Brightspace/d2l-ajax-ui#^0.1.0"
```
ryantmer: Oh, the version in the repo right now is 1.0.0, are you intending to use an old one here?

d2l-joelp: I never intended for the first version to be 1.0.0 since what was initially in there wasn't really ready to be used. So I downgraded the version to 0.1.0 and tagged a commit with that version #. Not sure if that was the right thing to do or not.

ryantmer: Ah, gotcha. Seems right to me :)

PR #7, opened 2016-05-02
US70781 course tile styling

ryantmer:
Adds styling to the course tile as per Bob's designs. This doesn't include the layout of the tiles themselves, just within the tiles. The layout, including the number of columns/row of course tiles, and the sizes of said tiles, will have to be calculated within `d2l-my-courses`.


d2l-course-tile.html
```diff
@@ -1,39 +1,66 @@
...
+			background-position: center;
+			border-top-left-radius: 15px;
+			border-top-right-radius: 15px;
+		}
+		.course-text {
+			width: 100%;
+			height: 60px;
+			overflow: hidden;
+			position: relative;
+			font-family: 'Lato-Regular', sans-serif;
```
d2l-joelp: Can/should we inherit the font-family from the page? Or do we want to force these fonts in our widget?

d2l-joelp: @thehappypixel ? ^^

d2l-course-tile.html
```diff
@@ -1,39 +1,66 @@
...
-      font-family: "Lato-Light" sans-serif;
-      font-weight: 300;
-      font-size: 20px;
-      letter-spacing: 0.3px;
-    }
-    </style>
+	<template>
+		<style>
+		:host {
+			width: 270px;
```
d2l-joelp: I would comment on the hard-coded size of the element, but this will most definitely change when we implement the fancy column stuff. I guess this counts as commenting on it...

d2l-course-tile.html
ryantmer: Good question actually. The more I think about it, the more inheriting from the page makes sense, although doing it here does protect against something going wrong on the page...

d2l-course-tile.html
ryantmer: Yeah; this is just a placeholder size, really. The column count/sizing stuff will be a bigger challenge, so separate US/PR.

d2l-course-tile.html
d2l-joelp: Yeah, and my main concern is what is the intent from the designs. Was it, use what the page is using and the page happens to use Lato in Daylight, or is it, use Lato because that's what I want always.

d2l-joelp: Prob don't need to hold up this PR waiting for this answer. There's lots of styling to go yet. I say :shipit:

thehappypixel: Regarding the font. I think it's best that we use the font provided by the system. My understanding is that Lato should be set as part of VUI. There will be instances where you'll need to set the font weight though.

ryantmer: Cool; I'll inherit for the course tile at least in that case :)

PR #10, opened 2016-05-06
View pinned courses

d2l-joelp:
US69200 - Initial My Courses Widget - View Pinned Courses
https://rally1.rallydev.com/#/51191528503d/detail/userstory/53551933198

Note: If testing against an actual instance of UIS, will need set `FEATURE_PINNED_ENABLED=true`, `VALENCE_ENROLLMENTS_ROUTE_SEMVER=unstable`, and `FEATURE_VALENCE_URL_RESOLVER_ENABLED=true` as well as point your LMS at the UIS instance using the config variable `d2l.Tools.Homepages.UserInfoServiceUrl`.

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -18,7 +18,6 @@
 		.course-image {
 			width: 100%;
 			height: 60%;
-			background-image: url(http://lorempixel.com/640/480/cats/);
 			background-size: cover;
 			background-position: center;
 			border-top-left-radius: 10px;
@@ -55,7 +54,30 @@
 				},
 				name: {
 					type: String
+				},
+				imageSrc: {
+					type: String
+				},
+				imageCategories: {
+					type: Array,
+					readOnly: true,
+					value: function () {
+						return [
+							'abstract', 'animals', 'business', 'city', 'food', 'people', 'nature', 'sports', 'transport'
+						];
+					}
 				}
+			},
+			ready: function () {
+				var courseImage = this.$$('.course-image'),
+					courseImageSrc = this.imageSrc || this._getRandomImageUrl();
+
+				Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + courseImageSrc + ');');
+			},
+			_getRandomImageUrl: function () {
+				var index = Math.floor((Math.random() * 10) + 1); // 1 - 10 inclusive
+				var category = this.imageCategories[Math.floor((Math.random() * this.imageCategories.length))];
+				return 'http://lorempixel.com/640/480/' + category + '/' + index;
 			}
 		});
 	</script>
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -1,47 +1,9 @@
 <link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-course-tile-styles.html">
 
 <dom-module id="d2l-course-tile">
 	<template>
-		<style>
-		:host {
-			display: block;
-		}
-		a, a:hover, a:visited {
-			text-decoration: none;
-			color: inherit;
-		}
-		.course-image-container {
-			position: relative;
-		}
-		.course-image-container:before {
-			display: block;
-			content: "";
-			width: 100%;
-			padding-top: 43%; /* image aspect ratio */
-		}
-		.course-image {
-			position: absolute;
-			top: 0;
-			left: 0;
-			right: 0;
-			bottom: 0;
-			background-size: cover;
-			background-position: center;
-			border-top-left-radius: 10px;
-			border-top-right-radius: 10px;
-		}
-		.course-text {
-			width: 100%;
-			max-height: 75px;
-			overflow: hidden;
-			position: relative;
-			padding-top: 9px;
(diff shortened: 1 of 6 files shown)


d2l-course-tile.html
```diff
@@ -4,29 +4,35 @@
...
+			bottom: 0;
 			background-size: cover;
 			background-position: center;
 			border-top-left-radius: 10px;
 			border-top-right-radius: 10px;
 		}
 		.course-text {
 			width: 100%;
-			height: 68px;
+			max-height: 75px;
```
ryantmer: This being set to 75px was causing extra padding to show up between the course text and the bottom of the image, is that no longer the case?

d2l-joelp: I don't _think_ so because this is `max-height`, not height. But I can check.

ryantmer: I am bad at box model, so can't remember whether that will change this case or not :< Worth checking though, as @thehappypixel pointed it out previously

d2l-joelp: There is ~14px between the bottom of the image and the top of the text as measured by some Chrome extension I just installed.

ryantmer: Huh, alright then! I think 15px is the intended spacing, whereas mine was 24px for some reason... sounds good!

d2l-my-courses.html
```diff
@@ -42,12 +92,105 @@
 			},
 			enrollmentsUrl: {
 				type: String
+			},
+			pinnedCoursesUrl: {
+				type: String,
+				computed: '_getPinnedCoursesUrl(enrollmentsUrl)'
+			},
+			windowResizeHandler: {
```
ryantmer: Would we need to set this handler? (i.e. does it have to be a property?)

d2l-my-courses.html
```diff
@@ -42,12 +92,105 @@
...
 			}
+		},
+		_getPinnedCoursesUrl: function (enrollmentsUrl) {
+			return enrollmentsUrl + '?filter=pinned';
+		},
+		_calcNumColumns: function (node, itemCount) {
+			var width = node.getBoundingClientRect().width;
+			var numCols = 1;
+
+			if (width < 768) {
```
ryantmer: Shouldn't `width < 768` go down to one column? Or am I misunderstanding how this works... it could very well be that :P

d2l-my-courses.html
d2l-joelp: Hmm, yeah, probably doesn't need to be a property. The only purpose of it was to have a function that creates an event handler with `this` properly set to the web component, otherwise this will be the window object.

d2l-my-courses.html
d2l-joelp: Not always, no. As per [the mocks](https://zpl.io/1yLApS), if there's 4 or more items we should use 2 columns on mobile.

ryantmer: Oh! Am I crazy, or were they single-column before? :o

d2l-joelp: You might be thinking of the homepage layout itself which does go to a single column (100% width) at < 768px.

d2l-my-courses.html
```diff
@@ -42,12 +92,105 @@
 			},
 			enrollmentsUrl: {
 				type: String
+			},
+			pinnedCoursesUrl: {
```
ryantmer: I'm kinda on the fence about this one - I almost feel like this should not be a property, as it seems client-y. Thoughts?

d2l-my-courses.html
```diff
@@ -7,28 +7,78 @@
 
 		<style>
 		:host {
-			display: inline-block;
+			display: block;
```
ryantmer: Due to the addition of a fair amount of logic to this WC, can we move the style into a separate .css file or something?  Dunno if that's abnormal for WCs, granted.

ryantmer:
Code LGTM :+1: Didn't find any issues aside from the collapsed widget one that @thehappypixel got to before me :)

d2l-my-courses.html
d2l-joelp: So far I've only seen external style files when trying to share styles between components. But I don't think having them in a separate file would hurt anything since we're bundling everything up with shards anyway. Otherwise I would've been concerned with an unnecessary extra server call to get the style file.

d2l-my-courses.html
d2l-joelp: True, it is private, but I wanted to use it in a binding so I thought I needed to declare it as a property for that. I can try marking it as readOnly to be safe.

d2l-my-courses.html
d2l-joelp: Oh now I remember why I did this. I do want it to be a property because I want to keep the function around so that I can unsubscribe property in `detached`. Basically I need to keep the result of `.bind(this)` around because that creates a new function and I need that function when calling `removeEventListener`.

ryantmer: Ah, yes that makes sense. Didn't realize that, coolio :)

ryantmer:
Continues to LGTM with new commits! :shipit:

PR #12, opened 2016-05-12
Use ImageUrl from enrollments response for course image

ryantmer:
~~This is fine for demo purposes, but we will have to address how we should behave if the desired image does not exist.~~ As per @thehappypixel's design, a course without an image will show a Pressicus grey background instead (for this thin slice, at least).

~~Since the `course-image` comes back as a Siren link, so I'm passing the whole `links` of the enrollment into the course tile... might be nice to just pass the `course-image`'s href?~~ Just the href is now being passed to the course tile.

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -34,8 +34,6 @@
 				var courseImage = this.$$('.course-image');
 				if (newValue) {
 					Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + newValue + ');');
-				} else {
-					Polymer.dom(courseImage).setAttribute('style', 'background: #E57231');
 				}
 			}
 		});
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -26,28 +26,17 @@
 					type: String
 				},
 				imageSrc: {
-					type: String
-				},
-				imageCategories: {
-					type: Array,
-					readOnly: true,
-					value: function () {
-						return [
-							'abstract', 'animals', 'business', 'city', 'food', 'people', 'nature', 'sports', 'transport'
-						];
-					}
+					type: String,
+					observer: '_imageSrcChanged'
 				}
 			},
-			ready: function () {
-				var courseImage = this.$$('.course-image'),
-					courseImageSrc = this.imageSrc || this._getRandomImageUrl();
-
-				Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + courseImageSrc + ');');
-			},
-			_getRandomImageUrl: function () {
-				var index = Math.floor((Math.random() * 10) + 1); // 1 - 10 inclusive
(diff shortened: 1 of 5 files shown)


d2l-joelp:
I agree that it would be nicer to pass just the image URL to the course tile instead of all the links. The tile shouldn't need to know how to deal with that I don't think. Should be able to use a computed property or binding in `d2l-my-courses`.

demo/my-courses.json
```diff
@@ -150,6 +150,12 @@
             "self"
           ],
           "href": "http://localhost:3000/enrollments/123059"
+        },
+        {
+          "rel": [
+            "course-image"
+          ],
+          "href": "http://host-lms.com/d2l/api/lp/1.9/courses/123059/image"
```
d2l-joelp: So the demo will now have all blank tiles? I'd prefer that we point these images somewhere (lorempixel?) so that the demo page looks better and makes devving easier which is it's main purpose.

d2l-course-tile.html
```diff
@@ -26,28 +26,22 @@
...
+					courseImageUrl = "";
 
-				Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + courseImageSrc + ');');
-			},
-			_getRandomImageUrl: function () {
-				var index = Math.floor((Math.random() * 10) + 1); // 1 - 10 inclusive
-				var category = this.imageCategories[Math.floor((Math.random() * this.imageCategories.length))];
-				return 'http://lorempixel.com/640/480/' + category + '/' + index;
+				if (this.imageSrc) {
+					var courseImageSrc = this.imageSrc.find(function(item) {
```
d2l-joelp: `find` does not seem to have good browser support: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

d2l-course-tile.html
```diff
@@ -26,28 +26,22 @@
...
+						return item.rel.indexOf('course-image') > -1;
+					});
+
+					courseImageUrl = courseImageSrc ? courseImageSrc.href : "";
+				}
+
+				Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + courseImageUrl + ');');
```
d2l-joelp: If there is no `courseImageUrl` there's not really a point in setting the background-image to `url()`

demo/my-courses.json
ryantmer: Totally neglected our poor demo :( Truth. However, miiiight be getting default image behaviour from Bob soon, so if that happens then I'll update to use that.

d2l-course-tile.html
ryantmer: Oof, good call. Yeah I'll get rid of that.

d2l-joelp:
LGTM

PR #13, opened 2016-05-17
Fix course image scaling

ryantmer:
Two problems addressed here:

1. Due to the ordering of CSS, if a small image was uploaded to Nitro, the course tile image was not being stretched to fill the course tile
2. If a larger image is uploaded, Nitro was giving us back a smaller (640x480) image. Fix this by specifying that we want a 768px wide (our max course tile width) image, which we can then shrink as required.


d2l-joelp:
LGTM

PR #14, opened 2016-05-19
Add pinned state to course tile

ryantmer:
Includes some basic styling. Doesn't actually do anything thus far, just shows pin/unpin option as determined by the current pin state of the course.


d2l-course-tile-styles.html
```diff
@@ -40,6 +40,34 @@
...
+			visibility: hidden;
+			position: absolute;
+			top: 0;
+			left: 0;
+		}
+		.menu-item {
+			color: #FFFFFF;
+			background: #565A5C;
+			border-radius: 5px;
+			padding-top: 5px;
```
d2l-joelp: Use shorthand property `padding: 5px 10px`?

d2l-course-tile.html
```diff
@@ -28,13 +35,37 @@
...
 				if (newValue) {
 					Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + newValue + ');');
 				}
+			},
+			_pinnedChanged: function (pinned) {
+				var pinImage = this.$$('.menu-item img'),
+					pinText = this.$$('.menu-item span');
+
+				if (pinned) {
+					Polymer.dom(pinImage).setAttribute('src', 'unpin.png');
```
d2l-joelp: I'm wondering if instead of png images we should be using icons via `iron-icon` like in (navigation-ui)[https://github.com/Brightspace/d2l-navigation-ui]?

d2l-course-tile.html
```diff
@@ -28,13 +35,37 @@
...
 					observer: '_imageSrcChanged'
+				},
+				pinned: {
+					type: Boolean,
+					observer: '_pinnedChanged'
 				}
 			},
+			showHoverMenu: function() {
+				var hoverMenu = this.$$('.hover-menu');
+				Polymer.dom(hoverMenu).setAttribute('style', 'visibility: visible;')
```
d2l-joelp: Any reason why you're using `visibility` instead of `display:none`?

d2l-course-tile.html
d2l-joelp: And regarding the icons, I actually started using them in the upcoming all courses view. So when that gets merged adding a new icon will be easy peasy.

d2l-course-tile.html
ryantmer: tbh, not sure. But I guess `display:none` is more correct in this context, yeah.

d2l-course-tile.html
ryantmer: Nice. Should I wait for that, switch to `iron-icon`, or leave as-is for now do you think?

d2l-joelp: I would say let's leave it for now. Should be an easy change later.

d2l-course-tile.html
d2l-joelp: I generally only use visibility if I want the element to still take up space even when it's hidden, otherwise I hide all things using display:none. But I guess since you're using absolute positioning it doesn't currently make a difference.

ryantmer:
Cleanup up and reviewed with Bob briefly, so should be good to go for a first pass!

d2l-joelp:
LGTM

PR #15, opened 2016-05-20
View all courses

apalaniuk:
@d2l-joelp, @ryantmer 

There are still a couple of issues, but I don't want them to delay other testing/feedback any longer than I already have while I look at them, especially as one of them relates to a Polymer behavior:

* Keyboard navigation in the All Courses view when not using Shadow DOM seems to be problematic, preventing iron-overlay-behavior wraparound logic from working. 
* Translating isn't implemented as the library/element to use hasn't been completely decided yet. That's intentional and will be revisited.

### DIFF
--- a/d2l-all-courses-styles.html
+++ b/d2l-all-courses-styles.html
@@ -2,7 +2,6 @@
 <link rel="import" href="d2l-course-tile-region-styles.html">
 
 <dom-module id="d2l-all-courses-styles">
-
 	<template>
 		<style include="vui-typography"></style>
 		<style include="d2l-course-tile-region-styles"></style>
--- a/d2l-all-courses-styles.html
+++ b/d2l-all-courses-styles.html
@@ -17,10 +17,9 @@
 
 		h3.header {
 			@apply(--vui-heading-3);
-			margin-top: 0px;
+			margin-top: 10px;
 			margin-bottom: 20px;
 		}
-		
 		</style>
 	</template>
 </dom-module>
--- a/d2l-all-courses-styles.html
+++ b/d2l-all-courses-styles.html
@@ -14,15 +14,11 @@
 			letter-spacing: 1px;
 			color: var(--vui-color-ferrite);
 		}
-		
+
 		h3.header {
 			@apply(--vui-heading-3);
 			margin-top: 0px;
 			margin-bottom: 20px;
-		}
-		
-		#all-courses-overlay {
-			z-index: 1000;
 		}
 		
 		</style>
--- a/d2l-all-courses-styles.html
+++ b/d2l-all-courses-styles.html
@@ -1,0 +1,30 @@
+<link rel="import" href="../vui-typography/typography.html">
+<link rel="import" href="d2l-course-tile-region-styles.html">
+
+<dom-module id="d2l-all-courses-styles">
+
+	<template>
+		<style include="vui-typography"></style>
+		<style include="d2l-course-tile-region-styles"></style>
+		<style>
+		:host {
+			@apply(--vui-font);
+			display: block;
+			font-family: 'Lato-Regular', sans-serif;
+			letter-spacing: 1px;
+			color: var(--vui-color-ferrite);
+		}
+		
+		h3.header {
+			@apply(--vui-heading-3);
+			margin-top: 0px;
+			margin-bottom: 20px;
+		}
+		
+		#all-courses-overlay {
+			z-index: 1000;
+		}
+		
+		</style>
+	</template>
+</dom-module>
--- a/d2l-all-courses-styles.html
+++ b/d2l-all-courses-styles.html
@@ -16,8 +16,7 @@
 
 		h3.header {
 			@apply(--vui-heading-3);
-			margin-top: 10px;
-			margin-bottom: 20px;
+			margin: 10px 0;
 		}
 		</style>
 	</template>
--- a/d2l-all-courses-view.html
+++ b/d2l-all-courses-view.html
@@ -2,7 +2,7 @@
 <link rel="import" href="d2l-all-courses-styles.html">
 <link rel="import" href="d2l-simple-overlay.html">
 <link rel="import" href="d2l-course-tile.html">
-<link rel="import" href="../d2l-ajax-ui/d2l-ajax.html">
+<link rel="import" href="../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="../vui-typography/typography.html">
 <link rel="import" href="d2l-formatjs-behavior.html">
 <link rel="import" href="d2l-lang-terms.html">
--- a/d2l-all-courses-view.html
+++ b/d2l-all-courses-view.html
@@ -4,207 +4,232 @@
 <link rel="import" href="d2l-course-tile.html">
 <link rel="import" href="../d2l-ajax-ui/d2l-ajax.html">
 <link rel="import" href="../vui-typography/typography.html">
+<link rel="import" href="d2l-formatjs.html">
+<link rel="import" href="d2l-lang-terms.html">
 
 <dom-module id="d2l-all-courses-view" class="vui-typography">
-        <template>
-		<style include="d2l-all-courses-styles"></style>
-		
-		<d2l-ajax
-			id="allEnrollmentsRequest"
-			on-response="allEnrollmentsOnResponse"
-			url="[[enrollmentsUrl]]"
-			headers='{ "Accept": "application/vnd.siren+json" }'>
-		</d2l-ajax>
+    <template>
+        <style include="d2l-all-courses-styles"></style>
+        <d2l-lang-terms></d2l-lang-terms>
 
-		<d2l-simple-overlay
-			id='all-courses-overlay'
-			title="All Courses">
-			
-			<div class="dialog-content">
-				<h3 class="header">Pinned<h3>
-				<div class="my-courses-container">
-					<template is="dom-repeat" items="[[pinnedCoursesEntities]]">
-					<d2l-course-tile
-							target="/d2l/home/[[item.properties.id]]"
-							name="[[item.properties.name]]"></d2l-course-tile>
-					</template>
-				</div>
-				<h3 class="header">Unpinned</h3>
-				<div class="my-courses-container">
-					<template is="dom-repeat" items="[[unpinnedCoursesEntities]]">
-					<d2l-course-tile
-							target="/d2l/home/[[item.properties.id]]"
-							name="[[item.properties.name]]"></d2l-course-tile>
-					</template>
-				</div>
-			</div>
-		</d2l-simple-overlay>
-	</template>
-	
-	<script>
-	Polymer({
-		is: 'd2l-all-courses-view',
-		properties: {
-			pinnedCoursesEntities: {
-				type: Array
-			},
-			unpinnedCoursesEntities: {
-				type: Array,
-				readOnly: true
-			},
-			allEnrollmentsEntities: {
-				type: Array,
-				readOnly: true
-			},
-			enrollmentsUrl: {
-				type: String
-			},
-			windowResizeHandler: {
-				type: Object,
-				readOnly: true,
-				value: function () {
-					return this._handleWindowResize.bind(this);
-				}
-			}
-		},
-		allEnrollmentsOnResponse: function (response) {
-			if (response.detail.status === 200) {
-				this._setAllEnrollmentsEntities(response.detail.xhr.response.entities);
-				this._filterUnpinnedCourses(this.pinnedCoursesEntities, this.allEnrollmentsEntities);
-				this._rescaleCourseTileRegions();
-			}
-		},
-		open: function() {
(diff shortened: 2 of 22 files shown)


ryantmer:
I think my merge gave you some conflicts, you'll wanna rebase :(

d2l-all-courses-styles.html
```diff
@@ -0,0 +1,24 @@
...
+	<template>
+		<style include="vui-typography"></style>
+		<style include="d2l-course-tile-region-styles"></style>
+		<style>
+		:host {
+			@apply(--vui-font);
+			display: block;
+			font-family: 'Lato-Regular', sans-serif;
+			letter-spacing: 1px;
+			color: var(--vui-color-ferrite);
```
ryantmer: Nice, so we can apply VUI styles now? Nice.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
+<link rel="import" href="../polymer/polymer.html">
```
d2l-joelp: nitpick: Probably don't need to have '-view' in this file name.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+<link rel="import" href="d2l-all-courses-styles.html">
+<link rel="import" href="d2l-simple-overlay.html">
+<link rel="import" href="d2l-course-tile.html">
+<link rel="import" href="../d2l-ajax-ui/d2l-ajax.html">
```
d2l-joelp: If you rebase, this reference will change to `../d2l-ajax/d2l-ajax.html`

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+
+		<d2l-simple-overlay
+			id='all-courses-overlay'
+			title={{allCourses}}
+			with-backdrop
+			restore-focus-on-close>
+
+			<div class="dialog-content">
+				<h3 class="header">{{pinnedCourses}}</h3>
+				<div class="my-courses-container" role="region" aria-label="Pinned Courses">
```
d2l-joelp: Should localize the `aria-label`.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+					<template is="dom-repeat" items="[[pinnedCoursesEntities]]">
+						<d2l-course-tile
+							target="/d2l/home/[[item.properties.id]]"
```
d2l-joelp: Since we're now specifying the url to the course in multiple places, maybe we should move it into the course tile itself and just pass in the course id? Unless we want to make the course tile ignorant of where it points to.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+							name="[[item.properties.name]]">
+						</d2l-course-tile>
+					</template>
+				</div>
+				<h3 class="header">{{unpinnedCourses}}</h3>
+				<div class="my-courses-container" role="region" aria-label="Unpinned Courses">
```
d2l-joelp: Localize `aria-label`

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+					readOnly: true
+				},
+				allEnrollmentsEntities: {
+					type: Array,
+					readOnly: true
+				},
+				enrollmentsUrl: {
+					type: String
+				},
+				windowResizeHandler: {
```
d2l-joelp: Should probably implement this window resize handling as a behavior now that it's being used in two spots as there's a fair amount of code that was copied.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+					value: function () {
+						return this._handleWindowResize.bind(this);
+					}
+				}
+			},
+			observers: [
+				'_updateData(messages, locale)'
+			],
+			behaviors: [
+				Polymer.d2l.formatJSBehavior
```
d2l-joelp: Let's chat about how we're doing localization :)

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+					this._rescaleCourseTileRegions();
+					this.$.allEnrollmentsRequest.generateRequest();
+					this.$$('d2l-simple-overlay').open();
+			},
+			_isCourseUnpinned: function(pinnedCoursesEntities) {
+				return function(element) {
+					var isUnpinned = true;
+
+					pinnedCoursesEntities.forEach(function(pinnedCourse) {
+						if(element.properties.id == pinnedCourse.properties.id) {
```
d2l-joelp: You can't exit a `forEach` early, but if you use a regular loop you can. Then, if you find the element you're looking for you should be able to just `return false` immediately and `return true` at the end if it wasn't found.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+				this._getTranslations();
+			},
+			_getTranslations: function() {
+				if(this.messages) {
+					this.allCourses = this.format(this.getCultureTerm('allCourses'), null);
+					this.pinnedCourses = this.format(this.getCultureTerm('pinnedCourses'), null);
+					this.unpinnedCourses = this.format(this.getCultureTerm('unpinnedCourses'), null);
+				}
+			},
+			focusableNodesOverride: function() {
```
d2l-joelp: I will admit, I don't understand what this focusable nodes stuff is for :)

d2l-simple-overlay.html
```diff
@@ -0,0 +1,136 @@
...
+			this.close();
+		},
+		_isMobile: function () {
+			var width = Math.max(document.documentElement.clientWidth, window.innerWidth || 0);
+			return width < 768;
+		},
+		_updateBackdropStyle: function () {
+			var style = document.createElement('style', 'custom-style');
+			style.innerHTML = "iron-overlay-backdrop { --iron-overlay-backdrop-opacity: 0.0; }";
+			document.body.appendChild(style);
```
d2l-joelp: I think you should be able to style this normally using this style in `d2l-simple-overlay-styles`
```css
iron-overlay-backdrop {
  --iron-overlay-backdrop-opacity: 0;
}
```

apalaniuk: I did try this at first, but because the backdrop is applied to the body element, it didn't override the default value. I'll take another look to see if I can it working in a better way again.

d2l-joelp: Sure. It just looked from the 'docs' like that should be possible: https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-backdrop.html#L19. I could be wrong.

d2l-all-courses-styles.html
```diff
@@ -0,0 +1,24 @@
+<link rel="import" href="../vui-typography/typography.html">
```
ryantmer: I'm thinking this file shouldn't really be required - since the all courses view should have the same styling as the pinned-only view... or maybe we should just look at combining them into one file? Outside of this PR, either way. I'll make an issue for it.

d2l-all-courses-view.html
ryantmer: (mentioned earlier, but to make it all official-like) as I'm adding the API calls to the course tile, it looks like it might actually be cleaner to pass the whole enrollment Entity. So, I'll be updating this anyway, so should be fine for now.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+							isUnpinned = false;
+							return;
+						}
+					});
+
+					return isUnpinned;
+				}
+			},
+			_filterUnpinnedCourses: function(pinnedCoursesEntities, allCoursesEntities) {
+				if(pinnedCoursesEntities && allCoursesEntities) {
```
ryantmer: Don't actually need to do a comparison with the pinned courses here; the entities have a `pinDate` property, so you can just check for that to know if they're pinned or unpinned.

d2l-all-courses-view.html
```diff
@@ -0,0 +1,253 @@
...
+			},
+			focusableNodesOverride: function() {
+				var courseTileFocusables = [];
+
+				this.getContentChildren().forEach(function(contentChild) {
+					var courseTiles = Polymer.dom(contentChild).querySelectorAll('d2l-course-tile');
+
+					[].forEach.call(courseTiles, function(courseTile) {
+						Array.prototype.push.apply(courseTileFocusables, Polymer.dom(courseTile.root).querySelectorAll('a'));
+					})
```
ryantmer: Missing semicolon - weird that the linter didn't get that :confused:

d2l-course-tile-styles.html
```diff
@@ -1,22 +1,38 @@
...
+		:focus .course-text {
+			text-decoration: underline;
+		}
+
 		a, a:hover, a:visited {
 			text-decoration: none;
 			color: inherit;
 		}
+
+		a:focus .course-text {
```
ryantmer: If you move the `:focus .course-text` block below the `a, a:hover, a:visited` block, shouldn't need this I don't think?

d2l-course-tile-styles.html
```diff
@@ -25,20 +41,23 @@
 			bottom: 0;
 			border-top-left-radius: 10px;
 			border-top-right-radius: 10px;
-			background: #B9C2D0;
+			background: var(--vui-color-pressicus);
```
ryantmer: :+1:

d2l-formatjs-behavior.html
```diff
@@ -0,0 +1,45 @@
+<link rel="import" href="../polymer/polymer.html">
+<!--<link rel="import" href="../formatjs-element/formatjs-behavior.html">-->
```
ryantmer: Is this intentionally excluded? If so, can it be removed?

d2l-formatjs-behavior.html
```diff
@@ -0,0 +1,45 @@
...
+			locale: {
+				type: String,
+				value: 'en-US'
+			}
+		},
+		format: function(msg, values, language) {
+			language = language || this.locale;
+
+			// @TODO: Implement translation, once a suitable library/element is determined/created.
+			//return Polymer.formatJSBehavior.translate(msg, values, language);
```
ryantmer: For possible-future-reference, I think this should be `Polymer.d2l`, ja?

d2l-all-courses-view.html
apalaniuk: iron-overlay-behavior tries to select focusable elements automatically (so that keyboard focus wraparound restricted to dialog elements alone works.. or is supposed to at least..), but its internal selector isn't suitable for how we're rendering course tiles, so we need to override it so it knows what the first and last children are.

I'm also not 100% sure I understand it..

d2l-joelp: Makes enough sense for me :)

d2l-formatjs-behavior.html
apalaniuk: @ryantmer: Polymer.formatJSBehavior is from [formatjs-element](https://github.com/felixzapata/formatjs-element), but I should remove any traces of implementation if it's effectively unimplemented.

d2l-simple-overlay.html
apalaniuk: I tried this about every way imaginable, including by styling :host and :root from a custom Polymer element that was also child of the body element, but it doesn't seem possible to do it from a web component. I will make it a bit more idiomatic though.

d2l-course-tile.html
```diff
@@ -9,6 +9,7 @@
 			<div
 				class="course-image-container">
 				<div class="course-image"></div>
+				<div class="course-image"></div>
```
d2l-joelp: Looks like this got duplicated. Maybe a rebase issue?

d2l-course-tile.html
```diff
@@ -60,7 +64,7 @@
 				if (newValue) {
 					Polymer.dom(courseImage).setAttribute('style', 'background-image: url(' + newValue + ');');
 				}
-			},
+			},
```
d2l-joelp: nit: extraneous whitespace

d2l-course-tile-styles.html
```diff
@@ -1,22 +1,34 @@
...
 		}
+
+		:focus .course-text {
+			text-decoration: underline;
+		}
+
 		.course-image-container {
 			position: relative;
 		}
+
```
ryantmer: the smallest of nits: extra line breaks

d2l-course-tile.html
```diff
@@ -47,6 +47,9 @@
 					observer: '_pinnedChanged'
 				}
 			},
+			ready: function () {
+				this.setAttribute('role', 'link');
```
ryantmer: I r confus by this - why does this have to wait until `ready` to be set? :confetti_ball:

ryantmer:
This LGTM :+1: - any further thoughts? @d2l-joelp @apalaniuk

apalaniuk:
@ryantmer @d2l-joelp 

Anything additional I'll put into an additional PR.

PR #17, opened 2016-05-24
Mouseover image saturation

ryantmer:
Applies a filter to course image on mouseover, to give us the colorization effect. This required switching to an `` rather than using `background-image`, which is something we wanted to do anyway.

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -8,27 +8,17 @@
 			display: block;
 			@apply(--vui-font);
 		}
-
 		a, a:hover, a:visited {
 			text-decoration: none;
 			color: inherit;
 		}
-
 		:focus .course-text {
 			text-decoration: underline;
 		}
-
 		.course-image-container {
 			position: relative;
+			padding-top: 43%;
 		}
-
-		.course-image-container:before {
-			display: block;
-			content: "";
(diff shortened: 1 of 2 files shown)


ryantmer:
@apalaniuk @d2l-joelp Just a little thing I had mostly done on Friday (but forgot to push (and then found bugs in (and then fixed today))).

d2l-course-tile-styles.html
```diff
@@ -38,8 +38,27 @@
...
 			border-top-right-radius: 10px;
 			background: #B9C2D0;
-			background-position: center;
-			background-size: cover;
+			overflow: hidden;
+			z-index: 0;
+		}
+		.course-image img {
+			width: 100%;
+			position: absolute;
```
d2l-joelp: Do we need `position:absolute` and top, bottom, etc. on the course-image div itself in addition to the same styles on the `img`?

ryantmer: Turns out they are required, but the `:before` stuff can mostly go. I've pushed a commit, but GH is currently experiencing some issues.

ryantmer: Yay, GH has caught up now.

d2l-joelp:
LGTM

PR #19, opened 2016-05-25
Localization using Polymer.AppLocalizeBehavior

d2l-joelp:
[US72169] Use AppLocalizeBehavior in web component
https://rally1.rallydev.com/#/51191528503d/detail/userstory/56362137143

Also change the `Polymer.d2l` namespace a bit.


d2l-joelp:
@apalaniuk @ryantmer ?

d2l-simple-overlay.html
```diff
@@ -8,12 +8,13 @@
...
 <link rel="import" href="d2l-simple-overlay-styles.html">
 <link rel="import" href="icons.html">
+<link rel="import" href="localize-behavior.html">
 
 <dom-module id='d2l-simple-overlay'>
 	<style include="d2l-simple-overlay-styles"></style>
 	<template>
 		<div class="max-width">
-			<button class="close-button pull-right" on-tap="handleClose" alt=[[closeSimpleOverlayAltText]]>
+			<button class="close-button pull-right" on-tap="handleClose" alt$="{{localize('closeSimpleOverlayAltText')}}">
```
ryantmer: What's the `$=` used for? Haven't seen that thus far

d2l-joelp: It's a Polymer [native attribute binding](https://www.polymer-project.org/1.0/docs/devguide/data-binding#native-binding)

ryantmer:
:+1:

localize-behavior.html
```diff
@@ -0,0 +1,58 @@
...
+				type: String,
+				computed: '_determineLanguage(locale, resources)'
+			},
+			locale: {
+				type: String,
+				value: 'en-US'
+			},
+			resources: {
+				value: function () {
+					var en = {
```
apalaniuk: Do we care about making this more generic, and having the consumer/more specific behaviour set/override resources? Or is that not important for now?

d2l-joelp: Longer term yeah we'll probably want to have a behavior that other projects can use that can re-use the logic around choosing a language if the one you asked for doesn't exist. But for this project or for a single project at a time I would want all the strings defined for the entire app so we don't have terms spread around in a whole bunch of files. Thoughts?

apalaniuk: I don't have much of a problem with it since it will be simple enough to generalize when necessary, the impact on this project will be low, there may need to be changes when the next team wants to use it anyway, etc.

apalaniuk:
:+1:

localize-behavior.html
d2l-joelp: Yeah that's a good point. We should make sure it works well for us first before spreading this other teams.

localize-behavior.html
```diff
@@ -15,7 +15,11 @@
 			},
 			locale: {
 				type: String,
-				value: 'en-US'
+				value: function () {
+					return document.documentElement.lang
```
apalaniuk: Didn't realize this was set there. Neat.

PR #21, opened 2016-05-26
US69204 pinning API call

ryantmer:
In order for the course tile to do the API call in a cleaner way, I reworked the tile so that the widget just passes in the entire enrollment Entity. Changing the enrollment (i.e. setting it) triggers the observer to parse the enrollment and update the pin button as required. Clicking the pin/unpin button triggers the API request, and updates the pin button as required. The API reply also updates the pin button as required.

Disclaimer: haven't tested this in the LMS yet. Once I overcome my SQL woes, I'll also take a look at amping up the demo so that all of these features will work in that by mocking the token/API requests.


d2l-all-courses.html
```diff
@@ -1,9 +1,12 @@
...
+<link rel="import" href="../d2l-siren-parser/d2l-siren-parser.html">
+<link rel="import" href="../d2l-ajax/d2l-ajax.html">
+<link rel="import" href="../vui-typography/typography.html">
 <link rel="import" href="d2l-all-courses-styles.html">
 <link rel="import" href="d2l-simple-overlay.html">
 <link rel="import" href="d2l-course-tile.html">
-<link rel="import" href="../d2l-ajax/d2l-ajax.html">
-<link rel="import" href="../vui-typography/typography.html">
+<link rel="import" href="d2l-formatjs-behavior.html">
+<link rel="import" href="d2l-lang-terms.html">
```
d2l-joelp: We are not using `d2l-formatjs-behavior` or `d2l-lang-terms` anymore (as of this morning). Did you rebase?

d2l-all-courses.html
```diff
@@ -98,14 +91,13 @@ <h3 class="header">{{localize('unpinnedCourses')}}</h3>
...
 					this._setUnpinnedCoursesEntities(
 						allCoursesEntities.filter(
 							function(element) {
-								var isUnpinned = this.getPinnedState(element.entities) !== true;
-								return isUnpinned;
-							}.bind(this)
-						)
-					);
+								var entity = parser.parse(element);
+								return !entity.hasEntityByClass('pinned');
```
d2l-joelp: So clean!

d2l-course-tile.html
```diff
@@ -1,86 +1,143 @@
...
-		<a href="[[target]]" on-mouseover="showHoverMenu" on-mouseout="hideHoverMenu">
+		<d2l-ajax
+			id="update-preferences"
+			url="[[_updatePreferencesUrl]]"
+			method="[[_updatePreferencesMethod]]"
+			body="[[_updatePreferencesBody]]"
+			headers='{ "Accept": "application/vnd.siren+json" }'
+			on-response="updatePreferencesOnResponse"></d2l-ajax>
+
+		<a href="" on-mouseover="hoverCourseTile" on-mouseout="unhoverCourseTile">
```
d2l-joelp: Does the course tile not navigate anymore? Curious about `href=""`.

d2l-course-tile.html
```diff
@@ -1,86 +1,143 @@
...
+			},
+			_enrollmentEntityChanged: function (entity) {
+				if (entity.hasLink('course-image')) {
+					var courseImageLink = entity.getLinkByRel('course-image');
+					var courseImage = this.$$('.course-image img');
+					Polymer.dom(courseImage).setAttribute('src', courseImageLink.href);
 				}
+
+				var anchor = this.$$('a');
+				Polymer.dom(anchor).setAttribute('href', '/d2l/home/' + entity.properties.id);
```
d2l-joelp: Oh, I see. nvm my comment above. But it feels like this can still be done with binding, no?

d2l-course-tile.html
```diff
@@ -1,86 +1,143 @@
...
 			},
 			ready: function () {
 				this.setAttribute('role', 'link');
 			},
-			showHoverMenu: function() {
+			updatePreferencesOnResponse: function (response) {
+				if (response.detail.status === 200) {
+					var parser = document.createElement('d2l-siren-parser');
+					var preferencesEntity = parser.parse(response.detail.xhr.response);
+					this._pinned = preferencesEntity.properties.hasOwnProperty('pinDate');
```
d2l-joelp: Do we really want to wait until the call comes back to update the pin state? To make the UI feel more responsive I feel like we should assume it worked but if it fails we can revert the change to pin state. Thoughts?

d2l-all-courses.html
ryantmer: I did, although I cleaned up the order here a bit, so it lost it it seems. I shall remove!

d2l-course-tile.html
ryantmer: Yeah, definitely a good case for binding I'd say. I only included the `href=""` since strictly speaking an anchor requires an `href`, pretty sure. But yeah, binding is mo' betta.

d2l-course-tile.html
ryantmer: In the `pinClickHandler` I change the value of `this._pinned` immediately, which calls `_pinnedChanged` and does update the UI immediately. In theory, this line here should have no effect, since `this._pinned` should already be set to the correct value.

d2l-course-tile.html
```diff
@@ -1,86 +1,143 @@
...
+				}
+			},
+			pinClickHandler: function(e) {
+				// Prevent the click from triggering navigation to the course
+				e.preventDefault();
+
+				this._pinned = !this._pinned;
+
+				var updatePreferences = this.$$('#update-preferences');
```
ryantmer: Aw. Yeah probably. Forgot we can't really use the advantages of Shadow DOM just yet :<

d2l-course-tile.html
d2l-joelp: My bad. Missed that.

ryantmer: That said, I realize now that if the value comes back as a non-200, I'm not setting the icon to go back to the previous one. I'll add that in.

d2l-course-tile.html
```diff
@@ -1,86 +1,143 @@
...
+
+				var updatePreferencesBody = {};
+				this._updatePreferencesFields.forEach(function(field) {
```
d2l-joelp: For the future, this type of stuff might be useful to put in `d2l-siren-parser`. And by 'this type of stuff' I mean things to help you build a correct request when executing actions.

ryantmer: Yeah, this came up a while ago as well. I think it would be handy to have a module that does it, but I think it would be good to have it separate from the parser itself, since it's pretty simple and network requests are less so. But something that complements it, fo sho.

d2l-joelp: Sorry, I just meant to help create the body, not actually send the request. Seems like every time you'd want to call an action you'll loop through the fields to add them all to the body and set the appropriate default values. How you end up calling the action is app specific for sure.

ryantmer: Ah! I gotcha. Although much simpler, I still feel that that would be more at home in a `node-siren-request` or something, as the parser is more for deconstructing rather than constructing.

d2l-joelp: Good point. I'm with ya.

ryantmer:
Unfortunately the pin/unpin icon still has somewhat wonky spacing, but this is caused by the SVG itself rather than any sort of padding/margin, so I'll have to check with @thehappypixel on that once he gets back.

PR #22, opened 2016-05-26
Change animation for desktop overlay open/close

d2l-joelp:
- Use scale, translateY, and fade instead of slide from top
- Introduce `forceWidth` param to responsive container calculations so that we can use a width from a specific container. This was needed because the scale transform affects the width.

Also updated the `d2l-ajax` reference since it was accidentally pinned to v1.0.1

### DIFF
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -2,9 +2,10 @@
 <link rel="import" href="../iron-overlay-behavior/iron-overlay-behavior.html">
 <link rel="import" href="../neon-animation/neon-animation-runner-behavior.html">
 <link rel="import" href="../neon-animation/animations/slide-from-bottom-animation.html">
-<link rel="import" href="../neon-animation/animations/slide-from-top-animation.html">
 <link rel="import" href="../neon-animation/animations/slide-down-animation.html">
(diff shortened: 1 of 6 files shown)


d2l-simple-overlay.html
```diff
@@ -2,9 +2,10 @@
 <link rel="import" href="../iron-overlay-behavior/iron-overlay-behavior.html">
 <link rel="import" href="../neon-animation/neon-animation-runner-behavior.html">
 <link rel="import" href="../neon-animation/animations/slide-from-bottom-animation.html">
-<link rel="import" href="../neon-animation/animations/slide-from-top-animation.html">
 <link rel="import" href="../neon-animation/animations/slide-down-animation.html">
-<link rel="import" href="../neon-animation/animations/slide-up-animation.html">
+<link rel="import" href="../neon-animation/animations/transform-animation.html">
```
ryantmer: These `neon` elements are nice af

ryantmer:
LGTM - tried this out on desktop and mobile and it appears to act as it should. Not sure if you want to address the `width` problem here? Seems like a bug that should maybe have its own PR.

d2l-joelp:
@ryantmer As discussed, I fixed that width issue in this PR as the fix affected the overlay layout too.

PR #23, opened 2016-05-27
Fix issue in IE where widget would not load

d2l-joelp:
Use iron-ajax `lastResponse` instead of `xhr.response` because lastResponse will be properly parsed according to `handleAs`.

Also needed to bump the `d2l-ajax` version to get a required bug fix.


ryantmer:
:+1:

PR #24, opened 2016-05-30
Improve a11y

d2l-joelp:
[US69235] [UI] Verify Accessibility
https://rally1.rallydev.com/#/51191528503d/detail/userstory/53570115313

- Course tile should read name of course only
- Pin/unpin button should include name of course to provide context
- After pin/unpin should announce action result

### DIFF
--- a/localize-behavior.html
+++ b/localize-behavior.html
@@ -31,10 +31,10 @@
 						'closeSimpleOverlayAltText' : 'Close Dialog',
 						'pin': 'Pin',
 						'pinCourse': 'Pin {course}',
-						'pinActionResult': 'Your course has been pinned',
+						'pinActionResult': '{course} has been pinned',
 						'unpin': 'Unpin',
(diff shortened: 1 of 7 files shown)


d2l-joelp:
@ryantmer @apalaniuk any comments would be most welcome.

ryantmer:
Singular comment supplied! LGTM.

localize-behavior.html
```diff
@@ -30,7 +30,11 @@
 						'unpinnedCourses' : 'Unpinned',
 						'closeSimpleOverlayAltText' : 'Close Dialog',
 						'pin': 'Pin',
-						'unpin': 'Unpin'
+						'pinCourse': 'Pin {course}',
+						'pinActionResult': 'Your course has been pinned',
```
d2l-joelp: Yeah, that's probably better. I was just going by what @thehappypixel put in the user story. I'll change it to this and ask him about it when he gets back.

PR #26, opened 2016-06-01
Mobile pinning interaction

apalaniuk:


d2l-course-tile.html
```diff
@@ -79,13 +91,31 @@
...
...zhAX4OEzyNeQqz3Jh/IuN+xz9cRbwG8vTKssk0FDthPIMN+xKbKTLp4GgheAfQPqlBcFjhA24vt3kU8VZVA1woE3kPP0wD/T+EKOkqBA/YGXLP6dE/ZJPVNJXAcI3YDeuFFHAXggG1vFLTXD9wxNa0GVBH4Dhe3AfrEmwgnvvgAtp0fQgh2FUrfNWN4KGOaz5gAjt4DiJ35VR20JWAsjWne8pKCO3Enek8g8r58lA7aEFjF5LtQWr62JKHD7U4ItlGJ24xp7vKLgOHu+9H5RVwdZbuEwANw+dLY4S9cclGn8QlcZDyCu+1ru309XRdfQxkDAmdxvN4cvgMh2AGZGR0a4x0G/OEZCShtkcBOA75Y7FfPjAhsNuBzM0qutEUCcwJehDLLnjn7lPI4FAx8Ks1K3MFIi3kZeU5FyhUjTXqvx4QBtiJita0xKMXMkZKzY+4r2VwCTr41Ai7gZAJkOTlcwMkEyHJyuICTCZDl5HABJxMgy8nhAk4mQJaTwwWcTIAsJ4cLOJkAWU4OF3AyAbKcHC7gZAJkOTlcwMkEyHJyuICTCZDl5HABJxMgy8nhAk4mQJaTwwWcTIAsJ4cLOJkAWU4OF3AyAbKcHC7gZAJkOTlcwMkEyHJyuICTCZDl5HABJxMgy8nhAk4mQJaTwwV8zUkw+DcCB8thuZJqyTk8ezjukQiUjmS5IqSKlyI54NnWPoywxRg5IixjMkWSjz7F02luwDJ/RaydXK777BxG3gSH/+GeQRqYpMMzUEdbMDiaImzbT5LAM9BtSkKbuS3u8/SpSZYUWzbKir1N5RdE3XcN2StebobD7fGuybVkHZ4BW2pAbClV2LaXZIFnoJuUhiZzGtzXHk5BWdmAOI/wNhtrb2hJtiXtcJSGMyC3XIPecjanxhTu0KSBZyjqlIg6Y7mku6KGEjGH8JQVG6MHF8e4sQC5jJjW6pSeGMtqlKMLJcU25ikVnjGNIA1uEqy9EXGhwuJ2bePgwMxywwB6vAK4vWyuE60rJcVgVpWMqmuduBHJLRLunq9w+HxyC+7DggB8pQT6Spf21qWSYlzLSkdZX5fuQbprhbsXSxy+mO6Ke7AyAD8RQLfXUXaqda2kGNywhITHnQLfmcXC3VsCh2/pzMK7vFAA/8aii3uo+/+Fqeyxs6VEwMkW+h/eVVjVW58VpQAAAABJRU5ErkJggg=="
```
ryantmer: Can we use the same pin/unpin `iron-icon`s that the course tile is using for this? (I realize they're currently incorrect, but that should be fixed quite soon)

d2l-course-tile.html
```diff
@@ -47,6 +49,16 @@
 				</button>
 			</div>
 		</div>
+
+		<d2l-touch-menu>
+			<d2l-touch-menu-item
+				id="pin-touch-menu-item"
```
ryantmer: I believe we're avoided using IDs since we're not actually using Shadow DOM

(@d2l-joelp correct me if I'm wrong?)

d2l-course-tile.html
```diff
@@ -179,6 +211,24 @@
...
+			},
+			_updateTouchMenu: function() {
+				if(this._touchMenu && this._enrollmentEntity) {
+					var pinTouchMenuItem = this._touchMenu.querySelector('#pin-touch-menu-item');
+
+					if(this._pinned) {
+						Polymer.dom(pinTouchMenuItem).setAttribute('text', this.localize('unpin'));
+						Polymer.dom(pinTouchMenuItem).setAttribute('background-image', this._unpinBackgroundImage);
+						Polymer.dom(pinTouchMenuItem).setAttribute('action-description', this.localize('unpinCourse', 'course', this._enrollmentEntity.properties.name));
+					}
```
ryantmer: nit: `else` on the same line

d2l-course-tile-styles.html
```diff
@@ -7,7 +7,10 @@
 		:host {
 			display: block;
 			@apply(--vui-font);
+			transform: scale(1.0);
+			-webkit-backface-visibility: hidden;
 		}
+
```
ryantmer: :bike: :house: : extra line break

d2l-course-tile.html
```diff
@@ -79,13 +91,31 @@
...
+					type: Object
 				}
 			},
 			behaviors: [
 				Polymer.D2L.MyCourses.LocalizeBehavior
 			],
 			ready: function () {
+				this._touchMenu = this.$$('d2l-touch-menu');
+			},
+			attached: function () {
```
ryantmer: I'll admit, I don't understand this - can we not `setAttribute` and update the menu in `ready`? Why do we have to wait until it's actually attached to the DOM? :confused:

d2l-course-tile.html
```diff
@@ -179,6 +211,24 @@
 					Polymer.dom(pinText).setAttribute('style', 'display: inline');
 					Polymer.dom(unpinText).setAttribute('style', 'display: none');
 				}
+
+				this._updateTouchMenu();
+			},
+			_updateTouchMenu: function() {
```
ryantmer: Hmm... can this functionality be merged into `_pinnedChanged`? I realize there is the extra check at the beginning, but do we need that since this will only ever get called after `ready`, at which point `_touchMenu` and `_enrollmentEntity` should be non-null?

d2l-course-tile.html
apalaniuk: I'll check to see brace-style is set in eslint correctly, since I depend on that to get me out of some of my old style habits.

ryantmer: (as per Slack discussion) it is not, and it turns out we're not extending the Brightspace eslint settings. Added #27 to address this.

d2l-course-tile.html
apalaniuk: @ryantmer: That's fine with me.

d2l-course-tile.html
apalaniuk: @ryantmer: This was definitely something waiting for Bob confirmation - since he confirmed this morning that it's the same icon, I'm updating it to align with your hover button.

d2l-touch-menu-item-styles.html
```diff
@@ -0,0 +1,83 @@
...
+			-webkit-backface-visibility: hidden;
+		}
+
+		.longpress-menu-item {
+			text-align: center;
+			position: relative;
+		}
+
+		.longpress-menu-item-hover div {
+			background-color: #006FBF;
```
ryantmer: Since these are Daylight colours and such, can you do the `import vui-typography` magic here as well?

d2l-course-tile.html
apalaniuk: @ryantmer: It previously was part of _pinnedChanged, and broken out for reasons that I think no longer make sense - moving it back in tests correctly. So, will update.

d2l-course-tile.html
d2l-joelp: @ryantmer Yes, that's correct. You end up with duplicate IDs and the LE will complain and pop up a dialog telling you about it, which is annoying.

d2l-touch-menu.html
```diff
@@ -0,0 +1,279 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-touch-menu-styles.html">
+
+<dom-module id="d2l-touch-menu">
+    <template>
+        <style include="d2l-touch-menu-styles"></style>
+
+        <div id="long-press-menu">
```
ryantmer: Another ID

d2l-course-tile.html
```diff
@@ -179,6 +211,24 @@
...
+				if(this._touchMenu && this._enrollmentEntity) {
+					var pinTouchMenuItem = this._touchMenu.querySelector('#pin-touch-menu-item');
```
d2l-joelp: I don't think you need to keep a reference to `_touchMenu` around. You should be able to use the polymer helpers to find your menu item just fine (i.e. `this.$$('#pin-touch-menu-item')` should work.)

d2l-touch-menu.html
```diff
@@ -0,0 +1,279 @@
...
+
+                    // We need to catch contextmenu events both directly from this element, and separetly from the touch target,
+                    // as each will be the target of that event under different circumstances (typically, this element will be the target
+                    // when there is no active scroll event, and the touch target will be the event target when there is a scroll event)
+                    this.listen(this.touchTarget, 'contextmenu', 'contextMenuHandler');
+
+                    // Bind touch menu to menu items
+                    this._menuItems.forEach(function(menuItem) {
+                        menuItem._touchMenu = this;
+                    }.bind(this), false);
```
ryantmer: Is this intended to be `forEach(function() {}, this)`? Haven't ever seen a bool as an arg for `foreach` :deciduous_tree:

d2l-course-tile.html
ryantmer: Nice. Yeah it's defined in `icons.html` already and all; once I've got the new ones from Bob I'll update them and all will be magical :D

d2l-touch-menu.html
apalaniuk: forEach, yes - bool no. Already removed in the revision (after trying to divine any possible meaning for that arg).

d2l-course-tile.html
ryantmer: Can `position: absolute` not be used to display it above other elements rather than reparenting? (Having only tried this with jQuery, I am unfamiliar with the limitations that Polymer imposes on things like that =/)

apalaniuk: I don't think it's Polymer specific - I could always certainly use `position: absolute` to display it outside of the d2l-my-courses widget, but not eg. above d2l-my-courses' siblings that have at least the same z-index (eg. the widget following d2l-my-courses).

d2l-touch-menu.html
```diff
@@ -0,0 +1,279 @@
...
+
+                clearTimeout(this._timerID);
+
+                this._menuItems.forEach(function(menuItem) {
+                    menuItem.visible = false;
+                });
+
+                //If no item was selected, we can reparent immediately
+                if(!this._itemSelected) {
+                    Polymer.dom(this._owner).appendChild(this);
```
ryantmer: In the jQuery implementation of this, I believe I just created and attached the menu once, then just hid it when the user cancelled (or after the animation, if it was selected). When a new touch event occurs, move the menu to be there, then re-display it. I think taking this approach can remove the need to reparent/`appendChild` things, which should reduce a fair amount of complexity in the menu and menu items.

That said, not sure how to get that to play nice with Polymer, as I'm uncertain of the limitations it has in this regard.

d2l-touch-menu.html
```diff
@@ -0,0 +1,279 @@
...
+                }
+
+                this.touchcancelHandler(e);
+            },
+            touchcancelHandler: function () {
+                this._touchStartEventData = null;
+
+                clearTimeout(this._timerID);
+
+                this._menuItems.forEach(function(menuItem) {
```
ryantmer: Rather than the `forEach`, can you just `visibility: inherit` on the menu items, and change the visibility on the entire menu?

d2l-course-tile.html
ryantmer: Hm, fair point. That said, I think if you give it a 1000 z-index or whatever, you can probably safely assume that other widgets won't display on top of it. (I should hope that... oh man, I should hope that none of the widgets have some ridiculously high z-index, as that would be... not nice.)

d2l-touch-menu.html
apalaniuk: I did run into issues at the onset when porting to Polymer that made it much easier to allow it to be declared within the element rather than declared at a higher level, but I can't say that's wasn't due to issues I could solve now being more familiar with it. I'll take another look at that.

d2l-touch-menu-item.html
```diff
@@ -0,0 +1,209 @@
...
+
+        <div class="longpress-menu-item">
+            <div class="longpress-menu-item-text">{{_displayText}}</div>
+
+            <div class="longpress-menu-item-image-background">
+                <img class="longpress-menu-item-image" src$="{{resolveUrl(_displayBackgroundImage)}}">
+            </div>
+        </div>
+
+        <div class="d2l-offscreen" aria-live="assertive">{{ariaMessage}}</div>
```
d2l-joelp: I _think_ you should try using `polite` and see if that works because assertive will always interrupt and I think that can be annoying for people. Also, we're Canadian so it just seems more appropriate ;)

d2l-touch-menu.html
```diff
@@ -0,0 +1,279 @@
...
+                        menuItem.visible = true;
+                        menuItem.offset = menuItemOffset;
+                        menuItem.touchRegion = {
+                            left: left + menuItemOffset.x,
+                            top: top + menuItemOffset.y
+                        };
+                    }.bind(this));
+
+                    this.visible = true;
+                    this.fire('touch-menu-opened', {element: this}, {node: this._owner});
```
d2l-joelp: I can't find any uses of this event. Also, when trying to pass data along with an event I believe you should put it all in the same object (ie. `{ element: this, node: this._owner }`). Then you can access the data in the event handler using `e.detail`.

test/d2l-touch-menu-item/d2l-touch-menu-stub.html
```diff
@@ -0,0 +1,135 @@
+<link rel="import" href="../../../polymer/polymer.html">
+
+<dom-module id="d2l-touch-menu-stub">
```
d2l-joelp: What's the purpose of this element? I don't see it used anywhere.

d2l-touch-menu.html
apalaniuk: The third parameter is an optional event target, and is only required if the intended listener isn't a parent (eg. because the element has been reparented..), so may not stay if not-reparenting works =)

d2l-touch-menu.html
apalaniuk: It's no problem - if aligning with your PoC makes more sense and is doable in a reasonable amount of time, that should be done (and I don't think it would be terribly difficult to do). 

After thinking about it, the only question that I have is about the best paradigm. Eg, the desktop hover menu comes "for free" with a course tile, whereas the touch menu wouldn't. I would think that has more to do in your case because positioning within the element is a factor in your case, whereas that's not the case with a hover menu. The only other situation I can think of is if you don't want an entire custom element to be the target of a longpress.

I'm looking at aligning it more with your PoC now. Thanks!

d2l-touch-menu.html
d2l-joelp: OK, cool, didn't know that. But is anything listening for this event? Seems we could probably just remove it.

apalaniuk: Not at the moment - fancy mouseover styling, as there's a class applied to the touch target on menu item hover, but it would be simple enough to re-add at that time.

d2l-touch-menu-item.html
apalaniuk: I did have issues with using 'polite' and the announcement being delayed in cases where there was nothing else being announced for a significant period of time, then announced once something else gained focus (mainly seen in the All Courses view). I will take another look at this, though.

d2l-touch-menu.html
ryantmer: I _believe_ `visibility: hidden` takes care of the issue of the hidden menu being the target of the longpress; or, more accurately, I don't remember coding anything to specifically get around that from happening, as a hidden element I don't think responds to the touch event (or it just propagates it up to the parent, which in this case would be the course tile).

apalaniuk: Not the hidden menu, but an element within a WC itself. However, that's likely making it too generalized for situations that might not arise.

d2l-touch-menu.html
```diff
@@ -0,0 +1,234 @@
...
+                    this._menuItems.forEach(function(menuItem) {
+                        menuItem.hoverState = menuItem.pointInTouchRegion(touchPoint);
+                    }.bind(this));
+                } else {
+                    this._cancelMenu();
+                }
+            },
+            touchendHandler: function (e) {
+                // When finger is released, if the item is selected, then do the fancy animation
+                this._menuItems.forEach(function(menuItem) {
```
ryantmer: Minor, but if you switch this to a regular `for` you can break when you find the correct menu item.

d2l-touch-menu.html
```diff
@@ -0,0 +1,234 @@
...
+                this.listen(this.touchTarget, 'touchend', 'touchendHandler');
+
+                var touchData = {
+                    startX: touchstartEvent.changedTouches[0].pageX,
+                    startY: touchstartEvent.changedTouches[0].pageY,
+                    clientX: touchstartEvent.changedTouches[0].clientX,
+                    clientY: touchstartEvent.changedTouches[0].clientY
+                };
+
+                // Initial position of menu - if near the top of screen, show it below the initial tap
```
ryantmer: There is some weird behaviour here, but can probably be addressed in another PR when we look at deciding where to spawn the menu - right now, if it spawns below the finger, it moves toward the finger when you go to select it.

icons.html
```diff
@@ -8,26 +8,4 @@
 			</g>
 		</defs>
 	</svg>
-	<svg>
```
ryantmer: No more pin icons? :(

ryantmer:
LGTM I'd say - any remaining things we can address as they come up :)

PR #28, opened 2016-06-03
Empty states

d2l-joelp:
[US71383] - [UI] Empty State for Courses
https://rally1.rallydev.com/#/51191528503d/detail/userstory/55417104504

### DIFF
--- a/d2l-alert-styles.html
+++ b/d2l-alert-styles.html
@@ -1,0 +1,36 @@
+<link rel="import" href="../vui-typography/typography.html">
+<link rel="import" href="../vui-colors/colors.html">
+
+<dom-module id="d2l-alert-styles">
+	<template>
+		<style include="vui-typography"></style>
+		<style>
+		:host {
+			@apply(--vui-font);
+			display: block;
+			font-family: 'Lato-Regular', sans-serif;
+			font-size: 0.7em;
+		}
+		.message-wrapper {
+			position: relative;
+			border: 1px solid var(--vui-color-gypsum);
+			border-radius: 3px;
+			padding: 16px 16px 16px 20px;
+		}
+		.message-highlight {
+			position: absolute;
+			top: 0;
+			left: 0;
+			bottom: 0;
+			width: 6px;
+			background-color: #1598cb;
+			border-top-left-radius: 3px;
+			border-bottom-left-radius: 3px;
+			border-top-right-radius: 0;
+			border-bottom-right-radius: 0;
+			margin: -1px;
+		}
+		/* TODO: make this work for LTR */
+		</style>
+	</template>
+</dom-module>
--- a/d2l-alert-styles.html
+++ b/d2l-alert-styles.html
@@ -30,7 +30,14 @@
 			border-bottom-right-radius: 0;
 			margin: -1px;
 		}
-		/* TODO: make this work for LTR */
+		:host-context([dir='rtl']) .message-highlight {
+			left: auto;
+			right: 0;
+			border-top-left-radius: 0;
+			border-bottom-left-radius: 0;
+			border-top-right-radius: 3px;
+			border-bottom-right-radius: 3px;
+		}
 		</style>
 	</template>
 </dom-module>
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -1,0 +1,33 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-alert-styles.html">
+
+<dom-module id="d2l-alert">
+	<template>
+		<style include="d2l-alert-styles"></style>
+
+		<div class="message-wrapper" hidden$="{{_hideAlert}}">
+			<div class="message-highlight"></div>
+			<content></content>
+		</div>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-alert',
+			properties: {
+				visible: {
+					type: Object,
+					value: function () {
+						return false;
+					}
+				},
+				_hideAlert: {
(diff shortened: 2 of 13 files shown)


d2l-alert-styles.html
```diff
@@ -0,0 +1,43 @@
...
+			border-radius: 3px;
+			padding: 16px 16px 16px 20px;
+		}
+		.message-highlight {
+			position: absolute;
+			top: 0;
+			left: 0;
+			bottom: 0;
+			width: 6px;
+			background-color: #1598cb;
```
ryantmer: Is this a Daylight colour? It seems to be somewhere between Zircon and Azurite .-.

d2l-alert.html
```diff
@@ -0,0 +1,33 @@
...
+			<content></content>
+		</div>
+	</template>
+	<script>
+		Polymer({
+			is: 'd2l-alert',
+			properties: {
+				visible: {
+					type: Object,
+					value: function () {
```
ryantmer: Kinda confused by this one - you've got this as an `Object` here, but when you're instantiating a `d2l-alert` you're giving it a computed `Boolean`? =/

d2l-alert-styles.html
d2l-joelp: It did not appear to be daylight color, at least not what was in Zeplin. @thehappypixel, thoughts?

d2l-alert.html
d2l-joelp: I meant for consumers to be able pass in anything 'truthy' and have it work which is why I chose Object. The reason for the computed boolean is that attributes like `hidden` need a boolean value for them to work, otherwise basically everything results in 'true'.

ryantmer: Ah, gotcha. Yep, makes sense!

d2l-my-courses.html
```diff
@@ -11,27 +14,47 @@
...
+			id="pinnedCoursesRequest"
 			url="[[pinnedCoursesUrl]]"
 			headers='{ "Accept": "application/vnd.siren+json" }'
-			on-response="enrollmentsOnResponse"
-			last-response="{{enrollmentsResponse}}">
+			on-response="onPinnedCoursesResponse"
+			last-response="{{pinnedCoursesResponse}}">
 		</d2l-ajax>
 
+		<d2l-ajax
```
ryantmer: Can the request be removed from `d2l-all-courses` now, in that case?

d2l-alert-styles.html
d2l-joelp: Just chatted with Bob and and he said this is the colour that was provided to him and though it is not currently a daylight colour, we will stick with it until told otherwise.

d2l-my-courses.html
d2l-joelp: Probably, though I left it in to handle the case where you stay on the page for a long time without opening All Courses, then when you open it the course list might be stale. It's a corner case for sure but I was going for safety over performance. Thoughts?

d2l-my-courses.html
```diff
@@ -97,6 +164,9 @@
 			_ariaAnnounce: function (message) {
 				this._setAriaMessage('');
 				this._setAriaMessage(message);
+			},
+			_showAlertMessage: function (message) {
```
ryantmer: This appears to be unused, leftover from prior stuff maybe?

d2l-my-courses.html
ryantmer: Makes sense to me! Although, do we need to pre-fetch in that case? Can we just leave the all courses request until the point when the dialog is opened? Not sure I see the advantage of pre-fetching.

d2l-my-courses.html
d2l-joelp: Good catch. If only I had Resharper ;)

d2l-my-courses.html
ryantmer: Nvm. Need to do it to know which message to display. Ignore me :stuck_out_tongue:

d2l-my-courses.html
```diff
@@ -40,11 +63,25 @@
 		Polymer({
 			is: 'd2l-my-courses',
 			properties: {
+				delayLoad: {
```
ryantmer: This is being checked, but not set - is this just for future use?

ryantmer: Sigh... ignore me again... I ctrl+f'd up.

d2l-joelp: It's really just meant for helping us in the unit tests (which were giving me lots of trouble). I generally hate adding code just for testing purposes, but theoretically this could be used by consumers as well.

ryantmer: Yeah, it's reasonable either way I'd say. Plus, testable code is good code!

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -1,30 +1,76 @@
 /* global describe, it, beforeEach, afterEach, fixture, expect, sinon */
 
 describe('smoke test', function() {
-	var ajaxElement,
-		ajaxElements,
-		server,
+	var server,
 		widget,
-		enrollmentsResponse = {
+		emptyResponse = {
```
ryantmer: nit: stick these in a `d2l-my-courses.data.js` or something?

ryantmer:
Few minor things, but LGTM otherwise :ship: :it:

test/d2l-my-courses/d2l-my-courses.js
d2l-joelp: Agreed, but perhaps in a follow on PR

PR #30, opened 2016-06-03
Hover menu style updates

ryantmer:
- Change from colour swap to just increasing the opacity of the background
- Decrease amount of contrast/saturation on hover


ryantmer:
(Going to merge, since this is itty bitty and is just a couple style updates.)

PR #31, opened 2016-06-03
Improve transition smoothness / other styling

ryantmer:
Turns out, using a translateZ instead of a scale kicks the transformation off to the GPU, resulting in a much smoother transition. Neat!

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -82,20 +82,21 @@
 			margin-top: 10px;
 		}
 		.menu-item {
-			color: #FFFFFF;
+			color: rgba(255,255,255,1);
 			background: rgba(0,0,0,0.5);
 			border: none;
 			border-radius: 5px;
 			padding: 10px 15px;
 			float: left;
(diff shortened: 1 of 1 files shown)


ryantmer:
@thehappypixel If you want to give this a whirl, let me know and I'll set up my local install to be on this branch :)

omsmith:
For your choppiness, consider using [will-change](https://developer.mozilla.org/en/docs/Web/CSS/will-change).

A good way to use it might be something like...

```css
.widgets-box:hover .widget {
  will-change: transform;
}
```

where `.widgets-box` is the parent of all ze widgets. I dunno, could give it ago anyway.

ryantmer:
I believe I did give that a shot, but it didn't seem to have a noticeable effect :( I'll confirm, though, thanks! :)

ryantmer:
@apalaniuk mind taking a quick peek? Not particularly big, just want to make sure :)

apalaniuk:
@ryantmer My quick peek says :+1:

PR #33, opened 2016-06-07
Brightspace ESLint

### LINKED ISSUE #27: Use Brightspace eslint
Currently just extending the `eslint:recommended` settings, should be using https://github.com/Brightspace/eslint-config-brightspace

ryantmer:
Update to use the Brightspace ESLint rules and changes to reflect that. Fixes #27.

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -95,10 +95,10 @@
 			behaviors: [
 				Polymer.D2L.MyCourses.LocalizeBehavior
 			],
-			created: function () {
+			created: function() {
 				this.setAttribute('role', 'link');
 			},
-			updatePreferencesOnResponse: function (response) {
+			updatePreferencesOnResponse: function(response) {
 				if (response.detail.status === 200) {
 					var parser = document.createElement('d2l-siren-parser');
 					var preferencesEntity = parser.parse(response.detail.xhr.response);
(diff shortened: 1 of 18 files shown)


ryantmer:
@apalaniuk you'll defs want to rebase after this is merged :P

d2l-course-tile.html
```diff
@@ -107,14 +107,14 @@
...
-			updatePreferencesOnError: function (error) {
+			updatePreferencesOnError: function(error) {
 				error.preventDefault();
 				error.stopPropagation();
 				this._pinned = this._enrollmentEntity.getSubEntitiesByRel('preferences')[0].properties.hasOwnProperty('pinDate');
 			},
 			pinClickHandler: function(e) {
 				// Prevent the click from triggering navigation to the course
-				if(e != null) {
+				if (e !== null) {
```
apalaniuk: I believe this is called without arguments when pinning from the touch menu, so 'e' will be undefined in that case? This looks like the only case where moving from equality to identity operators may cause an issue.

d2l-touch-menu.html
```diff
@@ -2,233 +2,232 @@
...
+				this._menuItems = [].slice.call(Polymer.dom(this).querySelectorAll('d2l-touch-menu-item'));
+			},
+			attached: function() {
+				if (!this._initialized) {
+					// Only run once, as attaching to body will cause this to be called again
+					this._initialized = true;
+					this._owner = this.parentNode;
+					this._host = Polymer.dom(this).parentNode.host;
+
+					if (this.touchTarget === null) {
```
apalaniuk: This one too. For simplicity, I believe that check can be removed entirely.

d2l-course-tile.html
ryantmer: Ah, fair point. I'll just replace with `if (e)` in that case, which is more idiomatic to javascript anyway :)

apalaniuk:
LGTM :+1:

PR #36, opened 2016-06-09
fixing location of the "d2l-icons" component

dlockhart:
@ryantmer FYI had to fix this since BSI was expecting it in its "new" location (without the "ui").


ryantmer:
Ohhh man, sorry about that. Worked locally, but I guess something else was imporing them. Thanks! :)

dlockhart:
I don't think it was anything on your end, just referencing 2 different versions from my-courses and BSI which ended up in conflict.

PR #37, opened 2016-06-09
Make private properties private

ryantmer:
Well, pseudo-private at least. Any property beginning with a _ in Polymer is treated as "private"; more importantly, this makes it easier to know which ones should/can be set when instantiating the WC via attributes, and which ones are just used internally.


ryantmer:
@apalaniuk This looks kinda nasty, but all I did was rename some things to start with "_" if they're "private". Also moved non-private properties to the top of the `properties` object, just to make them easier to find.

apalaniuk:
Looks both nasty and good to me.

PR #38, opened 2016-06-09
Mobile pinning touch menu layout/menu styling

apalaniuk:


apalaniuk:
@ryantmer There are certainly a few things that will change based on rebasing once #37 is merged, so I'll do that once that's done.

d2l-touch-menu-item-styles.html
```diff
@@ -9,76 +10,92 @@
...
 			visibility: inherit;
+			transform: scale(0.1);
+			transition: 0.2s ease-in-out;
 		}
 
-		.longpress-menu-item-hover div {
-			background-color: var(--vui-color-celestine);
-			transform: scale(1.2) translateY(-1.0em);
+		.longpress-menu-item.visible {
+			transition: 0.2s ease-in-out;
```
ryantmer: nit: I believe this can just be `transition: inherit`

d2l-touch-menu-item.html
```diff
@@ -64,22 +64,36 @@
 				},
 				_hostNode: {
 					type: Object
+				},
+				touchRadiusPadding: {
```
ryantmer: Should this be private? (`_touchRadiusPadding`) If it's only being used by the component and not set elsewhere, then probs should

d2l-touch-menu.html
```diff
@@ -184,20 +186,58 @@
 				// By default, set menu below finger
```
ryantmer: Comment is no longer applicable, yay!

d2l-touch-menu-item.html
apalaniuk: I think you're right, that default as private makes sense.

d2l-touch-menu.html
```diff
@@ -184,20 +186,58 @@
...
+					var rotate2d = function(point, rads) {
+						var newPoint = {
+							x: point.x * Math.cos(rads) - point.y * Math.sin(rads),
+							y: point.x * Math.sin(rads) + point.y * Math.cos(rads)
+						};
+
+						return newPoint;
+					};
+
+					for (var counter = 0; counter < rotations.length; counter++) {
```
ryantmer: So, because this is checking through "okay" positions, I am seeing some inconsistent behaviour. For example, if I long press near the edge of the screen, it comes up diagonally from my finger, whereas I'd expect it to be out from the side (opposite the edge of the screen I'm near). I made an art to describe what I mean - 

Finger area | Menu position (relative to finger)
--------------- | -----------------------------------------
1 | below-right
2 | below-left
3 | above-left
4 | above-right
5 | below
6 | right
7 | left
8 | above
anywhere else | above

L'image

It's also computationally easier to do this afaik, since no trig is required, just need to add/remove horizontal/vertical offset based on the touch event's X and Y (relative to the element).

d2l-touch-menu-item.html
ryantmer: I'm also planning on expanding our built-in demos a bit, so knowing which properties can/"cannot" be set will make that easier :stuck_out_tongue_winking_eye:

d2l-touch-menu.html
ryantmer: Off-github conversation summary - we want to favour the menu being above the touch target, so current behaviour is indeed correct. Some of the weirdness was caused by Chrome and hover-on-long-press type stuff, but that'll be a future thing to look into.

d2l-touch-menu-item-styles.html
apalaniuk: I believe the element itself would need to have a transition set for it to inherit, which doesn't give desired results. A good candidate for simplification though.

ryantmer: I'm just leveraging the fact that any `.longpress-menu-item.visible` is guaranteed to have a `.longpress-menu-item` parent - unless this should be `.longpress-menu-item .visible`?

d2l-touch-menu-item-styles.html
```diff
@@ -9,76 +10,92 @@
...
+			position: relative;
+			box-shadow: 0px 1px 1px 0px rgba(86,89,92,0.20);
+			display: flex;
 		}
 
 		.longpress-menu-item-text {
+			@apply(--vui-font);
 			background-color: var(--vui-color-pressicus);
-			border: 0.1em solid white;
+			border: 0em transparent;
```
ryantmer: This can be removed

d2l-touch-menu-item-styles.html
```diff
@@ -9,76 +10,92 @@
...
 			-webkit-backface-visibility: hidden;
+			transition: fill 0.3s ease-in-out;
+			left: 0.21em;
+			top: 0.21em;
+		}
+
+		.longpress-menu-item-image-background {
+			background-color: var(--vui-color-white);
+			border-radius: 50%;
+			border: 0em transparent;
```
ryantmer: Oh, same ting, can be removed.

d2l-touch-menu.html
```diff
@@ -175,31 +177,63 @@
 
 				// Initial position of menu - if near the top of screen, show it below the initial tap
 				// TODO: Add positioning for side edges; should be total of 8 possible positions
```
ryantmer: No longer applicable, ja?

ryantmer:
Few little things, but LGTM :shipit: I feel like the position logic could be simpler, since it's kinda hard to follow at the moment, but a) that could just be Friday brain now allowing me to math, and b) it works, so all good!

apalaniuk:
@ryantmer It we don't care as much about displaying above the touch point, and make horizontal layout symmetric, definitely. But, I think layout will be changed once we add more items, so I'll revisit then =)

ryantmer:
Hm, yeah fair point. That'll complicate things fo sho.

PR #40, opened 2016-06-10
US72668 - Cleanup styles

### LINKED ISSUE #16: US72668 - Consolidate stylez
Styling is kind of spread out across the various components. Now that we can more directly use VUI styles, we should a) use them, and b) move styling up as much as possible, and inherit things.

ryantmer:
Reasonings and stuff:

- We shouldn't be importing `vui-typography`; the LMS is importing this already
- We _should_ explicitly import `vui-small-text`, as it's not as-guaranteed to be present
- `vui-typography` has moved away from variables to just classes; so, use classes rather than `@apply` rules (except for small text)
- Use Daylight styles as much as possible, and only override where we explicitly want to

TODO

- [x] Review explicit overrides with @thehappypixel 

Fixes #16

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -1,15 +1,15 @@
 <link rel="import" href="../vui-colors/colors.html">
+<link rel="import" href="../vui-typography/small-text.html">
 
 <dom-module id="d2l-course-tile-styles">
 	<template>
 		<style include="vui-colors"></style>
+		<style include="vui-small-text"></style>
 		<style>
 		:host {
 			display: block;
-			@apply(--vui-font);
 			-webkit-backface-visibility: hidden;
 		}
-
 		.tile-container {
 			position: relative;
 		}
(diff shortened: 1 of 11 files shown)


d2l-course-tile-styles.html
```diff
@@ -54,13 +52,16 @@
...
-			@apply(--vui-heading-4);
 			width: 100%;
 			max-height: 75px;
 			overflow: hidden;
 			position: relative;
 			margin-top: 0.4em;
 			margin-bottom: 1.9em;
+			/* overrides to vui-heading-4 */
+			font-weight: 400;
+			font-size: 20px;
```
apalaniuk: Curious: Any reason why font-size is overridden when vui-heading-4 sets it as 1rem and we use 20px as the root font size?

apalaniuk:
LGTM!

d2l-course-tile-styles.html
ryantmer: Without the override it ends up being 16px for whatever reason... the computed size is 1rem, from vui-typography, but that ends up being 16px for whatever reason :confused:

ryantmer: Huh. Upon further inspection, it's only 16px in the demo; in the LMS, it is indeed 20px. So, away this line goes! Thanks :D

PR #41, opened 2016-06-13
DE21953 - Exclude inactive courses

ryantmer:
While we may want these again in the future, for now they are going to show and since they may no longer exist on the LMS, we cannot change their pin state.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -32,7 +32,7 @@
 					{{localize('noPinnedCoursesMessage')}}
 				</d2l-alert>
 				<div class="my-courses-container" role="region">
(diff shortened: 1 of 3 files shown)


ryantmer:
@apalaniuk Mind taking a look? This is what's currently causing the extra courses that are showing up on the demo site (that can't be unpinned, and don't have course images).

apalaniuk:
@ryantmer LGTM - I'll have to think about how this affects the "dynamically update lists" story, but this looks good as-is.

ryantmer:
Hm, that's a good question... I'd image if you apply the same filter you should be okay? But not sure how you're going about that :=

PR #42, opened 2016-06-13
Use d2l-offscreen web component for offscreen text

ryantmer:
New WC that is intended to have the proper formatting/styles for offscreen components. We were previously relying on shady DOM leakage for this to work, but this method will now work with both shady and shadow DOM.


apalaniuk:
LGTM!

PR #43, opened 2016-06-13
Add interaction interface detection and only enable one of hover/touc…

apalaniuk:
- Will now detect (to a reasonable degree) whether hover and/or touch interfaces are supported, enabling only the detected interface, defaulting to both enabled if it can't be reasonably determined. This will need improvement in the future, but works reasonably well for now. Allows for more normal interaction on desktop (context menu, selection, etc)

- Rather than simulating click events on short touches, allow platform's click event to run if long-press menu wasn't opened. Also adds click prevention for a short time after closing to catch delayed click events on iOS.

- Corrected a transition that should have only applied to background-colour on touch menu items, but also delayed the menu item from being hidden.

- Disable pointer events on hover button in non-hover interface, so that it doesn't swallow click events on touch devices.

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -104,13 +104,26 @@
 			vertical-align: middle;
 		}
 		.no-tap-interaction {
-			-webkit-user-select: none;
 			-webkit-touch-callout: none;
 			-webkit-tap-highlight-color: rgba(0,0,0,0);
 			-moz-tap-highlight-color: rgba(0,0,0,0);
 			tap-highlight-color: rgba(0,0,0,0);
 			touch-action: pan-x pan-y;
 		}
+
+		@media not all and (hover: hover) {
+			a, .hover-menu, .menu-item {
+				-webkit-user-select: none;
+			}
+
+			:focus .course-text {
+				text-decoration: none;
+			}
+
+			.hover-menu {
+				display: none;
+			}
+ 		}
 		</style>
 	</template>
 </dom-module>
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -50,7 +50,7 @@
 			</div>
 		</div>
 
-		<d2l-touch-menu>
+		<d2l-touch-menu enabled="[[touchEnabled]]">
 			<d2l-touch-menu-item
 				text="{{localize('pin')}}"
 				background-image="d2l-tier1:pin-filled"
@@ -67,6 +67,14 @@
 				enrollment: {
 					type: Object,
 					observer: '_enrollmentChanged'
+				},
+				hoverEnabled: {
+					type: Boolean,
+					value: true
+				},
+				touchEnabled: {
+					type: Boolean,
+					value: true
 				},
 				_enrollmentEntity: {
 					type: Object,
@@ -137,11 +145,13 @@
 				});
 			},
 			hoverCourseTile: function() {
-				var hoverMenu = this.$$('.hover-menu');
-				Polymer.dom(hoverMenu).classList.add('hover');
+				if (this.hoverEnabled) {
+					var hoverMenu = this.$$('.hover-menu');
+					Polymer.dom(hoverMenu).classList.add('hover');
 
-				var courseImage = this.$$('.course-image img');
-				Polymer.dom(courseImage).classList.add('hover');
+					var courseImage = this.$$('.course-image img');
+					Polymer.dom(courseImage).classList.add('hover');
+				}
 			},
 			unhoverCourseTile: function(force) {
 				if (force === true || Polymer.dom(this.root).querySelectorAll(':hover').length === 0) {
--- a/d2l-touch-menu.html
+++ b/d2l-touch-menu.html
@@ -14,6 +14,9 @@
 		Polymer({
 			is: 'd2l-touch-menu',
 			properties: {
+				enabled: {
+					type: Boolean
+				},
 				_touchTarget: {
 					type: Object,
 					observer: '_touchTargetChanged'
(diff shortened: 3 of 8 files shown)


d2l-course-tile-styles.html
```diff
@@ -104,13 +104,26 @@
...
+			a, .hover-menu, .menu-item {
+				-webkit-user-select: none;
+			}
+
+			:focus .course-text {
+				text-decoration: none;
+			}
+
+			.hover-menu {
+				pointer-events: none;
```
apalaniuk: On second thought, `display: none` is probably better here.

d2l-course-tile.html
```diff
@@ -68,6 +68,14 @@
 					type: Object,
 					observer: '_enrollmentChanged'
 				},
+				hoverEnabled: {
```
ryantmer: Quuuuuessstion about this - aren't hover and touch basically opposites? As in, we'd only ever want to do one XOR the other, correct? If that is the case, is it possible to sort of combine them to default to hover, but use touch if available? Just thinking it might make the decision logic a bit simpler.

d2l-touch-menu.html
```diff
@@ -252,17 +235,37 @@
 				e.preventDefault();
 			},
 			_cancelMenu: function() {
-				if (this._timerID) {
-					clearTimeout(this._timerID);
+				if (this._openMenuTimerID) {
```
ryantmer: This whole block can just be `clearTimeout(this._openMenuTimerID)` - the function [is defined](https://html.spec.whatwg.org/multipage/webappapis.html#timers) to do nothing if the passed value is undefined/null or invalid.

d2l-touch-menu.html
```diff
@@ -252,17 +235,37 @@
...
 			},
-			_onMenuOpenChanged: function(openMenu) {
-				this._menu.style.visibility = openMenu ? 'visible' : 'hidden';
+			_onMenuOpenChanged: function(isOpen, wasOpen) {
+				this._menu.style.visibility = isOpen ? 'visible' : 'hidden';
+
+				if (isOpen) {
+					this._cancelClicks = true;
+				} else if (wasOpen) {
+					this._enableClicksTimerID = setTimeout(this._enableClicks.bind(this), 100);
```
ryantmer: I r confus by this :( Why do we need to wait to re-enable tapping/clicking?

d2l-touch-menu.html
```diff
@@ -252,17 +235,37 @@
...
-					clearTimeout(this._timerID);
+				if (this._openMenuTimerID) {
+					clearTimeout(this._openMenuTimerID);
 				}
-				this._timerID = null;
+				this._openMenuTimerID = null;
 			},
-			_onMenuOpenChanged: function(openMenu) {
-				this._menu.style.visibility = openMenu ? 'visible' : 'hidden';
+			_onMenuOpenChanged: function(isOpen, wasOpen) {
```
ryantmer: Shouldn't `isOpen === !wasOpen` always hold true, so can just have the `isOpen` parameter here?

d2l-course-tile.html
apalaniuk: My concern was the current inability to reliably detect a touchscreen device - checking for touch event support isn't sufficient (modern Chrome and Firefox on desktop support touch events), and it seems a small fraction of mobile browsers don't support the media queries that can more easily determine this, so it falls back to the current 'all enabled' if unsure. That would generally mean that a desktop platform that can't detect this would suffer (user-select: none, etc) but thinking about it, that also means the CSS media query to disable the hover button itself would fail, making mobile a bit trickier anyway, because the hover menu is only hidden through opacity.

It seems like if we're fine with [not supporting older mobile browsers](http://caniuse.com/#feat=css-media-interaction), then first checking for @media (hover: hover) being false ("The primary interaction interface doesn't allow you to easily hover"), plus any other reasonable checks for touch support, then using hover only if those checks fail, would seem to work.

d2l-touch-menu.html
apalaniuk: Hm, good to know. The wording was unclear to me as to whether a non-number was an acceptable handler not in the timer list, and [examples seem to have checks](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/clearTimeout), so thanks for letting me know it can be simplified!

d2l-touch-menu.html
apalaniuk: That could be the case - looking at Polymer's observer example, it implies that setting a property itself will trigger the observer, rather than a change, so it's slightly defensive in that regard. In practice I believe it could be removed (that is, I saw no issues without it).

d2l-touch-menu.html
apalaniuk: The previous method of mediating fake clicks/taps was not working as expected, so we're depending on the standard click event being fired for that case. On iOS, you'll receive a click event after what we consider a longpress, after the menu is determined to be closed, so simply canceling clicks while the menu is open will miss the delayed click. I'm not too thrilled about this solution either, but I can't find a clean way to handle it.

d2l-touch-menu.html
ryantmer: Yeah it seems to be pretty forgiving - `clearTimeout({})`, `clearTimeout('asdasgd')`, `clearTimeout(5)`, none seem to complain :octocat:

d2l-touch-menu.html
ryantmer: Ah, kk. Up to you which way to go :)

d2l-touch-menu.html
ryantmer: Ah butts, that's kinda dumb. Only on iOS too? Hmm... could you add a comment in that case? If we come up with something better, awesome, but if not then this is kinda confusing :P

d2l-course-tile.html
ryantmer: Hm, no Firefox either... =/ I get what you're saying though - I suppose with a touch-enabled desktop, there are actually circumstances where we'd want both behaviours to be available, so makes sense. Thanks! :)

ryantmer:
All LGTM I'd say :) Not an easy task to detect this sort of thing, apparently!

PR #44, opened 2016-06-15
Supercharge demo and docs

ryantmer:
Just wanted to be able to more easily test things without having to use them in the LMS; also means we can make sure our components are robust in a more isolated fashion, so that we can break them out with confidence if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,26 @@ are present, no request will be done to fetch all of a user's enrollments
 Demo:
 > http://localhost:8080/components/d2l-my-courses/demo/d2l-all-courses-demo.html
 
+### d2l-touch-menu
+
+The `d2l-touch-menu` is simply a container for `d2l-touch-menu-item`s. See the
+`d2l-touch-menu-item` demo for usage.
+
+### d2l-touch-menu-item
+
+The `d2l-touch-menu` is a long-press menu that is used on mobile, and can
+contain one or more `d2l-touch-menu-item`s.
+
+Properties:
+
+- `actionDescription` _String_: used for A11Y offscreen description
+- `backgroundImage` _String_: an icon to use for the menu item
+- `text` _String_: Text to display on menu item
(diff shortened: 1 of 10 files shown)


demo/d2l-touch-menu-demo.html
```diff
@@ -0,0 +1,69 @@
...
+						background-image="actions:close"
+						selection-handler="xHandler"
+						text="The X">
+					</d2l-touch-menu-item>
+				</d2l-touch-menu>
+			</div>
+			<script>
+				// This is some trickery required for the demo. In normal usage,
+				// this would not be required.
+				var menuParent = document.getElementById('menu-parent');
```
ryantmer: Just want to bring attention to this - I'm thinking we should update the touch-menu/touch-menu-item to not require the parent be a Polymer component, and instead just have the `selectionHandler` property be any function. Thoughts?

apalaniuk: That makes sense, and it's how I initially designed it (so long as you wanted to bind the selection handler imperatively).

README.md
```diff
@@ -51,38 +114,36 @@ to link to the local d2l-my-courses-ui project
 
 5) Build and run brightspace-integration (will have to be rebuilt on any changes to d2l-my-courses-ui)
 
-Note: On Windows, there exists an issue with relative paths, which will prevent web-component-shards from completing successfully without modifying vulcanize to not use path.posix. See: https://github.com/Polymer/vulcanize/issues/338
+Note: On Windows, there exists an issue with relative paths, which will
```
apalaniuk: This actually doesn't seem to be an issue now, but I added it, so I can correct it in another PR =)

apalaniuk:
LGTM, so long as you merge before #45 as that will change this documentation - then I'll update the docs in that PR.

README.md
ryantmer: Ah, no worries, might as well remove it here

PR #45, opened 2016-06-15
US72391 - Update Course Lists on Pin/Unpin

apalaniuk:
Add the ability to have the result of pin/unpin actions be reflected immediately (ie, pinning a course moves it to the unpinned courses list, and vice-versa).

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -15,10 +15,9 @@
 
 		<d2l-ajax
 			id="allEnrollmentsRequest"
-			url="[[enrollmentsUrl]]"
+			url="{{effectiveEnrollmentsUrl()}}"
 			headers='{ "Accept": "application/vnd.siren+json" }'
-			on-response="allEnrollmentsOnResponse"
-			last-response="{{_allEnrollmentsResponse}}">
+			on-response="allEnrollmentsOnResponse">
 		</d2l-ajax>
 
 		<d2l-simple-overlay
@@ -29,7 +28,7 @@
 
 			<div class="dialog-content">
 				<h3 class="d2l-heading-3">{{localize('pinnedCourses')}}</h3>
-				<d2l-alert visible="{{_hasPinnedCourses}}">
+				<d2l-alert visible="{{!_hasPinnedCourses}}">
 					{{localize('noPinnedCoursesMessage')}}
 				</d2l-alert>
 				<div class="my-courses-container" role="region">
@@ -59,25 +58,21 @@
 			is: 'd2l-all-courses',
 			properties: {
 				pinnedCoursesEntities: {
-					type: Array
+					type: Array,
+					notify: true
 				},
 				unpinnedCoursesEntities: {
 					type: Array,
-					readOnly: true
+					notify: true
 				},
-				allEnrollmentsEntities: {
-					type: Array,
-					readOnly: true
+				testEnrollmentsUrl: {
+					type: String
 				},
 				enrollmentsUrl: {
 					type: String
 				},
-				_allEnrollmentsResponse: {
-					type: Object
-				},
 				_hasPinnedCourses: {
-					type: Boolean,
-					computed: '_isEmpty(pinnedCoursesEntities)'
+					type: Boolean
 				},
 				_windowResizeHandler: {
 					type: Object,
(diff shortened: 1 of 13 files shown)


d2l-all-courses.html
```diff
@@ -28,11 +27,11 @@
 
 			<div class="dialog-content">
 				<h3 class="vui-heading-3">{{localize('pinnedCourses')}}</h3>
-				<d2l-alert visible="{{_hasPinnedCourses}}">
+				<d2l-alert visible="{{!_hasPinnedCourses}}">
```
ryantmer: O_O

Wait... how was this even working before?!

d2l-all-courses.html
```diff
@@ -58,46 +57,55 @@ <h3 class="vui-heading-3">{{localize('unpinnedCourses')}}</h3>
...
-					this._filterUnpinnedCourses(this.pinnedCoursesEntities, this.allEnrollmentsEntities);
+					var parser = document.createElement('d2l-siren-parser');
+					var allEnrollments = parser.parse(response.detail.xhr.response);
+					var allCourses = allEnrollments.getSubEntitiesByClass('course-offering');
+
+					var newPinnedCourses = [];
+					var newUnpinnedCourses = [];
+
+					for (var i = 0; i < allCourses.length; i++) {
+						if (allCourses[i].class.indexOf('inactive') === -1) {
```
ryantmer: `allCourses[i].hasClass('inactive')` is a thing too :D

apalaniuk: Good catch! [I don't remember where I picked that up from, though](https://github.com/Brightspace/d2l-my-courses-ui/commit/2c40d6f7c1f85aa65c1a0a0ab26e678c54620eb9#diff-98931a3c4fd43464bdaaf7d5c18ce360R98)... =P

Hm, it seems like the active enrollments just have the 'active' class set. Is one of active/inactive guaranteed?

ryantmer: Hahaha, shhhh... :P

And [yep, we good to have one or the other!](https://github.com/Brightspace/user-info-api/blob/47303877e85de6dbfe63ce9322abfe195a56bcc3/src/enrollments/transformers/siren.js#L87)

d2l-all-courses.html
```diff
@@ -58,51 +57,60 @@ <h3 class="vui-heading-3">{{localize('unpinnedCourses')}}</h3>
...
+						}
+					}
+
+					this.set('pinnedCoursesEntities', newPinnedCourses);
+					this.set('unpinnedCoursesEntities', newUnpinnedCourses);
+
 					this._rescaleCourseTileRegions(this._getOverlayWidth());
 				}
 			},
 			isActive: function(item) {
```
ryantmer: This function can go, not being used anymore :)

d2l-course-tile-region-behavior.html
```diff
@@ -8,6 +8,21 @@
 		Polymer.D2L.MyCourses = Polymer.D2L.MyCourses || {};
 
 		Polymer.D2L.MyCourses.CourseTileRegionBehavior = {
+			properties: {
```
ryantmer: Hmm, just thinking... does this actually need to be a property? Since we're not binding it to the DOM or watching for changes or anything, can it just be a property on `CourseTileRegionBehaviour`?

d2l-my-courses.html
```diff
@@ -80,28 +80,27 @@
 				enrollmentsUrl: {
 					type: String
 				},
+				testEnrollmentsUrl: {
```
ryantmer: Is this value actually set anywhere? It appears to only be read .-.

d2l-my-courses.html
```diff
@@ -113,43 +112,63 @@
...
 			attached: function() {
-				this._addThrottledEventListener('resize', this._windowResizeHandler);
 				if (!this.delayLoad) {
 					this.$.pinnedCoursesRequest.generateRequest();
 				}
 			},
-			detached: function() {
-				window.removeEventListener('resize', this._windowResizeHandler);
-			},
 			isActive: function(item) {
```
ryantmer: Same thing, can ditch this I do believe

d2l-my-courses.html
apalaniuk: @ryantmer: It's set in the demo - it's used to load a separate "All Enrollments" JSON file, as it previously just loaded up the same file both times, with a query string. Could be renamed to be clearer about its purpose.

ryantmer: Hm, I see. Not sure about adding a property that's only used in the demo, though... Is it possible for the demo to just cheat a bit (see [this comment](https://github.com/Brightspace/d2l-my-courses-ui/pull/44/files#r67193163) on my open PR), rather than adding the functionality here?

d2l-touch-menu.html
```diff
@@ -103,8 +103,6 @@
 				}
 			},
 			touchstartHandler: function(e) {
-				e.stopPropagation();
```
ryantmer: i r confus - why do we no longer need this? :(

apalaniuk: I'm not sure that we needed it in the first place (no observable change in behavior), but it seems best practice to leave it in.

demo/index.html
```diff
@@ -10,15 +10,24 @@
...
 	<body unresolved class="d2l-typography">
 		<main class="d2l-max-width">
-			<d2l-my-courses enrollments-url="my-courses.json"></d2l-my-courses>
+			<d2l-my-courses
+				id="my-courses"
+				enrollments-url="all-courses.json"
+				></d2l-my-courses>
+			<script>
+				// In the demo, we need to override the pinned courses request URL, as it's normally created by adding a query string to the enrollments URL -
+				// but we're loading a different file altogether.
```
ryantmer: Yeah - even though this approach requires some DOM-hacking/fiddling, I think it is overall nicer tbh, plus the comment makes is abundantly clear what is going on :)

demo/index.html
```diff
@@ -10,15 +10,24 @@
...
+		</script>
 		<link rel="stylesheet" href="demo.css">
 		<link rel="import" href="../d2l-my-courses.html">
 		<link rel="import" href="../../d2l-typography/d2l-typography.html">
 		<style is="custom-style" include="d2l-typography"></style>
 	</head>
 	<body unresolved class="d2l-typography">
 		<main class="d2l-max-width">
-			<d2l-my-courses enrollments-url="my-courses.json"></d2l-my-courses>
+			<d2l-my-courses
```
ryantmer: My #44 changed where the demos are, but apparently didn't cause conflicts... so I guess heads up for that :P

ryantmer:
LGTM, good stuffs! :smile:

PR #48, opened 2016-06-15
Fix: 4 courses should be 1 row, not 2

ryantmer:
Designs specify that having 4 courses should show up as 1 row of 4 courses, not 2 rows of 2 (the latter resulting in very large course tiles)


apalaniuk:
:+1:

PR #52, opened 2016-06-20
Unpin animation

apalaniuk:
Adds an animation on pinning/unpinning course tiles.

Also refactors common behavior of My Courses/All Courses into Polymer behavior, so they can handle course entity management independently; d2l-all-courses was previously depending on d2l-my-courses for state management.

### DIFF
--- a/d2l-course-management-behavior.html
+++ b/d2l-course-management-behavior.html
@@ -1,0 +1,133 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-siren-parser/d2l-siren-parser.html">
+
+<script>
+	(function() {
+		'use strict';
+
+		Polymer.D2L = Polymer.D2L || {};
+		Polymer.D2L.MyCourses = Polymer.D2L.MyCourses || {};
+
+		Polymer.D2L.MyCourses.CourseManagementBehavior = {
+			properties: {
+				pinnedCoursesEntities: {
+					type: Array,
+					value: [],
+					notify: true
+				},
+				unpinnedCoursesEntities: {
+					type: Array,
+					value: [],
+					notify: true
+				}
+			},
+			listeners: {
+				'course-pinned': '_onCoursePinned',
+				'course-unpinned': '_onCourseUnpinned',
+				'pin-animation-complete': '_onCoursePinAnimationComplete',
+				'unpin-animation-complete': '_onCourseUnpinAnimationComplete'
+			},
+			setPinnedCoursesFromResponse: function(response) {
+				var parser = document.createElement('d2l-siren-parser');
+				var allEnrollments = parser.parse(response.detail.xhr.response);
+				var allCourses = allEnrollments.getSubEntitiesByClass('course-offering');
+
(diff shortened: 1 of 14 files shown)


d2l-course-management-behavior.html
```diff
@@ -0,0 +1,160 @@
...
+			ready: function() {
+				this._announceElement = document.createElement('d2l-offscreen');
+				this._announceElement.setAttribute('aria-live', 'polite');
+			},
+			setPinnedCoursesFromSirenResponse: function(response) {
+				var parser = document.createElement('d2l-siren-parser');
+				var allEnrollments = parser.parse(response);
+				var allCourses = allEnrollments.getSubEntitiesByClass('course-offering');
+
+				this.set('pinnedCoursesEntities', allCourses.filter(function(item) {
```
ryantmer: The effect is the same, but you can use the siren parser again here - `this.pinnedCoursesEntities = allCourses.getSubEntitiesByClass('active')`.

ryantmer: Wait, nvm. Forgot that returns an array, which you can't `getSubEntitiesByClass` on. Carry on!

ryantmer: Ah, I remember why I was confused. At one point there was discussion of having `getSubEntitiesByClass(['course-offering', 'active'])` be a thing. Never happened though. Also might have been an `or` on the class names, not an `and`. So, doubly useless in this context.

Sorry for comment spam. :meat_on_bone:

d2l-course-management-behavior.html
```diff
@@ -0,0 +1,160 @@
...
+					return item.hasClass('active');
+				}));
+
+				this._hasPinnedCourses = this.pinnedCoursesEntities.length > 0;
+				this._hasCourses = this._hasPinnedCourses;
+			},
+			setAllCoursesFromSirenResponse: function(response) {
```
ryantmer: Hmm... now that I'm looking at these side-by-side, can we just ditch one of these response handlers? The only difference is that the one for pinned has one less level of nesting, but due to the circumstances it's not actually going to save anything anyway :balloon:

d2l-course-tile.html
```diff
@@ -103,9 +109,20 @@
 			behaviors: [
 				Polymer.D2L.MyCourses.LocalizeBehavior
 			],
+			listeners: {
+				'touch-pin-hover': '_onUnpinHover',
+				'touch-pin-select': 'pinClickHandler',
+				'animationend': '_onAnimationComplete'
```
ryantmer: (assuming this isn't a neon thing), can this be `animation-end` for consistency? (although I assume this is a neon thing)

apalaniuk: It's not a neon animation - it's the [CSS animationend event](https://developer.mozilla.org/en-US/docs/Web/Events/animationend), so I'm not sure if there's a good way to align naming?

ryantmer: Ah! Nvm then, all good as-is :)

ryantmer:
Many :+1:s!

PR #54, opened 2016-06-21
Fix movement of touch menu icon on transition

### LINKED ISSUE #39: Mobile pin icon on Safari moves slightly within parent container at beginning/end of transition
Observed only on mobile Safari; seems as though an ease-in-out transition is having an unintended effect on positioning.

apalaniuk:
Resolves #39.

Also fixes an issue where on unhovering menu item, menu item text would not fade out to match its fade-in transition.

### DIFF
--- a/d2l-touch-menu-item-styles.html
+++ b/d2l-touch-menu-item-styles.html
@@ -50,6 +50,9 @@
 			transition: 0.3s ease-in-out;
 			opacity: 0;
 		}
+		.longpress-menu-item.visible.selected .longpress-menu-item-text {
+			opacity: 0;
+		}
 
 		.longpress-menu-item-image {
 			--iron-icon-width: 0.58em;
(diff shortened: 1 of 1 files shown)


d2l-touch-menu-item-styles.html
```diff
@@ -58,6 +61,7 @@
 			-webkit-backface-visibility: hidden;
 			-webkit-transition: fill 0.3s ease-in-out;
 			transition: fill 0.3s ease-in-out;
+			transform: rotate(0.0001deg);
```
ryantmer: Huh, I'd read about this trick, but it didn't end up working to make the saturation/contrast transition any smoother. But it actually works for this? Weird... how2css.

ryantmer:
Code looks good; can't test it since I have not the Macs :( But I trust ye, so LGTM :lollipop:

PR #56, opened 2016-06-21
Refactor d2l-all-courses to not contain a d2l-simple-overlay

ryantmer:
This simplifies things a bit, as it presents d2l-all-courses as just a modified d2l-my-courses, which it kinda is. This should also enable removal of d2l-all-courses in the future if that's something we decide to do.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -3,7 +3,6 @@
 <link rel="import" href="../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="d2l-alert.html">
 <link rel="import" href="d2l-all-courses-styles.html">
-<link rel="import" href="d2l-simple-overlay.html">
 <link rel="import" href="d2l-course-tile.html">
 <link rel="import" href="d2l-course-tile-region-behavior.html">
 <link rel="import" href="localize-behavior.html">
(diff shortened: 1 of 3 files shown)


d2l-all-courses.html
```diff
@@ -21,41 +20,34 @@
...
-					<template is="dom-repeat" items="[[pinnedCoursesEntities]]">
-						<d2l-course-tile
-							enrollment="[[item]]"
-							hover-enabled="[[_hoverInteractionEnabled]]"
-							touch-enabled="[[_touchInteractionEnabled]]"
-							animate-insertion="[[removeCourseFromTransitionList(item.properties.id)]]">
-						</d2l-course-tile>
-					</template>
-				</div>
+		<div class="dialog-content">
```
ryantmer: Nothing actually changed here, aside from removing the outer `d2l-simple-overlay`, but github doesn't have a `diff -w` option apparently :crying_cat_face:

d2l-all-courses.html
```diff
@@ -72,9 +64,6 @@ <h3 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h3>
 				Polymer.D2L.MyCourses.InteractionDetectionBehavior,
 				Polymer.D2L.MyCourses.CourseManagementBehavior
 			],
-			ready: function() {
```
ryantmer: Since `all-courses` is now already attached to the DOM right away, fetching the courses on `ready` means the overlay has the wrong width when it opens - funny enough, it seems like the time it takes to fetch the enrollments (even if it's from a file, like in the demo) is long enough for the simple-overlay container to get a width, and therefore for the right number of columns to appear!

d2l-my-courses.html
```diff
@@ -90,12 +99,26 @@
 			],
 			ready: function() {
 				this._alertMessage = this.localize('noPinnedCoursesMessage');
+				this.$$('d2l-simple-overlay').focusableNodesOverride = this.focusableNodesOverride;
```
ryantmer: Does this make sense? I still don't fully understand this magic, but I just moved it up one layer... I think that's right @apalaniuk?

apalaniuk: I should document this =)

The point of focusableNodesOverride is that [iron-overlay-behavior selects elements that can get focus](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L189) for keyboard tabbing/wraparound, but only supports basic HTML focusables that are direct children, so we need to give it the list of elements for our case (really just first/last) of course tiles. It looks like it's not finding the course tile elements now. It seems like there are two ways around this:

1. Change the selector query to ('* /deep/ d2l-course-tile') (but /deep/ is deprecated now =( )
2. Put focusableNodesOverride in d2l-all-courses (which I think makes sense, since the query is tied to its local DOM), and rather than call getContentChildren() (which I'm not sure how that worked, since d2l-all-courses didn't have a <content> element), it gets all course tile elements using `Polymer.dom(this.root).querySelectorAll('d2l-course-tile')` which seems to work.

ryantmer: Ah, that makes sense. I was going to update this PR to include the removal of `all-courses`, which I'm assuming would simplify this, but I'll just fix this and make the removal a separate PR.

ryantmer:
hokay, after like 15 WIP-to-not-WIP transitions there, this should be good for review :P

apalaniuk:
LGTM! :+1:

PR #58, opened 2016-06-21
Remove all courses

ryantmer:
Removed `d2l-all-courses` as it doesn't really need to be its own thing. Also changed up fetching a bit - now, we just fetch all enrollments. This won't work for users with a lot of enrollments, but this wasn't working back-end either (i.e. we weren't truly getting all the pinned courses first), so this issue needs to be looked at there rather than here.


d2l-all-courses-styles.html
```diff
@@ -1,21 +0,0 @@
-<link rel="import" href="d2l-course-tile-region-styles.html">
-
-<dom-module id="d2l-all-courses-styles">
-	<template>
-		<style include="d2l-course-tile-region-styles"></style>
-		<style>
-		:host {
-			display: block;
-		}
-		h3 {
```
ryantmer: This is the only unique style in here - I did move it to the my-courses styles

d2l-my-courses.html
```diff
@@ -62,12 +54,34 @@
 			locale="[[locale]]"
 			with-backdrop>
 
-			<d2l-all-courses
-				enrollments-url="[[enrollmentsUrl]]"
-				pinned-courses-entities="{{pinnedCoursesEntities}}"
-				unpinned-courses-entities="{{unpinnedCoursesEntities}}"
-				locale="[[locale]]">
-			</d2l-all-courses>
+			<div class="overlay-content">
```
ryantmer: This is just the content of the former `d2l-all-courses`, copied in verbatim

d2l-my-courses.html
```diff
@@ -101,32 +111,27 @@
 				this._alertMessage = this.localize('noPinnedCoursesMessage');
 
 				var overlay = this.$$('d2l-simple-overlay');
-				var allCourses = this.$$('d2l-all-courses');
-				overlay.focusableNodesOverride = allCourses.focusableNodesOverride.bind(allCourses);
+				var allCourses = this.$$('.overlay-content');
+				overlay.focusableNodesOverride = this.focusableNodesOverride.bind(allCourses);
```
ryantmer: This... appears to have stopped working again. Damn I'm bad at this thing :P

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -263,6 +321,97 @@ describe('smoke test', function() {
 		});
 	});
 
+	describe('Course entity management', function() {
```
ryantmer: These tests were all just copied over, and updated the variable names as required. No behavioural change or anything.

ryantmer:
Hokay, this is actually good for review now. Had some minor bugs come up, but they have been summarily squashed.

d2l-my-courses.html
```diff
@@ -141,10 +146,10 @@
...
-			},
-			_getPinnedCoursesUrl: function(enrollmentsUrl) {
-				return enrollmentsUrl + '?filter=pinned';
+
+				setTimeout(function() {
+					this._rescaleCourseTileRegions();
+				}.bind(this), 100);
 			},
 			get courseTileItemCount() {
 				return this.pinnedCoursesEntities.length;
```
apalaniuk: This aligns the implementation with the design for desktop with four pinned courses :+1:

I wonder about the desktop zero/one pinned courses cases; zero pinned courses will display four columns, one pinned course will display two columns. I'm not sure if this is unintended since there aren't any designs for these cases, just bringing it up to get your thoughts.

apalaniuk:
LGTM!

d2l-my-courses.html
ryantmer: Ah, yep, that's what was being done in `courseTileItemCount` from `d2l-all-courses`. Good catch! I'll update it.

ryantmer:
As discussed, going to close this without merging. There are some advantages to this, but the unfortunately part is that by moving all-courses out of a web component, we (surprise!) lose a lot of the benefits of web components. Going to make a new PR to remove the extra fetching logic though, which will simplify things a bit.

PR #59, opened 2016-06-23
Always fetch all courses

ryantmer:
There are a few possibilities for the enrollments we can get back from an enrollments call:

1. The user has a small number of enrollments (pinned + unpinned)
2. The user has a lot of enrollments, many of them pinned
3. The user has a lot of enrollments, and a few pinned ones

In case 1, fetching all enrollments right away just makes sense - the reply isn't huge, and we avoid one network call overhead.

For 2 and 3, we can't actually sort by pin state on Valence, which means even though we're asking for pinned only from the UIS, there's no guarantee that the first set of enrollments coming back from Valence has all the pinned ones anyway. So, might as well just get them all. This will obviously change once we're on the new APIs.


test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -241,131 +47,62 @@ describe('smoke test', function() {
...
-				'GET',
-				widget.enrollmentsUrl,
-				function(req) {
-					expect(req.requestHeaders['accept']).to.equal('application/vnd.siren+json');
-					req.respond(200, allEnrollmentsResponse.headers, JSON.stringify(allEnrollmentsResponse.body));
-				});
-		});
-
-		describe('filter calculations', function() {
-			it('Should result in pinned and unpinned entity lists being populated with active courses only', function(done) {
```
apalaniuk: Are we treating this test as being satisfied by other tests, such as the 'should display appropriate message when no pinned courses' in d2l-my-courses?

apalaniuk:
[fetchEnrollments can be removed from the docs](https://github.com/Brightspace/d2l-my-courses-ui/blob/always-fetch-all-courses/README.md#d2l-all-courses).

test/d2l-all-courses/d2l-all-courses.js
ryantmer: Oh, whoops, did mean to call this out explicitly. Yeah, because of the fact that there's no longer a request occurring in all-courses, this test would (after simplification and correction) turn in to "make sure the thing I set is set to the thing I set it to" (since we're just giving it the pinned/unpinned entities directly now)

ryantmer:
Ah, and `enrollmentsUrl` too, good call.

test/d2l-all-courses/d2l-all-courses.js
apalaniuk: Yeah, in that capacity, d2l-all-courses was a sort of proxy for testing d2l-course-management-behavior as well. Since we're really doing BDD, I think everything's still fine in terms of testing through alerts, just want to make sure no gaps appear. :smile:

ryantmer: Oh, that's a fair point actually, we're not actually properly testing the behaviours... hmm... different PR. I'll make an issue for that, since that's where most of the complicated stuff is anyway :package:

apalaniuk:
LGTM and tests fine! :shipit:

PR #61, opened 2016-07-02
Fix Webkit image aliasing during transform

apalaniuk:
Fixes an issue in Webkit browsers where applying multiple filters in the same property and applying a transform causes aliasing/stepping, by distributing filters to parent elements so that each element has at most one filter applied.


ryantmer:
O_O

THIS IS SO MUCH BETTER.

ARE YOU A WIZARD?

LSFGTM.

PR #65, opened 2016-07-08
Fix for very wide images displaying incorrectly

ryantmer:
Since it was being displayed with width: 100%, space (and therefore the grey background) was showing up above/below the image.


ryantmer:
... I see no actual difference in the majority of the demo data. A ?w=1 will show that I just updated two of the course image URLs - one to have a very wide/short photo, and one to have a tall/skinny photo, to verify that it works as intended (it does).

PR #66, opened 2016-07-12
upgrading to offscreen 2.1.0

dlockhart:
Need to bump this as another thing in BSI needs a new feature in `2.1.0`. @ryantmer, @apalaniuk is it cool if I do a release of `my-courses` after this change is in and update BSI?


ryantmer:
LGTM :+1:

PR #68, opened 2016-07-13
Minor fixes

ryantmer:
- Remove unneeded sinon dependency
- add `use strict` to get rid of warnings (and remove eslint override)
- use latest Firefox - 47.0.1 was released last week and fixes the issue we were having before
- some tabbing/whitespace changes

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,6 @@
 extends: brightspace
 env:
   browser: true
-rules:
-  strict: warn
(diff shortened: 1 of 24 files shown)


ryantmer:
Going to just merge this - larger change coming, and this is pretty innocuous (in the sense that any of these changes would have blown up Travis if they'd been problematic :stuck_out_tongue_winking_eye: )

PR #69, opened 2016-07-14
Clarify naming

ryantmer:
I think this disambiguation makes sense:

- Enrollment - entity from /enrollments
- Organization - entity from /organizations
- Course - "touchy feely" term; used for more human-ish interactions (e.g. localized text) and to refer to things that are really enrollment + organization (e.g. the course tile, which "has both entities within it")

(Also removed the `icons.html` file that I forgot to remove in the last PR...)

PR #70, opened 2016-07-14
Simplify d2l-alert

ryantmer:
(I'll rebase this once #69 is merged, and it'll remove like 90% of this diff)

Just realized that d2l-alert was about 2x more complicated than it needs to be. Which, considering how un-complicated it is, meant it really wasn't that complicated. But, meh, might as well do the simpler way.

### DIFF
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -5,7 +5,7 @@
 	<template>
 		<style include="d2l-alert-styles"></style>
 
-		<div class="message-wrapper" hidden$="{{_hideAlert}}">
+		<div class="message-wrapper" hidden="{{!visible}}">
 			<div class="message-highlight"></div>
 			<content></content>
 		</div>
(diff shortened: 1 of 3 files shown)


d2l-alert.html
```diff
@@ -5,7 +5,7 @@
 	<template>
 		<style include="d2l-alert-styles"></style>
 
-		<div class="message-wrapper" hidden$="{{_hideAlert}}">
+		<div class="message-wrapper" hidden="{{!visible}}">
```
apalaniuk: Do we not still have to use $= to bind dynamic values to native attributes (especially boolean attributes)?

https://www.polymer-project.org/1.0/docs/devguide/data-binding#native-binding

apalaniuk:
If this is just waiting for a rebase, the actual simply alert commit LGTM

d2l-alert.html
ryantmer: Huh, weird. I forgot that was a thing, but apparently all good... apparently? :stuck_out_tongue_winking_eye:

PR #72, opened 2016-07-14
Add loading spinner to d2l-my-courses

apalaniuk:


d2l-my-courses-styles.html
```diff
@@ -35,6 +35,12 @@
 		d2l-alert {
 			margin-bottom: 30px;
 		}
+		.hidden {
```
ryantmer: There an advantage to having this as a class rather than just setting the style directly? Honestly no idea if one is better than the other.

apalaniuk: @ryantmer There is when you don't reinvent the wheel!

ryantmer:
The irony is that the loading spinner actually sometimes loads before even the widget title does, hah. LGTM! :shipit:

PR #77, opened 2016-07-17
Simplify how column recalculation is done

ryantmer:
Rather than calling into and out of various elements to manually trigger their `_rescaleCourseTileRegions` method, instead just fire the the `recalculate-columns` event whenever something happens that warrants recalculating how many columns should be displayed. Currently this includes:

- window resize
- receiving enrollments
- opening all courses overlay
- closing all courses overlay

It explicitly does *not* include when a tile is actually moved from one state to the other - the reason being, I'm not sure if this would be a jarring experience. This does mean that, if you're unpinning/unpinning courses, the number of columns won't change (until a refresh, resize, or if you're doing it in the all courses overlay and you close it).


d2l-all-courses.html
```diff
@@ -14,24 +14,25 @@
...
-			<template is="dom-if" if="[[!_hasPinnedEnrollments]]">
-				<d2l-alert visible="[[!_hasPinnedEnrollments]]">
-					{{localize('noPinnedCoursesMessage')}}
-				</d2l-alert>
-			</template>
-			<d2l-course-tile-grid
-				enrollments="{{pinnedEnrollments}}"
-				enrollments-queue="{{pinnedEnrollmentsQueue}}">
-			</d2l-course-tile-grid>
+		<h2 class="d2l-heading-3">{{localize('pinnedCourses')}}</h2>
```
ryantmer: `w=1` would show this for what it really is - I just removed the outer `dialog-content` div, since it's not required.

d2l-all-courses.html
```diff
@@ -46,12 +47,10 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 			],
 			ready: function() {
 				this.usePendingLists = true;
-			},
-			created: function() {
-				var courseTileGrids = Polymer.dom(this.root).querySelectorAll('d2l-course-tile-grid');
 
+				// Both course tile grids in this view should have the same number of columns, so use a custom getter
+				var courseTileGrids = Polymer.dom(this.root).querySelectorAll('d2l-course-tile-grid');
```
ryantmer: Not sure what black magic had this working before - `created` is fired before the local DOM is constructed, so no idea how this `querySelectorAll` was returning anything... either way, moved it into `ready` to more reliably work with this event-firing method.

d2l-course-tile-grid-styles.html
```diff
@@ -12,7 +12,7 @@
 			touch-action: pan-x pan-y;
 		}
 
-		.my-courses-container {
+		.course-tile-container {
```
ryantmer: (Just a name change to more accurately describe what these containers are)

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -332,83 +332,4 @@ describe('d2l-my-courses', function() {
 			});
 		});
 	});
-
```
ryantmer: Just moved these into the responsive grid tests, unmodified.

apalaniuk:
Awesome! :+1:

PR #80, opened 2016-07-18
Update d2l-ajax-ui

ryantmer:
This version includes the XSRF header on same-domain requests that aren't GET or HEAD, which is apparently a requirement of the LMS API routes. (We're using PUT to change the pin state, so this was a problem if not using canonical namespaces.)


apalaniuk:
LGTM

PR #82, opened 2016-07-22
Remove explicit enrollmentsUrl override

ryantmer:
The LMS is now setting this properly, so it can be removed from here.


ryantmer:
`bower link`ed this on my local instance, seems good. (I mean, it would either properly fetch enrollments or it wouldn't... and it did.)

PR #84, opened 2016-07-26
Fix animation in IE11

ryantmer:
Looks like IE doesn't support String.prototype.includes at all, so just switching back to good ol' indexOf.


apalaniuk:
:+1:

PR #88, opened 2016-07-29
Fix for extra space on right side of widget

ryantmer:
There were two problems at play here, it seems.

First, the width of the course-tile-container should be 100% + 20px, not 100 + 10px. If I'm honest, I don't fully understand why - the margin is -10px, so you'd think +10px would counteract that. But nope, extra 10px shows up on the right side. Okay CSS, you do you.

Second, the width calculation for margins for the 2-column layout were actually slightly wrong, so that was fixed. Also updated the other n-column margins to be the same.


apalaniuk:
Awesome!

PR #89, opened 2016-07-29
Aggregate styles

ryantmer:
Just combined them in the cases where inlining the styles didn't make things too frighteningly long. Could probably do this to everything, since it makes it easier to cross-reference styles with their classes in the JS, but... eh, we'll see.


d2l-course-tile-grid.html
```diff
@@ -1,5 +1,4 @@
 <link rel="import" href="../polymer/polymer.html">
-<link rel="import" href="d2l-my-courses-styles.html">
```
ryantmer: To clarify - these styles were being linked, and included, but not actually used as far as I can tell.

ryantmer:
(Realized that this is a non-behavioural change, so really doesn't need much review.)

PR #90, opened 2016-08-02
Use polylint to catch Polymer-y errors

ryantmer:
Just a few small fixes, we were mostly good to go (also rearranged the properties in course-tile to be alphabetical since I was in there anyway).

### DIFF
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -10,7 +10,7 @@
 	<template>
 		<style include="d2l-alert-styles"></style>
 
-		<div class="message-wrapper" hidden="{{!visible}}">
+		<div class="message-wrapper" hidden$="{{!visible}}">
 			<div class="message-highlight"></div>
 			<content></content>
 		</div>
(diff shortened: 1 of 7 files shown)


ryantmer:
(once Travis (finally) passes, just going to merge this as it's non-behavioural)

d2l-course-tile-grid.html
```diff
@@ -15,12 +15,6 @@
 		<style include="d2l-my-courses-styles"></style>
 		<style include="d2l-course-tile-grid-styles"></style>
 
-		<template is="dom-if" if="[[alertVisible]]">
```
ryantmer: Turns out this wasn't be used at all, I think it's a leftover from when this was broken out of `d2l-my-courses.html`

d2l-alert.html
```diff
@@ -10,7 +10,7 @@
 	<template>
 		<style include="d2l-alert-styles"></style>
 
-		<div class="message-wrapper" hidden="{{!visible}}">
+		<div class="message-wrapper" hidden$="{{!visible}}">
```
ryantmer: @apalaniuk turns out you were spot on about these!

PR #92, opened 2016-08-03
Remove d2l classes

ryantmer:
This came up because of `d2l-hidden` not working correctly in the demos, and led me to realize that these are implicitly-available external dependencies, which isn't overly great (the typography and other stuff is explicitly included, so it's okay). Plus, super simple classes, so not the end of the world to just redefine them and avoid any potential headaches.


ryantmer:
@apalaniuk quick peek? Shouldn't need any testing or anything, just want to make sure it's sane and all that :D

d2l-my-courses.html
```diff
@@ -46,6 +46,9 @@
 			.all-courses-button:hover {
 				text-decoration: underline;
 			}
+			.hidden {
```
apalaniuk: I'm almost positive this was a nit on one of my earlier PRs. Can't remember who wrote it, though...

ryantmer: Oh wait really? The nit being to use `d2l-hidden`? That could very well have been me in fact...

apalaniuk:
LGTM

PR #93, opened 2016-08-03
Fix hidden on d2l-alert

ryantmer:
Turns out that elements that have hidden=false are still kinda-displayed, in that they still take up space for their padding/margin. Use display:none to overcome this (still going to keep hidden=true/false, since it's a more semantic meaning).

### DIFF
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -63,7 +63,15 @@
 				*/
 				visible: {
 					type: Boolean,
-					value: false
+					value: false,
+					observer: '_visibleChanged'
+				}
+			},
+			_visibleChanged: function(visible) {
+				if (visible) {
+					this.style.display = 'block';
+				} else {
+					this.style.display = 'none';
 				}
 			}
 		});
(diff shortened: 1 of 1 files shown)


ryantmer:
@apalaniuk Another little one - turns out the HTML `hidden` doesn't stop the element from taking up space, which is problematic. So, set `display: none` as well!

apalaniuk:
I think the issue is that we leave display: block on d2l-alert, and display overrides the hidden attribute. Perhaps we should remove that attribute on any elements that we're using 'display' on, or have an attribute selector on [hidden] that sets display: none?

I can't find a good answer right now, so I'd say LGTM until I know better..

ryantmer:
`display` does override hidden, which is kind of weird, since they're not _really_ related. From what I can figure, `hidden` is intended for a more semantic, "this is no longer relevant" sort of thing, whereas `display` is a more "I don't want this in my DOM" thing. In this case, the alert is no longer relevant if you have courses, but we _also_ want it to not break spacing, so need the `display: none`.

In summary: yeah I just removed the `display: block`.

PR #94, opened 2016-08-05
US72258 Recent courses UI

ryantmer:
API changes for this were just merged, so you'll need to pull/rebuild to get it to work properly.

Introduced the `d2l-course-list` and `d2l-course-list-item` - these are basically just different views of the course tile/course tile grid we already have, with similar functionality. I'd imagine could probably be factored out to some degree, but I'd rather not make this already-large-PR even larger =/

### DIFF
--- a/d2l-course-list-item.html
+++ b/d2l-course-list-item.html
@@ -73,7 +73,7 @@ See the course tile docs for an important note about and enrollment vs an organi
 				* Course list item enrollment
 				*
 				* When the enrollment is set, the list item will automatically fetch information about the
-				* organization the enrollment is is.
+				* organization the enrollment is related to.
 				*
 				* @type {Entity}
 				*/
--- a/d2l-course-list-item.html
+++ b/d2l-course-list-item.html
@@ -56,12 +56,8 @@ See the course tile docs for an important note about and enrollment vs an organi
 			<div class="course-image">
 				<img aria-hidden="true" />
 			</div>
-			<div>
-				{{_organization.properties.name}}
-			</div>
-			<div class="course-code">
-				{{_organization.properties.code}}
-			</div>
+			<div class="course-text">{{_organization.properties.name}}</div>
+			<div class="course-code">{{_organization.properties.code}}</div>
 		</a>
 
 	</template>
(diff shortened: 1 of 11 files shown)


ryantmer:
This failure confuses me to no end... it says Firefox is not found, but Firefox is included in all Travis builds... I suspect something is wonky on their end.

d2l-my-courses.html
```diff
@@ -150,6 +198,21 @@
 				*/
 				_enrollmentsSearchUrl: {
 					type: String
+				},
+				/*
+				* URL constructed to fetch a user's recent enrollments with the enrollments search Action
+				* @type {String}
+				*/
+				_recentEnrollmentsSearchUrl: {
```
apalaniuk: Since you love conciseness: You probably already know this, but if type is the only field being defined on a property, you can just do eg.

`_recentEnrollmentsSearchUrl: String`

d2l-course-list-item.html
```diff
@@ -0,0 +1,130 @@
...
+
+		Polymer({
+			is: 'd2l-course-list-item',
+
+			properties: {
+				/*
+				* Course list item enrollment
+				*
+				* When the enrollment is set, the list item will automatically fetch information about the
+				* organization the enrollment is is.
```
apalaniuk: nit: 'is is'

d2l-my-courses.html
```diff
@@ -150,6 +198,21 @@
...
+					type: String
+				},
+				/*
+				* Currently-selected tab - set by paper-tabs, used by iron-pages
+				* @type {Number}
+				*/
+				_selectedTab: {
```
apalaniuk: Seems unused?

Also, it seems like the ARIA tab model fits well here. If you used paper-tabs, was there something about its behavior related to accessibility that didn't work well?

ryantmer: Oh, whoops. Yep, leftover from using `iron-pages` and `paper-tabs`. The reason for not using them was just because they can do a lot more than we need, and weren't overly customizable (at least, not in a nice way).

But, you are correct in that this should be A11y-d.

apalaniuk:
LGTM!

apalaniuk:
LGTM!

PR #96, opened 2016-08-08
Don't show Last Accessed for <20 enrollments

ryantmer:
Slightly increasing pageSize to accomodate for this check as well. If a user has <20 enrollments, only show the tile-based view.

Only increased it to 25 as this gives us enough information to know whether the user has "lots" of enrollments or not, and I'd rather not increase it too much until we've done better analysis on the API performance.


apalaniuk:
Looks good. Added a commit to fix a couple of things, so check that one yourself too!

PR #98, opened 2016-08-09
Live search

apalaniuk:
Slightly WIP, but mostly ready for PR. Some known issues:

* Search bar requires some general styling changes
* Probably a few other little things


demo/organizationsSearchResponse.json
```diff
@@ -0,0 +1 @@
..."href":"https://7e448bab-3d35-4791-a4dd-ea989243f73d.organizations.proddev.d2l/121701","rel":["self"]},{"href":"http://wl81-apalaniuk:44444/d2l/home/121701","rel":["https://proddev.d2l/rels/organization-homepage"],"type":"text/html"}]},{"class":["active","course-offering"],"properties":{"name":"Calculus 3 (Advanced Level)","code":"MATH247"},"rel":["https://proddev.d2l/rels/organization"],"links":[{"href":"https://7e448bab-3d35-4791-a4dd-ea989243f73d.organizations.proddev.d2l/121715","rel":["self"]},{"href":"http://wl81-apalaniuk:44444/d2l/home/121715","rel":["https://proddev.d2l/rels/organization-homepage"],"type":"text/html"}]}],"actions":[{"name":"search-course-collection","method":"GET","href":"https://7e448bab-3d35-4791-a4dd-ea989243f73d.organizations.proddev.d2l","fields":[{"name":"search","type":"search","value":"Calc"},{"name":"page","type":"number","value":"1"},{"name":"pageSize","type":"number","value":"10"}]}],"links":[{"rel":["self"],"href":"https://7e448bab-3d35-4791-a4dd-ea989243f73d.organizations.proddev.d2l?search=Calc&page=1&pageSize=10"},{"rel":["next"],"href":"https://7e448bab-3d35-4791-a4dd-ea989243f73d.organizations.proddev.d2l?search=Calc&page=2&pageSize=10"}]}
```
ryantmer: https://jsonformatter.curiousconcept.com/ \o/

d2l-utility-behavior.html
```diff
@@ -33,6 +34,11 @@
 				}
 			});
 
+			// Temporary override, as some actions don't specify embedDepth as a parameter on the action. Could invert above logic or, well, fix in LMS
```
ryantmer: Oh wait, /organizations doesn't have `embedDepth` listed as an action? O_o

d2l-search-listbox.html
```diff
@@ -0,0 +1,171 @@
...
+					}
+				}
+			},
+
+			/*
+			* Override IronMenuBehavior._focusPrevious so that the listbox owner can be focused
+			*/
+			_focusPrevious: function() {
+				var length = this.items.length;
+				var curFocusIndex = Number(this.indexOf(this.focusedItem));
```
ryantmer: `Number` is redundant no? That said, I can't actually find any documentation on `indexOf` for Polymer... is that a built-in thing?

apalaniuk: @ryantmer You're right, just adds clutter. These are unabashedly lifted from the underlying behavior implementation.

`indexOf` is a Polymer array method, but they only seem to mention it in the form of a narrative rather than in the API..

ryantmer: Yeah their docs are... somewhat lacking, it seems.

d2l-search-widget.html
```diff
@@ -0,0 +1,454 @@
...
+
+			d2l-dropdown {
+				min-width: 100%;
+			}
+
+			.search-bar {
+				display: flex;
+				flex-direction: row;
+				align-items: center;
+				border: 1px solid #EFEFEF;
```
ryantmer: Can this be a `--d2l-color` instead? (Since we have a wide variety of grays and all :grapes: )

d2l-search-widget.html
```diff
@@ -0,0 +1,454 @@
...
+				-moz-box-shadow: inset 0px 1px 3px 0px rgba(136,136,136,0.5);
+				box-shadow: inset 0px 1px 3px 0px rgba(136,136,136,0.5);
+			}
+
+			.search-icon {
+				flex-grow: 0;
+				background: none;
+				border: none;
+				margin: 0;
+				cursor: pointer;
```
ryantmer: Does this icon do anything? I try click, but nothing happen :( If it's not actually a link/button, probs shouldn't have the pointer.

d2l-search-widget.html
apalaniuk: @ryantmer Ah yes, that will be in an upcoming fix. There are a few cases where it could probably use `d2l-typography` etc as well.

d2l-search-widget.html
apalaniuk: @ryantmer If there's any text entered in the search box, it will act similarly to the existing My Courses search, in that it will go to the advanced search page - same as on enter.

There's a commit coming (fairly) shortly to align the search style with the existing search, which should hopefully make it clearer that it's like the existing search.

ryantmer: Ah, gotcha. As mentioned on Slack, I'm also seeing some really weird stuff, so I suspect something else is weird with my local instance.

d2l-search-listbox.html
```diff
@@ -0,0 +1,171 @@
...
+						return true;
+					}
+				});
+				return selectableItems.length > 0;
+			},
+
+			/*
+			* Focus the last selectable listbox item
+			*/
+			focusLast: function() {
```
ryantmer: (I'm not actually sure where to comment this, but I'll go with... HERE.)

Tabbing into the search area, I think the magnifying glass is getting focus, but it isn't selectable if it is (neither space nor enter do anything for that "?" stage between the menu and the text field)

d2l-search-listbox.html
```diff
@@ -0,0 +1,210 @@
...
+
+			/*
+			* Override IronMenuBehavior._focusPrevious so that the listbox owner can be focused
+			*/
+			_focusPrevious: function() {
+				var length = this.items.length;
+				var curFocusIndex = this.indexOf(this.focusedItem);
+
+				for (var i = 1; i < length + 1; i++) {
+					var newItemIndex = (curFocusIndex - i + length) % length;
```
ryantmer: `(n + x) % x == n % x ∀ n` :wink:

d2l-search-listbox.html
```diff
@@ -0,0 +1,210 @@
...
+						return;
+					}
+				}
+			},
+
+			/*
+			* Handle keyboard selection of listbox items, as IronMenuBehavior leaves it to the item to handle
+			*/
+			_onKeydown: function(e) {
+				if (e.detail.event === 'keydown' && this.keyboardEventMatchesKeys(e, 'space enter')) {
```
ryantmer: I'm certainly no expert here, but [this documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role) seems to suggest that you should either/or when it comes to space and enter. If it's a link, activate on enter, if it's a button, activate on space.

But, IANAE.

d2l-search-listbox.html
```diff
@@ -0,0 +1,210 @@
...
+			* Override IronMenuBehavior._focusedItemChanged to do nothing but focus the desired item
+			*/
+			_focusedItemChanged: function(focusedItem) {
+				if (focusedItem) {
+					focusedItem.focus();
+				}
+			},
+
+			/*
+			* Override IronMenuBehavior._focusedItemChanged to do nothing but focus the desired item
```
ryantmer: Copypasta strikes!

d2l-search-widget.html
```diff
@@ -0,0 +1,480 @@
...
+
+				if (currSearchHistoryIndex !== -1) {
+					this.splice('_previousSearches', currSearchHistoryIndex, 1);
+				}
+
+				if (this.unshift('_previousSearches', searchTerm) > 5) {
+					this.splice('_previousSearches', this._previousSearches.length - 1, 1);
+				}
+
+				window.localStorage.setItem(
```
ryantmer: This needs a `try/catch` around it (source: @omsmith), private browsing in Safari you can't do this it seems. Also for FRA situations.

d2l-search-widget.html
```diff
@@ -0,0 +1,480 @@
...
+								<option selectable data-text$="[[item.name]]" data-href$="[[item.href]]">
+									[[item.name]]
+								</option>
+							</template>
+						</d2l-search-listbox>
+					</div>
+
+					<div class="search-results-listbox-page">
+						<d2l-search-listbox id="no-results-listbox">
+							<div disabled>{{localize('noSearchResults')}}</div>
```
ryantmer: This is a multi-element suggestion (so likely for the next PR), but could a loading spinner be added here to bridge the API call gap? I just noticed that if the network is throttled, it's not clear that this is a "live" search on keystrokes.

d2l-search-listbox.html
apalaniuk: Good point. Will go with `space` as it's in line with the listbox role.

d2l-search-widget.html
```diff
@@ -0,0 +1,480 @@
...
+				DOWN: 40,
+				UP: 38,
+				ENTER: 13
+			},
+
+			/*
+			* URL for advanced search functionality (existing advanced search page from legacy My Courses widget).
+			* @type {String}
+			*/
+			_advancedSearchUrl: 'le/manageCourses/search/6606?searchTerm=',
```
ryantmer: I think this will be problematic - 6606 isn't universal across all instances and if it's not a valid org/orgUnit the link 403s. I don't have a good idea for how to construct this URL, granted :confused:

d2l-search-widget.html
```diff
@@ -0,0 +1,480 @@
...
+						this._addSearchToHistory(e.detail.item.dataset.text);
+					}
+
+					window.location.href = e.detail.item.dataset.href;
+				}
+			},
+
+			/*
+			* Indicates which dropdown context should be open.
+			* @param {Number} Page number to activate
```
ryantmer: It's a bit odd, but mind enumerating the pages in this docstring? (There are only the three, unless I'm totally misinterpreting this)

d2l-search-widget.html
```diff
@@ -0,0 +1,480 @@
...
+			_clearSearchResults: function() {
+				this._activateSearchDropdownPage(0);
+				this._searchString = '';
+				this.set('_courseSearchResults', []);
+			},
+
+			_onSearchStringChanged: function(newString) {
+				if (this._listboxes) {
+					if (newString.length > 0) {
+						this._liveSearch();
```
ryantmer: Could you add a short debounce in here? Maybe 50 or 100ms or something, just to reduce the number of somewhat-pointless API calls. This would pair well with the loading-spinner granted, so maybe do the two of those in the next PR kinda thing.

ryantmer:
Coolio - added some things, mostly minor stuff. The debounce/loading spinner I figure is a large enough chunk of work that it should be a different PR. Not sure about the advanced search URL... might be safe to keep it as-is for now, but the risk there is that if it doesn't work for a given instance, then they can't get to that page in any easy way at all... :confused:

d2l-search-listbox.html
apalaniuk: In JS, only for non-negative numbers!

ryantmer: I... completely forgot that mod is not well-defined for negative numbers.

#mathfail

d2l-search-widget.html
apalaniuk: Can just pass the orgId as another parameter for now?

ryantmer: Yeah, might be the only way to do it reliably actually. Would only be a small change within the LMS (not sure if it can be done by Friday, granted?)

d2l-search-widget.html
apalaniuk: Yeah - I was concerned about an effective debounce making the live search less responsive-feeling, but I agree, it's worthwhile to have something in there.

ryantmer: I think on its own it would seem laggy, but with a loading spinner it's a bit more clear that things are happening, the 50ms or whatever delay would be fairly unnoticeable.

PR #102, opened 2016-08-17
Fix issue where reordering of previous searches list may cause thrash…

apalaniuk:
Hm, so selecting a previous search at index two or above may cause thrashing of elements in the "Recent Searches" list, which may now cause selection events to fire, preventing navigation. Don't know why this wasn't happening last week under the same test conditions. Hooray.

Solution (for now) is to not update the previous searches array used in the dom-repeat template on new search, but rather only update the list in localStorage, and update the recent searches array on widget load.

Bah.


CodeBaboon:
Seems to fix the issue for me on Chrome (MacOS). In Firefox the search terms are still not actionable, and in Safari items are never added to the list.

d2l-search-widget.html
```diff
@@ -469,16 +480,37 @@
...
+
+			/*
+			 * Determines whether the target of a click event is a descendant of this element,
+			 * and closes the dropdown if not.
+			 */
+			_checkFocusLostClick: function(e) {
+				if (this.$.dropdown.opened && !this._isDescendant(Polymer.dom(e).rootTarget)) {
+					this.$.dropdown.close();
+				}
+			},
```
CodeBaboon: Would there be value in refactoring these focus lost functions a bit? So maybe something like:

```
_handleFocus: function(e) {
    this._checkFocusLost(this._isDescendant(document.activeElement));
}

_handleClick: function(e) {
    this._checkFocusLost(this._isDescendant(Polymer.dom(e).rootTarget));
}

_checkFocusLost(targetIsDescendant) {
    if (this.$.dropdown.opened & !targetIsDescendant) {
        this.$.dropdown.close();
    }
}
```

Keeps all the logic in one function, aside from determining the isDescendant, and also allows for other logic to be added independently to one of the event handlers.

bower.json
```diff
@@ -40,8 +40,8 @@
     "d2l-icons": "^2.4.0",
     "d2l-offscreen": "^2.1.0",
     "d2l-colors": "^2.0.0",
-    "d2l-typography": "^4.0.0",
-    "d2l-link": "^3.0.0",
+    "d2l-typography": "~4.0.0",
```
ryantmer: Hmm... I'd prefer to fix these in the problem repos tbh. I'll look into that, and update here if I do so. (Otherwise, we'll just have to remember to change them back to `^`)

ryantmer: Looks like it'll be a few PRs, but quite small ones. Going to start on that this afternoon, shouldn't take too long.

apalaniuk: As would I! Mainly a case of not having access to those repos.

ryantmer: Wait really? O_o That's... weird. Oh, are you adding the `-ui`? (e.g. it's `d2l-typography-ui`, not just `d2l-typography`)

CodeBaboon:
So this fixes up the FireFox stuff as indicated by https://github.com/Brightspace/d2l-my-courses-ui/pull/102/commits/ff3f142ca2488f78bfb9b1a383c794dd27402843.

Safari is still broken for me completely, in the sense that the previous searches are not displayed. I investigated and it is purely a display issue, so not related to this PR in any way. I have logged this as #104 .

ryantmer:
Hokay, fixed the deps problem, should be good to just use ^ now.

d2l-search-listbox.html
```diff
@@ -160,7 +162,7 @@
 			* Handle keyboard selection of listbox items, as IronMenuBehavior leaves it to the item to handle
 			*/
 			_onKeydown: function(e) {
-				if (e.detail.event === 'keydown' && this.keyboardEventMatchesKeys(e, 'space enter')) {
+				if (e.detail.event === 'keydown' && this.keyboardEventMatchesKeys(e, 'space')) {
```
ryantmer: :+1:

ryantmer:
Poked around with this - seems to make sense to load recent searches on page load anyway, LGTM! :shipit:

ryantmer:
...having said that, just looked at @CodeBaboon's comment and #105. That's a separate thing, though, I'd imagine? (Considering it exists with and without this PR)

CodeBaboon:
Yeah my PR is separate. This one LGTM.

PR #103, opened 2016-08-17
US74383 - Add load delay

ryantmer:
Kind of a WIP (haven't added tests yet), but works surprisingly well! The `d2l-course-tile` now supports a `delay-load` attribute, which will prevent it from fetching the organization when it's first rendered. Since we fetch a bunch of enrollments, but only display the pinned ones, this means we can save a lot of network chattiness at page load - possibly even avoid it entirely, if the user doesn't open the all courses view (e.g. an admin, with many courses, who may instead just go to Last Accessed, or do a search).

In testing this out with various network throttling settings in Chrome, only on GPRS does this give a noticeable delay to loading the course names/images when you open the overlay. Granted, our home page takes like 90 seconds on that sort of connection, so... yeah. Any other throttling, the loading is "near-instantaneous", partially thanks to the animation delay of opening the overlay.

The use case where this doesn't help would be somebody with a lot of pinned enrollments, since we'd still be fetching organizations for all of them. However, this doesn't make that case any worse, and we had discussed limiting the number of pinned enrollments anyway.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -36,14 +36,16 @@
 
 		<d2l-course-tile-grid
 			enrollments="{{pinnedEnrollments}}"
-			enrollments-queue="{{pinnedEnrollmentsQueue}}">
+			enrollments-queue="{{pinnedEnrollmentsQueue}}"
+			delay-course-tile-load="true">
 		</d2l-course-tile-grid>
 
 		<h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 
 		<d2l-course-tile-grid
 			enrollments="{{unpinnedEnrollments}}"
-			enrollments-queue="{{unpinnedEnrollmentsQueue}}">
+			enrollments-queue="{{unpinnedEnrollmentsQueue}}"
+			delay-course-tile-load="true">
 		</d2l-course-tile-grid>
 	</template>
 
@@ -92,6 +94,11 @@
 				var itemCount = Math.max(this.pinnedEnrollments.length, this.unpinnedEnrollments.length);
 
 				return itemCount;
+			},
+			loadCourseTiles: function() {
+				Polymer.dom(this.root).querySelectorAll('d2l-course-tile-grid').forEach(function(grid) {
+					grid.delayCourseTileLoad = false;
+				});
 			}
 		});
 	</script>
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -37,7 +37,7 @@
 		<d2l-course-tile-grid
 			enrollments="{{pinnedEnrollments}}"
 			enrollments-queue="{{pinnedEnrollmentsQueue}}"
-			delay-course-tile-load="true">
+			delay-course-tile-load="[[delayCourseTileLoad]]">
 		</d2l-course-tile-grid>
 
 		<h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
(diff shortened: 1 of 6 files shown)


ryantmer:
Hokay, tests added, and morning/day-after me still thinks this makes sense. Travis failure is the version mismatch thing, which I'll be addressing this afternoon.

d2l-course-tile.html
```diff
@@ -205,6 +196,21 @@
 					this.toggleClass('animate-insertion-pre', true, this);
 				}
 			},
+			ready: function() {
+				// Currently only needed in tests, but it is entirely possible
+				// to render the element without an enrollment initially
```
apalaniuk: Are there cases (other than testing) where we'd actually render the element without an enrollment? If so, I think `_organizationUrl` wouldn't be correctly set causing an issue if`delayLoad` is set to `false` afterward?

I also wonder if this makes using a Polymer template filter/sort more difficult in the future by setting properties in a lifecycle callback (or the result of an API call done within one) due to Polymer reusing elements and only copying properties on those actions (not that I can imagine us using them for anything in the near future). Is there any reason the old '_enrollmentChanges` can't check the input to avoid the issue where the observer is fired on element creation with a null value? Maybe due to some awesome new changes you alluded to that I'm unaware of? :smiley:

d2l-all-courses.html
```diff
@@ -92,6 +94,11 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				var itemCount = Math.max(this.pinnedEnrollments.length, this.unpinnedEnrollments.length);
 
 				return itemCount;
+			},
+			loadCourseTiles: function() {
+				Polymer.dom(this.root).querySelectorAll('d2l-course-tile-grid').forEach(function(grid) {
```
apalaniuk: Could `delayCourseTileLoad` be made a property on `d2l-all-courses` bound to child elements, then simply set to `false` here (unless it isn't something you want to expose)? It's probably just a side-step unless you start adding new web component types with a similar property, so more just a thought than anything.

ryantmer: Hm, I feel like Wednesday-night me had a reason for only going this deep, but Friday-night me doesn't know what it was. I do like that idea better. Leverage dat Polymer.

d2l-course-tile.html
ryantmer: To the first question, I'd say "no" - can't think of a circumstance where we'd render it without. My initial thought was that that seems anti-HTML-y, but thinking about it, `<a>` requires an `href` attribute to be valid, so it's totally fine for us to say that it won't work without an enrollment (I think).

Second paragraph response: you speak words, and they make sense. My main motivator for moving the organization-fetching logic to `ready` was just that it seemed more idiomatic, under the assumption that you would always have an enrollment on `ready` anyway. However, with the above conclusion, no longer the case, so I'll go back to `_enrollmentChanged`, with the extra check to make sure `delayLoad == false`.

Third paragraph response: yes. Yes it should. I'll look into that as well! (granted possibly in a different PR :rocket:)

ryantmer: Ah, I remember the other reason - meant the organization wouldn't be re-fetched when a pin/unpin happened. However, that can still be avoided (or rather, it already was being avoided, just with more words)

ryantmer: NOPE. Re-found my reasoning. Derp.

When `_enrollmentChanged` happens, there's no guarantee that `delayLoad` has been set yet, so is sometimes undefined, and so the course tile fetches the organization. This can be avoided by adding a default value, however some times we want to be true, some false, so that won't work particularly well. Alternatively, could just check to make sure `delayLoad` is explicitly false, suppose that would work too.

apalaniuk: @ryantmer I think you'd have to use a multi-property observer. But, I think this is nice and straightforward. :+1:

ryantmer: Hm. Yeah, I like the simplicity I think - especially since the two properties are fairly distinct in this case, so a multi-property observer sounds a bit odd tbh =/

PR #105, opened 2016-08-18
replace option element with div and role

### LINKED ISSUE #104: Safari does not display previous search terms
In Chrome or Firefox when I click into the search list box I see a list of previously searched terms:

recent-searches-chrome

However in Safari the terms do not appear:

recent-searches-safari

Looking into the source, the terms are being stored/retrieved and inserted into the document, so that is not an issue. It appears to be related to how Safari handles the `<option>` element. Note that in Chrome the user agent stylesheet sets `display: block`, as seen here:

option-element-chrome

However in Safari not only does the user agent stylesheet not set the display, the browser actively refuses to respect the setting that is being applied by the Daylight CSS (note that it is lightly greyed out and struck-through):

option-element-safari

Even if I apply `display: block` directly to the element Safari will strike it out and not show the element.

CodeBaboon:
This is to fix #104.

### DIFF
--- a/d2l-search-listbox.html
+++ b/d2l-search-listbox.html
@@ -92,7 +92,9 @@
 			*/
 			hasItems: function() {
 				var selectableItems = this.items.filter( function(item) {
-					if (item.tagName === 'OPTION' && !item.hasAttribute('disabled')) {
+					if (item.hasAttribute('role') &&
+						item.getAttribute('role') === 'option' &&
+						!item.hasAttribute('disabled')) {
 						return true;
 					}
 				});
--- a/d2l-search-widget.html
+++ b/d2l-search-widget.html
@@ -145,9 +145,9 @@
 						<d2l-search-listbox>
 							<div class="d2l-heading-4" data-list-title disabled>{{localize('recentSearches')}}</div>
 							<template is="dom-repeat" items="[[_previousSearches]]">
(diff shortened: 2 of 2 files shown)


apalaniuk:
Ah, I just put this in my other PR as well.

CodeBaboon:
@apalaniuk  We can kill this one if you prefer, doesn't matter to me.

apalaniuk:
Nah it's fine, this will be less confusing. I'll revert my change since it conflicts.

Thanks!

ryantmer:
In FF on Windows, this PR does fix the formatting (padding, mostly) of the recent searches, but I still can't up/down to select things. I presume that's a separate issue, though - if so, then LGTM!

CodeBaboon:
So this PR is actually breaking the up/down to select things on Chrome. I did not know that was a thing. So we can either address that whole issue across all browsers here before we break it on Chrome, or we can break it and then fix it for everyone in a different PR. Thoughts?

ryantmer:
I'm in the "fix first" camp, otherwise we'd be looking at a regression :cry:

apalaniuk:
@CodeBaboon  https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-search-listbox.html#L95

apalaniuk:
:+1:

PR #106, opened 2016-08-22
US75822 - Use sortBy=pinDate for default enrollments call

ryantmer:
This API change allows us to always get pinned courses, based off their pinDate, from the API. This solves the problem where an alphabetically-low pinned course wouldn't come back from the API, since it was ordering them by name and only grabbing the pinned ones.

(Note that I only just merged the API update today, so you'll have to pull LP for this to work properly.)


ryantmer:
@apalaniuk? :D

apalaniuk:
LGTM!

PR #111, opened 2016-09-01
Add lazy loading of enrollments

apalaniuk:
`iron-scroll-threshold` is a simple solution for getting this working, but (as you might see from the red X), [Polylint has a bug](https://github.com/Polymer/polymer-analyzer/issues/205) where it will break if an element (in this case, `iron-scroll-threshold`) uses a property defined on `Polymer.Base` (in this case, `isAttached`), which is perfectly valid. You can't configure or suppress errors, nor can you whitelist files/properties/etc.

Possible solutions:

1) Turn off Polylint (not ideal, but the Polymer core team seems to do this, [not only on `iron-scroll-threshold`](https://github.com/PolymerElements/iron-scroll-threshold/commit/35a555136e29c684dbe0dad2a6a2f06a7e67b097) but [ elsewhere as well ](https://github.com/PolymerElements/app-layout/commit/be2bf470e292b094fb160dc23ff2c0fb2904d32a))
2) Fork `iron-scroll-threshold` and change references to `isAttached` to refer to a property defined within that element (so long as it's possible without changing behavior)
3) Write our own behavior

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,6 +27,7 @@
     "/demo/"
   ],
   "dependencies": {
+    "polymer": "Polymer/polymer#^1.4.0",
     "app-localize-behavior": "https://github.com/Brightspace/app-localize-behavior.git#master",
     "d2l-ajax": "^2.0.0",
     "d2l-colors": "^2.0.0",
@@ -47,7 +48,8 @@
     "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.7.1",
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
-    "polymer": "Polymer/polymer#^1.4.0"
+    "iron-a11y-keys": "^1.0.6",
+    "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2"
   },
   "devDependencies": {
     "web-component-tester": "^4.2.2"
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,6 @@
     "/demo/"
   ],
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.4.0",
     "app-localize-behavior": "https://github.com/Brightspace/app-localize-behavior.git#master",
     "d2l-ajax": "^2.0.0",
     "d2l-colors": "^2.0.0",
@@ -48,7 +47,8 @@
     "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.7.1",
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
-    "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2"
+    "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
+    "polymer": "Polymer/polymer#^1.4.0"
   },
   "devDependencies": {
     "web-component-tester": "^4.2.2"
--- a/bower.json
+++ b/bower.json
@@ -48,7 +48,6 @@
     "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.7.1",
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
-    "iron-a11y-keys": "^1.0.6",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2"
   },
   "devDependencies": {
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -181,7 +181,12 @@
 				* Object containing the last response from an enrollments search request
 				* @type Object
 				*/
-				_lastEnrollmentsSearchResponse: Object
+				_lastEnrollmentsSearchResponse: Object,
+				/*
+				* Object containing the IDs of previously loaded pinned enrollments, to avoid duplicates
+				* @type Object
+				*/
+				_pinnedCoursesMap: Object
 			},
 			behaviors: [
 				Polymer.D2L.MyCourses.CourseTileResponsiveGridBehavior,
@@ -193,6 +198,7 @@
 				'open-change-image-view': '_openChangeImageView'
 			},
 			ready: function() {
+				this._pinnedCoursesMap = {};
 				this._alertMessage = this.localize('noCoursesMessage');
 
 				this.toggleClass('hidden', true, this.$.courseTiles);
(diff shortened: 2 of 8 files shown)


d2l-my-courses.html
```diff
@@ -186,6 +197,8 @@
 				if (!this.delayLoad) {
 					this.$.enrollmentsRootRequest.generateRequest();
 				}
+
+				this.$$('iron-scroll-threshold').scrollTarget = this.$$('d2l-simple-overlay').scrollRegion;
```
jstefaniuk-d2l: Now that there's 2 overlays here, you may want to use a more specific selector

ryantmer:
Given those options, I'd go with 1, at least for now. Polylint is nice, but doesn't provide _huge_ value I don't think, and the other two are much higher overhead/work options... so for now, I guess no polylint would make sense?

d2l-my-courses.html
```diff
@@ -223,19 +236,14 @@
 					var enrollments = parser.parse(response.detail.xhr.response);
 					var enrollmentEntities = enrollments.getSubEntitiesByClass('enrollment');
 
-					var pinnedEnrollments = [];
-					var unpinnedEnrollments = [];
-
 					enrollmentEntities.forEach(function(enrollment) {
 						if (enrollment.hasClass('pinned')) {
-							pinnedEnrollments.push(enrollment);
+							this.push('pinnedEnrollments', enrollment);
```
ryantmer: Nice, likey. Does this cause any issues with the change events firing this often, though?

apalaniuk: Not in practice, but I think you're right - I was originally thinking of pushing directly and using `notifyPath` afterward, which is just a better idea. :+1:

ryantmer: Oh damn, forgot about `notifyPath`. Yeah might be a better approach. (Not that this is probably causing any real problems, but hey... might as well and all that.)

bower.json
```diff
@@ -47,7 +47,8 @@
     "iron-input": "^1.0.10",
     "iron-pages": "^1.0.8",
     "d2l-dropdown": "^2.1.0",
-    "iron-a11y-keys": "^1.0.6"
+    "iron-a11y-keys": "^1.0.6",
+    "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2"
```
ryantmer: Isn't this available on bower itself? :o

d2l-my-courses.html
```diff
@@ -108,11 +114,20 @@
...
+
 			<d2l-all-courses
 				pinned-enrollments="{{pinnedEnrollments}}"
 				unpinned-enrollments="{{unpinnedEnrollments}}"
 				locale="[[locale]]">
 			</d2l-all-courses>
+			<d2l-loading-spinner
+				id="lazyLoadSpinner"
+				class="hidden"
+				size="150">
```
ryantmer: Should this not be the same size as the other one? Or does the design specify it being larger?

d2l-my-courses.html
```diff
@@ -222,20 +244,36 @@
...
-							unpinnedEnrollments.push(enrollment);
+							newUnpinnedEnrollments.push(enrollment);
 						}
-					});
+					}, this);
+
+					this.pinnedEnrollments = this.pinnedEnrollments.concat(newPinnedEnrollments);
+					this.unpinnedEnrollments = this.unpinnedEnrollments.concat(newUnpinnedEnrollments);
+
+					this.notifySplices('pinnedEnrollments', [{
```
ryantmer: Damn, that's cool.

d2l-simple-overlay.html
```diff
@@ -182,6 +182,9 @@ <h1 class="d2l-heading-2">{{title}}</h1>
 				var style = document.createElement('style', 'custom-style');
 				style.innerHTML = 'iron-overlay-backdrop { --iron-overlay-backdrop-opacity: 0.0; }';
 				document.body.appendChild(style);
+			},
+			get scrollRegion() {
+				return this.$$('.scrollable');
```
ryantmer: Is `scrollable` a class inside `iron-overlay`?

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -145,6 +145,17 @@ describe('d2l-my-courses', function() {
 	});
 
 	describe('Enrollments requests and responses', function() {
+		beforeEach( function() {
+			server.respondWith(
```
ryantmer: Doesn't this conflict with the request in the first test here, which is returning a 404 instead?

d2l-my-courses.html
apalaniuk: @ryantmer There's no design per se, minus that it exists, but I'll run it by Bob. I found 100px was a bit small in this context (but designed, I am not).

ryantmer: Haha, me neither, that's fo sho. But if it looks right, then all good!

bower.json
apalaniuk: @ryantmer Surprisingly, no. Perhaps because someone else claimed the name when porting `core` to `iron` components for their own use (and a `core` entry doesn't exist either).

ryantmer: Aw boo :( Ah well.

d2l-simple-overlay.html
apalaniuk: @ryantmer [It's a class used within this element](https://github.com/Brightspace/d2l-my-courses-ui/blob/492f51156e77e39289489acc852f2086d637c175/d2l-simple-overlay.html#L32) so it should be searchable? But I'm totally open to any changes I might be missing here.

ryantmer: Derp. Nope, that's me not looking hard enough, carry on! :+1:

test/d2l-my-courses/d2l-my-courses.js
apalaniuk: @ryantmer [Current behaviour](https://github.com/sinonjs/sinon/issues/498) (though not explicitly stated in the docs AFAIK) is that it responds with the most recently added match, but I am totally fine with reverting this to make this more explicit.

ryantmer: Herm, yeah I prefer explicit myself :x OTOH, I like that you've factored out this mock, so maybe just move the 404 into its own `describe('errors and whatnot')` scope or something?

ryantmer:
Did some testing and whatnot, works super well!

It seemed like the spinner sometimes wasn't loading, but it turns out it was, just that the network call was so fast that it was immediately being hidden. Yay for our APIs working better than expected? :P

ryantmer:
Erp, actually - there appears to be a bug. If a pinned course is then loaded by the load-more process, it appears twice when the overlay is closed. (e.g. if I pin "Biology", then open the all courses overlay, then scroll until Biology shows up in the list, then close the overlay, I see "Biology" twice in the list of pinned courses). I'm thinking some wires get crossed and the end up being added to the list of pinned courses twice?

ryantmer:
Looks like that fixed it, nice :)

Just noticed something else a bit odd though - when you scroll to load new organizations, there's an extra "/d2l/home" request going for each new organization that's being loaded... I'm thinking it's the iron component doing this? Either way, it's a pretty heavy request, so wouldn't want it to be doing that :(

apalaniuk:
@ryantmer Just because the `code-review` tag was dropped already - is this good to merge, or any other comments/etc?

ryantmer:
Checked it out again, yep, LGTM! Yay we finally have lazy loading :D

PR #113, opened 2016-09-06
Update to latest version of d2l-dropdown.

dbatiste:


d2l-search-dropdown-content.html
```diff
@@ -35,10 +35,6 @@
 		Polymer({
 			is: 'd2l-search-dropdown-content',
 
-			ready: function() {
```
dbatiste: This attribute is obsolete with the newest version of `d2l-dropdown`.  The border is provided by an element not implemented in this use of the content behavior.

ryantmer:
:+1:

PR #116, opened 2016-09-08
US76123 - Filter and Sort UI

ryantmer:
The filtering and sorting here applies to the values that are shown in the live search dropdown - since we elected to move advanced searching to the existing advanced search page, the UI won't actually update when you select a new filter/sort. To check it out, apply filters/sort, then type in the search field. It's not intended to be like this permanently, but updating the course tiles with your search/filters/sorting is a future thing.

TODO:

- [x] Fix search when canonical namespace feature is enabled
- [x] Prevent dropdown menu from moving around when selecting/unselecting filters
- [x] Fix dropdowns to use buttons rather than icons
- [x] Localize
- [x] Match UI to designs

Things that were TODO but are no longer TODO because reasons:

- ~~Update search when filters/sort order change~~ - Doesn't work for live search approach, but this will have to be done when updating tiles
- ~~Display results as course tiles, not just in dropdown/live search box~~ Separate PR; this will be a fairly large piece of work
- ~~Add search box in filters when required~~ Separate PR; during the leap sprint I'm going to break out the search widget into its own repo, which will mucho-facilitate this (see TA104610)
- ~~Mobile view~~ Separate PR; pseudo-blocked on the breaking out of the search widget as well (see TA104609)
- ~~`box-shadow` above/below dropdown menu contents~~ See https://github.com/Brightspace/d2l-dropdown-ui/issues/76

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -451,24 +451,18 @@
 				}
 			},
 			_selectSemesterList: function() {
-				this.toggleClass('hidden', false, this.$.semesterList);
-				this.toggleClass('hidden', true, this.$.departmentList);
-				this.toggleClass('dropdown-content-tab-button-selected', true, this.$.semesterListButton);
-				this.toggleClass('dropdown-content-tab-button-selected', false, this.$.departmentListButton);
-				this.toggleClass('invisible', false, this.$.semesterListHighlight);
-				this.toggleClass('invisible', true, this.$.departmentListHighlight);
-				this.$.semesterListButton.setAttribute('aria-pressed', true);
-				this.$.departmentListButton.setAttribute('aria-pressed', false);
+				this._toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, true);
+				this._toggleSelectedList(this.$.departmentList, this.$.departmentListButton, this.$.departmentListHighlight, false);
 			},
 			_selectDepartmentList: function() {
-				this.toggleClass('hidden', true, this.$.semesterList);
-				this.toggleClass('hidden', false, this.$.departmentList);
-				this.toggleClass('dropdown-content-tab-button-selected', false, this.$.semesterListButton);
-				this.toggleClass('dropdown-content-tab-button-selected', true, this.$.departmentListButton);
-				this.toggleClass('invisible', true, this.$.semesterListHighlight);
-				this.toggleClass('invisible', false, this.$.departmentListHighlight);
-				this.$.semesterListButton.setAttribute('aria-pressed', false);
-				this.$.departmentListButton.setAttribute('aria-pressed', true);
+				this._toggleSelectedList(this.$.departmentList, this.$.departmentListButton, this.$.departmentListHighlight, true);
+				this._toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, false);
+			},
+			_toggleSelectedList(list, button, highlight, selected) {
+				this.toggleClass('hidden', !selected, list);
+				this.toggleClass('dropdown-content-tab-button-selected', selected, button);
+				this.toggleClass('invisible', !selected, highlight);
+				button.setAttribute('aria-pressed', selected);
 			},
 			_toggleFilterDropdown: function() {
 				this.$.filterDropdown.toggleOpen();
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -1,7 +1,6 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown.html">
-<link rel="import" href="../d2l-dropdown/d2l-dropdown-button.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown-content.html">
 <link rel="import" href="../d2l-icons/d2l-icons.html">
 <link rel="import" href="../d2l-menu/d2l-menu.html">
@@ -132,9 +131,9 @@
 				visibility: visible;
 				width: 1rem;
 				height: 1rem;
-				margin-top: -0.2rem;
+				margin-top: -0.1rem;
 				margin-right: 0.5rem;
-				margin-left: -0.5rem;
+				margin-left: -0.2rem;
 				border-radius: 0.3rem;
 				border-style: solid;
 				border-color: var(--d2l-color-titanius);
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -43,6 +43,7 @@
 				border: none;
 				padding: 0;
 				cursor: pointer;
+				margin-left: 30px;
 			}
 			.filter-item {
 				display: block;
@@ -56,9 +57,6 @@
 			}
 			.search-and-filter > d2l-search-widget {
 				width: 50%;
-			}
-			.search-and-filter d2l-dropdown {
-				margin-left: 30px;
 			}
 			.search-and-filter .dropdown-icon {
 				margin-left: 12px;
@@ -135,11 +133,9 @@
 			</d2l-search-widget>
 
 			<div>
-				<d2l-dropdown id="filterDropdown" no-auto-open>
-					<div class="d2l-dropdown-opener">
-						<button class="d2l-dropdown-opener filter-button">Filter</button>
-						<iron-icon class="dropdown-icon" icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
-					</div>
+				<button class="filter-button" on-tap="_toggleFilterDropdown">Filter</button>
+				<d2l-dropdown id="filterDropdown">
+					<iron-icon class="d2l-dropdown-opener dropdown-icon" icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
 					<d2l-dropdown-content>
 						<div class="filter-and-clear">
 							<span>Filter by</span>
@@ -192,11 +188,9 @@
 					</d2l-dropdown-content>
 				</d2l-dropdown>
 
+				<button class="filter-button" on-tap="_toggleSortDropdown">Sort: Date Pinned</button>
 				<d2l-dropdown id="sortDropdown">
-					<div class="d2l-dropdown-opener">
-						<button class="d2l-dropdown-opener filter-button">Sort: Date Pinned</button>
-						<iron-icon class="dropdown-icon" icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
-					</div>
+					<iron-icon class="d2l-dropdown-opener dropdown-icon" icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
 					<d2l-dropdown-content>
 						Sort by
 						<button class="sort-button sort-button-selected" value="pinDate" on-tap="_updateSortBy">Date Pinned</a>
@@ -285,12 +279,9 @@
 				for (var i = 0; i < courseTileGrids.length; i++) {
 					courseTileGrids[i].getCourseTileItemCount = this.getCourseTileItemCount.bind(this);
 				}
-			},
-			attached: function() {
-				this.listen(this.$$('d2l-dropdown'), 'click', '_onFilterDropdownClick');
-			},
-			detached: function() {
-				this.unlisten(this.$$('d2l-dropdown'), 'click', '_onFilterDropdownClick');
+
+				this.$.semestersRequest.generateRequest();
+				this.$.departmentsRequest.generateRequest();
 			},
 			/*
 			* Fetches set of focusable elements (all `<d2l-course-tile>`s)
@@ -328,27 +319,13 @@
 				});
 				this._isFiltered = false;
 				this._parentOrganizationIds = [];
-				this.$$('.filter-button').innerHTML = 'Filter';
-			},
-			_onFilterDropdownClick: function() {
-				if (!this._semesterOrganizations) {
-					this.$.semestersRequest.generateRequest();
-					this.$.departmentsRequest.generateRequest();
-				}
+				this.$$('.filter-button').textContent = 'Filter';
 			},
 			_onDepartmentsResponse: function(response) {
 				if (response.detail.status === 200) {
 					var parser = document.createElement('d2l-siren-parser');
 					var departments = parser.parse(response.detail.xhr.response);
 					this._departmentOrganizations = departments.entities;
-
-					if (this._semesterOrganizations) {
-						// We delay the opening of the menu until we have a response, so that the
-						// dropdown is sized correctly. Once we have the semesters and departments,
-						// we can just immediately open it on future clicks
-						this.$.filterDropdown.removeAttribute('no-auto-open', false);
-						this.$.filterDropdown.toggleOpen();
-					}
 				}
 			},
 			_onSemestersResponse: function(response) {
@@ -356,14 +333,6 @@
 					var parser = document.createElement('d2l-siren-parser');
 					var semesters = parser.parse(response.detail.xhr.response);
 					this._semesterOrganizations = semesters.entities;
-
-					if (this._departmentOrganizations) {
-						// We delay the opening of the menu until we have a response, so that the
-						// dropdown is sized correctly. Once we have the semesters and departments,
-						// we can just immediately open it on future clicks
-						this.$.filterDropdown.removeAttribute('no-auto-open', false);
-						this.$.filterDropdown.toggleOpen();
-					}
 				}
 			},
 			_selectSemesterList: function() {
@@ -382,6 +351,12 @@
 				this.$.semesterListButton.setAttribute('aria-pressed', false);
 				this.$.departmentListButton.setAttribute('aria-pressed', true);
 			},
+			_toggleFilterDropdown: function() {
+				this.$.filterDropdown.toggleOpen();
+			},
+			_toggleSortDropdown: function() {
+				this.$.sortDropdown.toggleOpen();
+			},
 			_updateFilter: function(e) {
 				if (e.target.checked) {
 					this.push('_parentOrganizationIds', e.target.value);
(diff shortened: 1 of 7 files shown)


d2l-all-courses.html
```diff
@@ -21,19 +28,183 @@
...
+			.filter-button {
+				@apply(--d2l-small-text);
+				/* overrides to d2l-small-text */
+				font-size: 1rem;
+				color: var(--d2l-color-ferrite);
+				/* end overrides */
+				background: none;
+				border: none;
+				padding: 0;
+				cursor: pointer;
```
CodeBaboon: the cursor should probably be set on the `drop-down-opener` so that it appears when you are on top of the chevron thing in addition to the word.

ryantmer: Ooh, legit yeah.

CodeBaboon: Not sure where to put this so I'll add it here: when I click on the Filter word (button) and open the dropdown, clicking again on the Filter word (button) closes the dropdown. But if I click on the down-chevron it closes and re-opens the dropdown.

d2l-all-courses.html
```diff
@@ -87,6 +286,12 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 					courseTileGrids[i].getCourseTileItemCount = this.getCourseTileItemCount.bind(this);
 				}
 			},
+			attached: function() {
```
jstefaniuk-d2l: Will this pick up Trevor's dropdowns?

trevorjonesd2l: Maybe look for a specific `id`

d2l-all-courses.html
apalaniuk: @CodeBaboon @ryantmer I'm pretty sure this is due to both the containing div and the sort/filter buttons having `d2l-dropdown-opener` classes. I tried removing the div and making the `iron-icon` a child of the button, and it seems to work correctly.

d2l-all-courses.html
```diff
@@ -116,6 +321,96 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
 			loadCourseTiles: function() {
 				this.delayCourseTileLoad = false;
+			},
+			_clearFilters: function() {
+				Polymer.dom(this.root).querySelectorAll('.filter-item input').forEach(function(checkbox) {
+					checkbox.checked = false;
+				});
+				this._isFiltered = false;
+				this._parentOrganizationIds = [];
+				this.$$('.filter-button').innerHTML = 'Filter';
```
jstefaniuk-d2l: Is it better to use "textContent" for stuff like this?

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent

Differences from innerHTML

"innerHTML returns the HTML as its name indicates. Quite often, in order to retrieve or write text within an element, people use innerHTML. textContent should be used instead. Because the text is not parsed as HTML, it's likely to have better performance. Moreover, this avoids an XSS attack vector."

d2l-search-widget.html
```diff
@@ -400,38 +431,56 @@
...
+						// Needs to be done in a closure since XHR is async
+						(function(index, self) {
+							var enrollment = enrollmentEntities[i];
+							var xhr = new XMLHttpRequest();
+							self.push('_organizationRequests', xhr);
+							xhr.open('GET', enrollment.getLinkByRel(/\/organization$/).href, true);
+							xhr.onload = function() {
+								if (xhr.readyState === 4) {
+									if (xhr.status === 200 || xhr.status === 304) {
+										var organizationParser = document.createElement('d2l-siren-parser');
```
jstefaniuk-d2l: Should we be making a different parser for each iteration of the for loop? Or just 1 beforehand?

trevorjonesd2l:
Might keep a running comment of general issue like things I notice, feel free to also edit to point out whether they're legit or not

- Clicking anywhere in the dropdown other than the checkbox closes it - doesn't seem like expected behavior

d2l-search-widget.html
ryantmer: @CodeBaboon has actually discovered that this super doesn't work with canonical namespaces, so I'm in the process of changing it around anyway :)

ryantmer: Fixed - no longer creating multiple siren parser object, although still do have to create a d2l-ajax (formerly bare XHR) for each request, granted. But, slightly better!

ryantmer:
Addressed all the in-line comments in 47aa2b6.

To address @trevorjonesd2l's list, the first one is definitely wrong, but using a dropdown here is a janky anyway - it'll likely require a bit of a rethink. Checkboxes in a dropdown is a fairly non-standard thing, so I'd rather avoid spending too much time on that until I figure out what would be better to use instead.

The second point is... totally valid. I'll add that to the list of things for the follow up, though - how search results are displayed is also a bit weird (one would probably expect them in the full view, not just the text dropdown), so that'll require a larger change.

Finally, the last issue was something that was lost in a rebase looks like, but I've re-added the appropriate property and all is good now! If somebody wants to give another quick look around to make sure I haven't broken anything by fixing things that would be awesome; after that I'll release this and BSI and merge the BSI PR to LP :D

d2l-all-courses.html
```diff
@@ -21,19 +28,177 @@
...
+			.search-and-filter > d2l-search-widget {
+				width: 50%;
+			}
+			.search-and-filter .dropdown-icon {
+				margin-left: 12px;
+			}
+			.view-button {
+				@apply(--d2l-small-text);
+				/* overrides to d2l-small-text */
+				color: var(--d2l-color-ferrite);
```
dbatiste: Note: `--d2l-small-text` already specifies the ferrite color [here](https://github.com/Brightspace/d2l-typography-ui/blob/master/d2l-typography-shared-styles.html#L9).

d2l-all-courses.html
```diff
@@ -21,19 +28,177 @@
...
+				hidden$="[[_hasEnrollments]]"
+				search-root-url="[[searchUrl]]"
+				parent-organization-ids="[[_parentOrganizationIds]]"
+				sort-field="[[_sortField]]">
+			</d2l-search-widget>
+
+			<div>
+				<button class="filter-button" on-tap="_toggleFilterDropdown">Filter</button>
+				<d2l-dropdown id="filterDropdown">
+					<iron-icon class="d2l-dropdown-opener dropdown-icon" icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
```
dbatiste: The `d2l-dropdown-opener` class denotes the opener element for wire-up.  If the `escape` key is pressed in the context of `d2l-dropdown-content`, the dropdown is closed and focus is placed back on the opener.  In this case, I think it would try to focus on `<iron-icon>`, and if it's not focusable I _think_ the `body` would become the `activeElement`.  Even if the icon could receive focus (which is incorrectly the case in IE for `svg`), it seems this would get ignored due to `aria-hidden`.

Current opener implementation is that when the opener is activated when the dropdown is closed it will open, and if activated while open is will close.  It seems as though we're trying to achieve some different opener behavior here, yes?

d2l-all-courses.html
```diff
@@ -116,6 +312,80 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
 			loadCourseTiles: function() {
 				this.delayCourseTileLoad = false;
+			},
+			_clearFilters: function() {
+				Polymer.dom(this.root).querySelectorAll('.filter-item input').forEach(function(checkbox) {
+					checkbox.checked = false;
+				});
+				this._isFiltered = false;
+				this._parentOrganizationIds = [];
+				this.$$('.filter-button').textContent = 'Filter';
```
dbatiste: Needs to be localized?

d2l-all-courses.html
ryantmer: Slightly, yeah - the reasoning for this was that the text for the dropdown changes (e.g. when you've selected 3 filters, it changes from "Filter" to "Filter: 3 Filters"). When the length of the text changes, the drop-down moves (as it appears to centre on text), which is jarring. The thought here was that the chevron icon doesn't move, so anchoring it to that would prevent it from moving when the text changes.

Maybe a 0px button + making the text a label for said button would be a better approach?

d2l-all-courses.html
ryantmer: Yeah - it turns out this PR looks like it actually doesn't have to be in for 10.6.6 after all! Means I should be able to do all the to-dos in the description, which includes localization :)

d2l-all-courses.html
dbatiste: Yeah - the position behavior for the dropdown is: it tries to center the content on the opener.  It also tries to center the pointer on the opener.  If the content cannot be centered due to proximity to viewport edge it will shift the dropdown which ever way it needs so it will fit in viewport.  I think when I originally saw this code the opener was 100% width, and that positioning was not evident as a result.  I'm not completely up on all the changes made here... so not sure.  :)

The 0px button might work.  I think in the past I've seen 0 dimension elements not get focus.  I think I saw this in Chrome quite a while ago - might not be that way anymore.

Another alternative might be to set tabindex="-1" on the `<iron-icon>`, wire-up to its focus event, and when it gets focus (as a result of `escape`), redirect focus to the actual opener so that focus go to an unexpected place.

ryantmer: Oh, alternatively, would surrounding the `iron-icon` in a `button` make sense? That seems more semantically/a11y-correct since then it should act like a button for screen readers/focus. Then the text, while clickable, wouldn't be focusable, and the button would have an `aria-label` that matches the text?

dbatiste: Yep.  That sounds good.

ryantmer: Also I think means less work for me, and I'm all for that :stuck_out_tongue_winking_eye:

ryantmer:
Yay, things work!

I have a couple tests I want to add (this is mostly UI stuff, but there's a couple behavioural verifications I want to do), but beyond that, this is ready for review.

There are a few things that are not done yet, which I have listed in the description, along with reasons I'm not doing them yet. Hopefully these make sense/are valid - if not, I am open to discuss! (They're all "to be done", just kinda blocked on each other, and also kinda blocked on something else that I was planning to do this week anyway).

ryantmer:
Erm, note that this doesn't include the mobile view. I'll make an explicit note of that, but that is in the same category as search - it will be a lot less work/wasted code if I can move that out first, then go from there.

ryantmer:
Another follow-up: the automated tests I wanted to write for some test cases are actually more applicable elsewhere, so are also better suited to be written after I've done the breaking-out-of-search stuff. Future PR!

d2l-all-courses.html
```diff
@@ -1,4 +1,11 @@
 <link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../d2l-ajax/d2l-ajax.html">
+<link rel="import" href="../d2l-dropdown/d2l-dropdown.html">
+<link rel="import" href="../d2l-dropdown/d2l-dropdown-button.html">
```
apalaniuk: Unused?

d2l-search-widget.html
```diff
@@ -400,38 +441,53 @@
...
+					for (var i = 0; i < enrollmentEntities.length; i++) {
+						// Fetch each search result's organization's information
+						(function(enrollment, self, parser) {
+							var ajax = document.createElement('d2l-ajax');
+							ajax.url = enrollment.getLinkByRel(/\/organization$/).href;
+							ajax.method = 'GET';
+							ajax.onResponse = function(response) {
+								if (response.detail.status === 200 || response.detail.status === 304) {
+									var organization = parser.parse(response.detail.response);
+									self.push('_searchResults', {
```
apalaniuk: Do we care about enforcing the order of results? If so I think it can be done in another PR.

apalaniuk:
It looks pretty good to me - A couple of the minor things exist but I don't see an easy solution to them, and they may be addressed in other ways (eg. the dropdown closing when not clicking on a selectable element).

d2l-search-widget.html
ryantmer: Hah, we totally do. Yeah these should be updating things in order of the enrollments, I'll fix that.

ryantmer:
@apalaniuk Yeah that one confuses me... definitely will have to be addressed, but yeah - I'll fix the ordering and do any design review from @thehappypixel, then I think remaining stuff will be follow-ups as this one is getting aged.

apalaniuk:
LGTM!

d2l-all-courses.html
```diff
@@ -116,6 +398,113 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+			_updateSortBy: function(e) {
+				this._sortField = e.target.value;
+
+				var buttons = Polymer.dom(this.root).querySelectorAll('#sortDropdown .dropdown-content-item');
+				buttons.forEach(function(button) {
+					this.toggleClass('dropdown-content-item-selected', button.value === e.target.value, button);
+				}.bind(this));
+
+				var sortText;
+				switch (this._sortField) {
```
jstefaniuk-d2l: Rather than a switch-case, maybe this can be a dictionary lookup?

ryantmer: Ah yeah, thanks. Done! :)

d2l-all-courses.html
```diff
@@ -116,6 +414,117 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+				}
+			},
+			_onSemestersResponse: function(response) {
+				if (response.detail.status === 200) {
+					var parser = document.createElement('d2l-siren-parser');
+					var semesters = parser.parse(response.detail.xhr.response);
+					this._semesterOrganizations = semesters.entities;
+				}
+			},
+			_selectSemesterList: function() {
```
jstefaniuk-d2l: You could quarter the code in these functions by having a

```
_toggleListSelected(list, listButton, listHighlight, value) {
    this.toggleClass('hidden', !value, list);
    this.toggleClass('dropdown-content-tab-button-selected', value, listButton);
    this.toggleClass('invisible', !value, listHighlight);
    listButton.setAttribute('aria-pressed', value);
}
```

utility function

d2l-all-courses.html
```diff
@@ -25,15 +31,276 @@
...
+				width: 100%;
+				text-align: left;
+			}
+			.dropdown-content-item input[type=checkbox]:after {
+				content: '';
+				display: inline-block;
+				visibility: visible;
+				width: 1rem;
+				height: 1rem;
+				margin-top: -0.1rem;
```
jstefaniuk-d2l: Can you use margin shorthand here?

ryantmer: Could, although I tend to prefer the explicit one since I never remember the order of the shorthand :( Also seems to be more commonly used in the d2l-*-ui stuff (not everywhere, but I'm not the first in other words :P)

d2l-all-courses.html
ryantmer: Tis funny, every time I've added a line to this I'm like, "this is a pain in the ass and should make it better". So yeah, will do :)

d2l-all-courses.html
```diff
@@ -451,24 +451,18 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-				this.toggleClass('dropdown-content-tab-button-selected', false, this.$.semesterListButton);
-				this.toggleClass('dropdown-content-tab-button-selected', true, this.$.departmentListButton);
-				this.toggleClass('invisible', true, this.$.semesterListHighlight);
-				this.toggleClass('invisible', false, this.$.departmentListHighlight);
-				this.$.semesterListButton.setAttribute('aria-pressed', false);
-				this.$.departmentListButton.setAttribute('aria-pressed', true);
+				this._toggleSelectedList(this.$.departmentList, this.$.departmentListButton, this.$.departmentListHighlight, true);
+				this._toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, false);
+			},
+			_toggleSelectedList(list, button, highlight, selected) {
```
jstefaniuk-d2l: missing `function` (my bad)

ryantmer: Oh hah, whoops. Still works either way, but yeah I'll fix that.

d2l-all-courses.html
```diff
@@ -25,15 +32,304 @@
...
+				<span id="filterText" class="dropdown-text" on-tap="_toggleFilterDropdown">{{localize('filtering.filter')}}</span>
+				<d2l-dropdown id="filterDropdown">
+					<button
+						id="filterButton"
+						on-mouseenter="_onFocusFilterText"
+						on-mouseout="_onBlurFilterText"
+						class="d2l-dropdown-opener dropdown-button">
+						<iron-icon icon="d2l-tier1:chevron-down" aria-hidden="true"></iron-icon>
+					</button>
+					<d2l-dropdown-menu no-padding min-width="350">
```
dbatiste: Note: `d2l-dropdown-menu` expects a menu to be inside.  It is meant to be used in conjunction with `d2l-menu`.  It's purpose is to add top/bottom caps so that scrolling menus inside `d2l-dropdown` don't look weird, resets nested menus to root upon reopening (after drilling in then closing), calls `focus` on `d2l-menu` to set focus to first menu item, and closes then menu when a menu item has been selected.  I think you may get some JS errors if there is no such component inside.

ryantmer: Ah, damn. I was looking for those box-shadows that indicate more content is available really... This really isn't a menu though, so shoehorning it in just to get that is probably inadvisable, yeah?

dbatiste: Yeah... well, we could possibly add those caps for the generic dropdown that scrolls too.  Assuming the designers are on the same page, it would probably be pretty easy to add.  Some of the logic would probably just get moved into the underlying behavior.

dbatiste: ... I'd be more inlined to update the generic behavior.  I think the `dropdown-menu` is going to try and provide some behavior that won't quite work and possibly give errors.

ryantmer: Yeah, agreed. I'll move back to the `d2l-dropdown-content` in that case, and look into possibly doing that (might just make a GH issue/"feature request" if that makes sense? Just so the thought doesn't get lost once this PR is merged). Thanks! :)

dbatiste: Sounds good!

ryantmer:
Hokay, I think I've fixed everything that was brought up... if anybody wants to give this another look, coolio, if not then I will merge tomorrow morning at some point :)

jstefaniuk-d2l:
Code lgtm

PR #117, opened 2016-09-12
Add bottom border to list item if it's the last one

ryantmer:
Itty bitty PR. Found this while doing #116 but figured I'd keep it separate, since it's an unrelated change.

(Also use cursor pointer, since the listbox items are links.)


ryantmer:
(If someone wants to take a look, coolio, if not I'll just merge since it's just a slight styling change)

trevorjonesd2l:
:+1:

Issue #57, opened 2016-06-21
Keyboard focus doesn't pick up new course tiles


ryantmer:
When you use the keyboard to pin/unpin a course in the all courses view, the course tile(s) that switch between lists cannot be selected by the keyboard. This appears to be a known issue, https://github.com/PolymerElements/iron-overlay-behavior/issues/184.

Going to wait for a solution on that prior to addressing, as it would likely be non-trivial to fix it locally, and all for naught if it gets fixed in `iron-overlay` anyway.

PR #119, opened 2016-09-13
Give overlay content tabability

trevorjonesd2l:
There's no story.

This is a PR to address the problem where nothing other than the close button is keyboard accessible. With this change the content of the overlay should be accessible by tabbing/keyboard.

Added content of the overlay to the focusable nodes to make all the my-courses stuff focusable.

rrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrrr

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -45,7 +45,7 @@
     "iron-iconset-svg": "PolymerElements/iron-iconset-svg#^1.0.9",
     "iron-input": "^1.0.10",
     "iron-menu-behavior": "^1.1.9",
-    "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.7.1",
+    "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.8.6",
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
--- a/bower.json
+++ b/bower.json
@@ -45,7 +45,7 @@
     "iron-iconset-svg": "PolymerElements/iron-iconset-svg#^1.0.9",
     "iron-input": "^1.0.10",
     "iron-menu-behavior": "^1.1.9",
-    "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.8.6",
+    "iron-overlay-behavior": "https://github.com/PolymerElements/iron-overlay-behavior.git#42a0843af78d5a8e384f0643b9b02fc522d2a2e5",
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -124,8 +124,7 @@ The overlay supports using different animations and transitions for desktop and 
 				this._updateBackdropStyle();
 			},
 			ready: function() {
-				var mutationObserver = new MutationObserver(this.invalidateTabbables.bind(this));
-				mutationObserver.observe(this, {
+				new MutationObserver(this.invalidateTabbables.bind(this)).observe(this, {
 					childList: true,
 					subtree: true
 				});
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -10,6 +10,7 @@
 <link rel="import" href="../d2l-icons/d2l-icons.html">
 <link rel="import" href="d2l-simple-overlay-styles.html">
 <link rel="import" href="localize-behavior.html">
+<link rel="import" href="../d2l-polymer-behaviors/d2l-dom.html">
 
 <!--
 `<d2l-simple-overlay>` extends `iron-overlay` and adds some styling using various `neon` Polymer elements.
@@ -122,8 +123,15 @@ The overlay supports using different animations and transitions for desktop and 
 			created: function() {
 				this._updateBackdropStyle();
 			},
+			ready: function() {
+				var mutationObserver = new MutationObserver(this.invalidateTabbables.bind(this));
+				mutationObserver.observe(this, {
+					childList: true,
+					subtree: true
+				});
+			},
 			get _focusableNodes() {
-				return [this.$$('.close-button'), this.$$('content')];
+				return [D2L.Dom.Focus.getFirstFocusableDescendant(this), D2L.Dom.Focus.getLastFocusableDescendant(this)];
 			},
 			_handleClose: function() {
 				this.close();
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -132,6 +132,7 @@ The overlay supports using different animations and transitions for desktop and 
 				}
 
 				focusables.push(Polymer.dom(this.root).querySelector('.close-button'));
+				focusables.push(Polymer.dom(this.root).querySelector('content'));
 
 				return focusables;
 			},
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -131,8 +131,8 @@ The overlay supports using different animations and transitions for desktop and 
 					focusables = Polymer.IronOverlayBehaviorImpl.__lookupGetter__('_focusableNodes').call(this);
 				}
 
-				focusables.push(Polymer.dom(this.root).querySelector('.close-button'));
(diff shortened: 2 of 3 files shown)


apalaniuk:
It works much better. Woo!

Might be for a different reason than expected though; `focusableNodesOverride` (which is what creates the array of focusable nodes in the overlay) returns an empty array right now because the `d2l-course-tile` query will return no elements as they were moved into another component (`d2l-course-tile-grid`) a while ago, so the overlay manager (`iron-overlay-behavior`) didn't know about the course tiles. By pushing the `content` node onto the array, it doesn't get stuck on the close button because it doesn't think it's both the first and last element anymore, but it also will never wrap around (which is also making me rethink the accessibility of infinite scrolling..)

Incidentally, this does "solve" #57 which we were waiting to be fixed, but after looking at https://github.com/PolymerElements/iron-overlay-behavior/pull/199, the fix we were waiting for might actually be in! (Note: Did not look too hard at this yet.)

trevorjonesd2l:
It does wrap the `close-button` - I'll look more at how it wraps the end of the list and see if `invalidateTabbables()` can help with that wrapping

apalaniuk:
@trevorjonesd2l Hrm, I wasn't seeing it - but I'll check that I didn't make a mistake somewhere.

If it does work correctly, then I think we can get rid of the `focusableNodesOverride` code above it as well.

dbatiste:
It would always make me a bit nervous accessing behavior's private members:
 `focusables = Polymer.IronOverlayBehaviorImpl.__lookupGetter__('_focusableNodes').call(this);`  

For what it's worth, we have a few focus management helpers in [d2l-polymer-behaviors](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-focus.html) that work on composed DOM regardless of shady/shadow enabled.  I am guessing since we're relying on the iron-overlay stuff here, it's necessary to use the focus management that comes with it?  We found it limiting with respect to composed DOM, I think especially when shadow-DOM is enabled.

Re accessibility and infinite scrolling: at the risk of getting into design most accessibility reviews suggest its usually always bad.  One of the biggest issues is making the content that follows difficult to access for keyboard users because they may never reach the end of the list.

trevorjonesd2l:
I'm finding the focus-wrapping actually works as I'd expect it to, even with the infinite-scroll.

I think the accessibility concerns around the infinite-scroll would actually be a separate issue. I remember we had the same issue when working on binder web and ended up finding a way to address it. I'll see how it worked on binder and talk to the designer and see what can be done for that infinite-scroll accessibility in my courses. If things move smoothly I'll put up another PR this week to address those accessibility concerns.

apalaniuk:
I agree about the hacky getter access that was added a while ago - it's never called in practice and can be removed (and looks like it has been now). The infinite scroll issue you listed is one of the ones that concerned me as well, but as Trevor said, I should make a separate issue for that - thanks for the input!

I'm still not quite sure how wraparound would work, and in my testing it doesn't - eg. I can't shift+tab on the close button to focus on the last focusable element in the dialog; focus will stay on the close button (Chrome, Firefox). In Firefox, it will wrap around to the beginning of the dialog from the end, but only after first tabbing through elements on the main page behind the overlay because it doesn't know it's reached the last element (ie, focus is not confined to the overlay).

So, I think there are fixes for this (ensure that the correct first/last focusable elements are set, and maybe use `invalidateTabbables` where necessary), but we should probably replace this at some point with our own behavior because theirs seems very prone to issues.

dbatiste:
Yeah, I've only really read through the `iron-overlay` code, so I don't have too much knowledge of what works and what doesn't (besides that it's method of identifying focusable elements was not going to work for us).

That being said, the approach used in Nav ([here](https://github.com/Brightspace/d2l-navigation-ui/blob/master/slide-out-menu.html#L73)) to trap the focus in the mobile slide-out utilizes a focusable element before and after, and when focus lands on those elements the focus is programmatically set either to the last or first expected focusable elements respectively.  It effectively traps the focus.  @dlockhart : is this still the approach you'd recommend?  

If the above approach was used in a custom overlay implementation, initial focus would need to be placed on the first expected focusable element.  In such case I would implement `focus` on the overlay component to handle that, and I would probably use [getFirstFocusableDescendant](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-focus.html#L23) and [getLastFocusableDescendant](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-focus.html#L44) to identify the elements to focus on.

dlockhart:
The focus trapping approach is recommended for things that are intended to be modal, like a modal dialog or in the case you mentioned with the slide-out menu in navigation. The underlying page is shimmed out and not meant to be "visible" to AT.

In terms of making infinite scrolling accessible, I'd recommend placing an off-screen (visible to AT only) button below the last item that when clicked, loads in the next "page" of items, just as scrolling would have. That gives AT users a way to load the items while interacting with the virtual DOM, since they won't actually be scrolling the page.

bower.json
```diff
@@ -45,7 +45,7 @@
     "iron-iconset-svg": "PolymerElements/iron-iconset-svg#^1.0.9",
     "iron-input": "^1.0.10",
     "iron-menu-behavior": "^1.1.9",
-    "iron-overlay-behavior": "PolymerElements/iron-overlay-behavior#^1.7.1",
+    "iron-overlay-behavior": "https://github.com/PolymerElements/iron-overlay-behavior.git#42a0843af78d5a8e384f0643b9b02fc522d2a2e5",
```
trevorjonesd2l: :unamused: change we need hasn't been released yet

d2l-simple-overlay.html
```diff
@@ -122,18 +123,15 @@ <h1 class="d2l-heading-2">{{title}}</h1>
 			created: function() {
 				this._updateBackdropStyle();
 			},
+			ready: function() {
+				var mutationObserver = new MutationObserver(this.invalidateTabbables.bind(this));
```
apalaniuk: I think we can drop the temp variable since it isn't saved and the instance will keep a strong reference to the observer automatically, then can just call observe on the returned reference.

d2l-simple-overlay.html
```diff
@@ -122,18 +123,15 @@ <h1 class="d2l-heading-2">{{title}}</h1>
...
+				mutationObserver.observe(this, {
+					childList: true,
+					subtree: true
```
apalaniuk: One thing I realized is that we'd probably want to watch attributes that can affect whether an element is focusable, like tabindex (and hidden, though I think that one might require a bit of extra logic?), though a quick test of using `attribute` and `attributeFilter` doesn't seem to work (just `attribute` works, but seems aggressive).

This can probably wait until I rebase this into my breaking out of d2l-simple-overlay into a more generic dialog since I don't think this can cause a problem in the current case.

apalaniuk:
LGTM

d2l-simple-overlay.html
dbatiste: [This is the function](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blob/master/d2l-dom-focus.html#L198) we've been using to test for focusable.  It does some DOM walking to check visibility of ancestors, so that's something to be considered.  Anyway, worth knowing this is there if you need it.

PR #120, opened 2016-09-16
US76123 - Dropdown search

ryantmer:
Eyyy, first follow-up PR to #116. For this, I broke out and modified the search bar that already existed, and used it in the dropdowns. Unfortunately, the existing search has a lot of complexity and "custom-ness" about it (mainly the live, drop-down search results stuff), so replacing that with the broken out component would have greatly increased the complexity of the broken out component. What I'm thinking is that, if we change this to update the entire tile view with the search results, then we may possibly be able to forego the dropdown, and instead just display search results (live or not) as course tiles. (At the very least, we do want to update the course tile view to be the full search results when you hit Enter - next thing I'm working on.)

Currently this is locked to a specific version of `d2l-search-widget` on purpose, as it hasn't hit 1.0.0 yet, and won't until I evaluate the above w.r.t. replacing the other search bar with it as well (meaning it may have to at least support live search).

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -336,10 +336,6 @@
 								<iron-icon icon="d2l-tier1:check" aria-hidden="true"></iron-icon>
 								{{localize('sorting.datePinned')}}
 							</button>
-							<button class="dropdown-content-item" value="department" on-tap="_updateSortBy">
-								<iron-icon icon="d2l-tier1:check" aria-hidden="true"></iron-icon>
-								{{localize('sorting.department')}}
-							</button>
 							<button class="dropdown-content-item" value="lastAccessed" on-tap="_updateSortBy">
 								<iron-icon icon="d2l-tier1:check" aria-hidden="true"></iron-icon>
 								{{localize('sorting.lastAccessed')}}
@@ -468,7 +464,7 @@
 			},
 			attached: function() {
 				this._onResize();
-				window.addEventListener('resize', this._onResize);
+				window.addEventListener('resize', this._onResize.bind(this));
 
 				this.listen(this.$.semesterSearchWidget, 'search-results-changed', '_onSemesterSearchResults');
 				this.listen(this.$.departmentSearchWidget, 'search-results-changed', '_onDepartmentSearchResults');
@@ -478,7 +474,7 @@
 				this.$.scrollThreshold.scrollTarget = this.$.filterDropdownContent.querySelector('.d2l-dropdown-content-container');
 			},
 			detached: function() {
-				window.removeEventListener('resize', this._onResize);
+				window.removeEventListener('resize', this._onResize.bind(this));
 
 				this.unlisten(this.$.semesterSearchWidget, 'search-results-changed', '_onSemesterSearchResults');
 				this.unlisten(this.$.departmentSearchWidget, 'search-results-changed', '_onDepartmentSearchResults');
@@ -519,8 +515,11 @@
 				'courseCode': 'sorting.sortCourseCode',
 				'courseName': 'sorting.sortCourseName',
 				'pinDate': 'sorting.sortDatePinned',
-				'department': 'sorting.sortDepartment',
 				'lastAccessed': 'sorting.sortLastAccessed'
+			},
+			_changeTextFocus: function(element, focused) {
+				this.toggleClass('highlight', focused, element);
+				this.toggleClass('selected', focused, element);
 			},
 			_checkFilterEntities: function(entities) {
 				entities.forEach(function(entity) {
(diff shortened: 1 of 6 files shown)


d2l-all-courses.html
```diff
@@ -375,19 +374,39 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-				},
 				_isFiltered: {
 					type: Boolean,
 					value: false
 				},
-				_semesterOrganizations: Object,
-				_semestersUrl: {
-					type: String,
-					value: '/d2l/api/hm/organizations?embedDepth=1&organizationTypeIds=5'
+				_searchSemestersAction: {
```
ryantmer: This and `_searchDepartmentsAction` below are set to current known values, although it is possible for these to change. The correct way of doing this to use the `add-semester-filter` and `add-department-filter` actions that exist on the /organizations HM API, however it will make more sense to use those in the PR that will also have the display-search-results-as-course-tiles PR.

d2l-all-courses.html
```diff
@@ -480,19 +506,35 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-					var parser = document.createElement('d2l-siren-parser');
-					var departments = parser.parse(response.detail.xhr.response);
-					this._departmentOrganizations = departments.entities;
-				}
+			_onDepartmentSearchResults: function(e) {
+				this.set('_departmentOrganizations', e.detail.value);
+
+				// There is a slight delay between this.set and the DOM being updated; we
+				// need this to run after the latter, so give it a bit of a delay
+				setTimeout(function() {
```
ryantmer: @apalaniuk Did you ever see something like this, and do you know of a solution for it? The issue here is that if I do the `this.set` and then immediately `querySelectorAll`, I get outdated info. Is there a way to make sure we're doing this after `this.set` actually updated the DOM off the top of your head? (This works with `setTimeout`, it just puts my jimmies at fairly high rustle-level and seems janky.)

d2l-search-widget-custom.html
```diff
@@ -10,7 +10,7 @@
 <link rel="import" href="d2l-utility-behavior.html">
 <link rel="import" href="d2l-search-dropdown-content.html">
 
-<dom-module id="d2l-search-widget">
+<dom-module id="d2l-search-widget-custom">
```
ryantmer: Not a great name, but I'm thinking/hoping this can be removed and replacing with `d2l-search-widget` in my next PR.

localize-behavior.html
```diff
@@ -68,7 +68,7 @@
 							'filtering.semester': 'Semester',
 							'filtering.department': 'Department',
 							'filtering.semesterMultiple': 'Semester ({num})',
-							'filtering.departmentMultiple': 'Semester ({num})',
+							'filtering.departmentMultiple': 'Department ({num})',
```
ryantmer: This was just a straight-up mistake on my part from a previous PR, whoops.

d2l-all-courses.html
apalaniuk: You could listen for the `dom-change` event on the template. I imagine it's not possible to use data binding alone?

ryantmer: The issue that arose with the data binding was that the checkboxes wouldn't update (i.e. search, select 2nd checkbox in list, clear search, 2nd checkbox is still selected even though it's now a different option). Could be that I'm missing an obvious way to do this though, so I may look into it further.

jstefaniuk-d2l: What if you always clear them all, and just set the correct ones to `true` afterwards?

ryantmer: The problem here is (I think) ensuring that it is indeed "afterwards" - since in this case I need things to happen in a guaranteed order :< Which, now that I've said that out loud, makes me realize code like that is generally bad code in something that is designed to be async 😜

jstefaniuk-d2l: Or could you have a boolean `checked` value bound to the DOM (here: https://github.com/Brightspace/d2l-my-courses-ui/pull/120/files#diff-98931a3c4fd43464bdaaf7d5c18ce360L282), and have it update without a `querySelector` call?

(I'm not quite sure how polymer does things)

ryantmer: Oh shit, that definitely seems plausible, forgot you could pass a function to `checked`. I'll look into that, thanks!

jstefaniuk-d2l: Or you could also preprocess the `_departmentOrganizations` object to have the correct `checked` values (putting the info in a new object)

ryantmer: Oh, as a follow-up to this - tried a few things, which all didn't work, or muddled things significantly. While `setTimeout` isn't my favourite thing in the world, it is at least clear, so I think this should be okay.

jstefaniuk-d2l:
```
			_onDepartmentSearchResults: function(e) {
				e.detail.value.forEach(function(item) {
					if(this._parentOrganizationIds.indexOf(item.properties.id.toString()) > -1) {
						item.properties.checked = true;
					}
				}.bind(this));

				this.set('_departmentOrganizations', e.detail.value);
			},
```

Is this a bit better?

ryantmer: Hmm, that does seem alright. My only concern is that that makes the stored `_departmentOrganizations` a modified set of Siren entities, but that should be okay I suppose, since it's really not going anywhere.

ryantmer: Ah, damn. Doesn't actually work - I think the issue is again with the update timing that was showing up without the `setTimeout`. Ah well, not the end of the world I'd say.

jstefaniuk-d2l: Tried it out and it seems to work on my machine, is there any special things I need to do to trigger the bug?

ryantmer: Not as far as I know? I'm just going into the semester list, searching for something, selecting the nth item, then clearing the search - the checkbox remains on the nth item, rather than the correct item. (e.g. searching for "implatform", hit Enter, select the (one) option, then click the X to clear the search - the first item is selected, rather than "IMPLATFORM Test Semester")

jstefaniuk-d2l: Hmm, well maybe pull down this branch for a sec and see if it works for you, if not: LGTM

https://github.com/Brightspace/d2l-my-courses-ui/tree/testingDiff

ryantmer: ...I have no idea if you're magic, or if (quite likely) I managed to try implementing that the wrong way multiple times. I have no idea. But hey, works now (added in most recent commit), thanks Jeff :D

jstefaniuk-d2l:
:+1:

trevorjonesd2l:
So initial set of things to fix from testings:
- [ ] Clear search when switching tabs
- [ ] Some sort of paging in the dropdown so I can access all my Semesters/Departments
- [ ] Remove this from mobile until the mobile work is done (mobile not good right now with this)

d2l-all-courses.html
```diff
@@ -497,6 +532,18 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				this._numSemesterFilters = 0;
 				this._numDepartmentFilters = 0;
 			},
+			_loadMoreDepartments: function() {
+				console.log('load more departments');
+				if (this._hasMoreDepartments) {
+					console.log('but like legit tho');
```
jstefaniuk-d2l: :+1:

ryantmer: Haha, yeah - lazy load still isn't working, so meh. (I will remove these, o'course)

ryantmer:
@trevorjonesd2l Done! :)

d2l-all-courses.html
ryantmer: Lazy load is now working, and these have been removed - apparently with GH's new code review features, comments on out-of-date stuff don't get hidden, which is lame :(

d2l-all-courses.html
```diff
@@ -514,20 +569,61 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+			},
+			_onMoreSemestersResponse: function(response) {
+				if (response.detail.status === 200 || response.detail.status === 304) {
+					var parser = document.createElement('d2l-siren-parser');
+					var moreSemesters = parser.parse(response.detail.xhr.response);
+					moreSemesters.entities.forEach(function(semester) {
+						this.push('_semesterOrganizations', semester);
+					}.bind(this));
+
+					this.set('_moreSemestersUrl', moreSemesters.hasLinkByRel('next') ? moreSemesters.getLinkByRel('next').href : '');
```
jstefaniuk-d2l: https://github.com/Brightspace/d2l-my-courses-ui/pull/127#discussion-diff-79922020R589

ryantmer: Ah, yep, forgot about that. Will do.

d2l-all-courses.html
```diff
@@ -480,27 +568,62 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				this.toggleClass('highlight', true, this.$.sortText);
 				this.toggleClass('selected', true, this.$.sortText);
 			},
-			_onDepartmentsResponse: function(response) {
-				if (response.detail.status === 200) {
+			_onDepartmentSearchResults: function(e) {
```
jstefaniuk-d2l: As Dave mentioned on the other PR, some utility functions could go a long way here to reduce repeated code

d2l-all-courses.html
```diff
@@ -480,27 +568,62 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+					moreSemesters.entities.forEach(function(semester) {
+						this.push('_semesterOrganizations', semester);
+					}.bind(this));
+
+					this.set('_moreSemestersUrl', moreSemesters.hasLinkByRel('next') ? moreSemesters.getLinkByRel('next').href : '');
+					this.set('_hasMoreSemesters', this._moreSemestersUrl.length > 0);
 				}
 			},
+			_onResize: function() {
+				this.toggleClass('hidden', window.innerWidth < 768, this.$.filterAndSort);
```
jstefaniuk-d2l: I recommend stealing and using this util file: https://github.com/Brightspace/sequence-viewer/blob/master/src/utils/screenSize.js

jstefaniuk-d2l:
Are tests needed for any of this stuff?

d2l-all-courses.html
ryantmer: Yeah, taking a look at that now. The filter/sort duplication and semester/department duplication level is like 9001.

ryantmer:
Hm... maybe? Most of this would be a super pain to write tests for, but I'll see if some of the more behavioural things fit the bill (whereas most is UI-y stuff).

jstefaniuk-d2l:
(Not for this PR) This Dropdown stuff seems like it should eventually be moved to it's own component

ryantmer:
@jstefaniuk-d2l Which dropdown stuff? (I do plan on reducing the custom-ness of the dropdowns, but that'll be part of #127 as it's mostly related to A11y-ness)

jstefaniuk-d2l:
Possibly the whole search area? All-courses seems to be getting kinda bloated

I'm not saying this _needs_ to be done, just thinking aloud (err... with keystrokes)

ryantmer:
Ah, yeah no I get you. Once I use the existing `d2l-menu` stuff it should cut it down pretty significantly, so I'll re-evaluate once that's in.

trevorjonesd2l:
- [ ] Fix iE

ryantmer:
Phew - it was just a small bug in `d2l-search-widget` that was breaking IE. Phew.

(Explanation: the observer that observes the search URL being used by the widget was being called when it was initially created (as Polymer do), which other browsers didn't seem to mind, but IE did... For the first time in my life, it seems like IE was doing the expected thing?)

trevorjonesd2l:
There are problems with the dropdown in general on Safari and iPad (a11y story should fix Safari, iPad another story). Search works and scrolls and stuff LGTM

ryantmer:
🙌 🙌 🙌

Issue #125, opened 2016-09-19
Filtering in All Courses overlay not fully accessible


trevorjonesd2l:
Not fully tested, but here's what I notice right away:

- [ ] No focus state (or hover) on the clear search button
- [ ] No focus state on the Filter or Date Pinned buttons (slight focus state, should mimic the hover state)
- [ ] Filter menu:
 1. Can't switch tabs (Semester and Department)
 2. No hover/focus state of Semester, Department or Clear buttons
 3. Focus doesn't wrap
 4. Items should be navigated by arrows (probably)
- [ ] Date Pinned menu not accessible with keyboard (can't navigate the menu)

You should also check the screen-readers when testing this, didn't do any of the label testing but it will be worth looking at.

ryantmer:
🙌

ryantmer:
Hm, fun - it seems that the lack of up/down key interaction may be caused by `iron-overlay-behavior`, which seems to bubble key events up to the parent. Might require more in-depth changes than just `role="menu"`, which explains much.

PR #129, opened 2016-09-23
A11Y fixes for All Courses overlay

### LINKED ISSUE #121: Focus on All Courses overlay is sometimes lost
1. If you pin the last course on the list, you lose focus (it goes somewhere unexpected) - when this happens focus should move somewhere more expected, like the new last course tile

2. When you pin or unpin a course the focus isn't explicitly anywhere - this should also move the focus to the next course tile

### LINKED ISSUE #131: Can't use filter/sort dropdowns on Safari (Mac)

ryantmer:
This addresses the A11Y concerns outlined in #125. Should (hopefully) also fix #131. Also fix #121.


d2l-all-courses.html
```diff
@@ -661,16 +718,14 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
 			_updateSortBy: function(e) {
-				this._sortField = e.target.value;
+				this.set('_sortField', e.detail);
 
-				var buttons = Polymer.dom(this.root).querySelectorAll('#sortDropdown .dropdown-content-item');
-				buttons.forEach(function(button) {
-					this.toggleClass('selected', button.value === e.target.value, button);
+				var menuItems = Polymer.dom(this.root).querySelectorAll('d2l-menu-item-selectable');
+				menuItems.forEach(function(item) {
+					item.value === this._sortField ? item.setAttribute('selected', true) : item.removeAttribute('selected');
```
jstefaniuk-d2l: (Copied from other PR)

`===` seems like the wrong thing here

jstefaniuk-d2l: (Talked to Ryan, I read this statement wrong XD)

trevorjonesd2l:
Leaving a note here that this should (hopefully) fix [this Safari issue](https://github.com/Brightspace/d2l-my-courses-ui/issues/131)

ryantmer:
Updated to fix all but one of the issues outlined in #125 (which was a result of our discussion with Carin). The only outstanding one is the link list one - I can't actually get the link list to come up in NVDA, but I will work on it. Should be a fairly simple fix, as I think the problem is caused by some missing `aria-hidden`s.

ryantmer:
Note that this does change the UI of the filter menu, as I used the same custom `d2l-menu` element for both the filter and sort menus. I'll fix that tomorrow morning, just a bit of a styling change.

ryantmer:
Okay, all updated. This fixes all but one problem in #125, which was a result of our discussion with Carin. The only outstanding one is the link list one, which lists each course tile as "Math 101Math 101 course settings" (yes with the lack of space). I've tried a few things, but nothing seems to work (if I remove the `aria-label` from the dropdown then the second half goes away, but then the dropdown doesn't have a label when it's tabbed to, which is more important).

Carin had mentioned that using the link/element list is a fairly uncommon thing to anyway, so unless anybody else has any brainwaves then I'm thinking we might have to let it be? An accessibility expert I am not :sob:

d2l-all-courses.html
```diff
@@ -3,14 +3,175 @@
...
+<link rel="import" href="../d2l-menu/d2l-menu-item-behavior.html">
+<link rel="import" href="../d2l-menu/d2l-menu-item.html">
 <link rel="import" href="../d2l-search-widget/d2l-search-widget.html">
 <link rel="import" href="d2l-course-tile-grid.html">
 <link rel="import" href="d2l-course-tile-responsive-grid-behavior.html">
 <link rel="import" href="localize-behavior.html">
 <link rel="import" href="d2l-course-management-behavior.html">
 <link rel="import" href="d2l-search-widget-custom.html">
 
+<dom-module id="d2l-menu-item-sort">
```
ryantmer: This and `d2l-menu-item-filter` are just added inline right now - I'm working on getting at least one of them added to `d2l-menu` itself, so I'll update/remove them from here once that is a thing.

dbatiste: Yeah - normally I would say that anything custom should not be named `d2l-menu-*`, but if some of this is going to land their then cool.  :)

d2l-all-courses.html
```diff
@@ -3,14 +3,175 @@
...
+					observer: '_onSelectedChanged',
+					reflectToAttribute: true
+				},
+				value: String
+			},
+			behaviors: [
+				window.D2L.PolymerBehaviors.MenuItemBehavior
+			],
+			hostAttributes: {
+				'role': 'menuitem'
```
dbatiste: I think [MenuItemBehavior](https://github.com/Brightspace/d2l-menu-ui/blob/master/d2l-menu-item-behavior.html#L10) automatically provides `role`.

d2l-all-courses.html
ryantmer: I'm actually factoring out the common behaviour here now anyway, and am thinking I might have a PR up for `d2l-menu` with it by the end of the day actually, sooner/easier than expected. So I'll update this once I've got that :)

d2l-all-courses.html
```diff
@@ -3,14 +3,175 @@
...
+			},
+			listeners: {
+				'tap': '_onTap',
+				'keydown': '_onKeydown'
+			},
+			_onSelectedChanged: function(selected) {
+				this.toggleClass('hidden', !selected, this.$$('iron-icon'));
+				selected ? this.setAttribute('aria-selected', true) : this.removeAttribute('aria-selected');
```
dbatiste: Is `aria-selected` valid on `role="menuitem"`?  Or, is this component supposed to have a different role.

d2l-all-courses.html
```diff
@@ -3,14 +3,175 @@
...
+			:host(:hover),
+			:host(:focus) {
+				color: var(--d2l-color-tungsten);
+			}
+			.hidden {
+				visibility: hidden;
+			}
+		</style>
+
+		<iron-icon icon="d2l-tier1:check" aria-hidden="true"></iron-icon>
```
dbatiste: Can we use [d2l-icon](https://github.com/Brightspace/d2l-icons-ui/blob/master/d2l-icon.html) instead?  It handles rtl icons for us, as well as makes the svg not focusable (an issue that comes up in IE).

d2l-all-courses.html
```diff
@@ -3,14 +3,175 @@
...
+		<span>[[text]]</span>
+		<content></content>
```
dbatiste: Does this component expect to distribute light-DOM nodes?  If not, I'd be an advocate of removing this `<content>` element.

d2l-all-courses.html
ryantmer: I did try `role="menuitemradio"`, but that seemed to break everything for whatever reason... I'll fiddle with a bit, though, see if I can figure that out since this really is a dropdown-radiolist.

dbatiste: Yeah, in that case I'd be cool with creating a `d2l-menu-item-radio`, similar to what I did for the checkbox one.  `d2l-menu` would likely need a couple of small updates.  I can help with it if need-be.  I think we should be able to get that working.

d2l-all-courses.html
ryantmer: I modeled this after [the existing `d2l-menu-item`](https://github.com/Brightspace/d2l-menu-ui/blob/master/d2l-menu-item.html#L40), but looking at the new [`d2l-menu-item-link`](https://github.com/Brightspace/d2l-menu-ui/blob/master/d2l-menu-item-link.html#L29) and others, they don't have it, nor do I intend this to have content, so away it goes! 🎁

d2l-all-courses.html
ryantmer: Cool, that'll simplify things quite a bit here in that case; I'll get going on that! (and will probably ask many questions of you along the way)

trevorjonesd2l:
👍

trevorjonesd2l:
Stuff to fix:
- [ ] Dropdown menu for filter/departments doesn't resize once the list is loaded
- [ ] No label on the filter and sort menus (chevrons?)
- [ ] No labels on dropdown filter sort bar

ryantmer:
I have pushed ze things! The dropdown not resizing I'm considering a separate issue, as it may be solved for free when I replace the custom elements in here with the ones in https://github.com/Brightspace/d2l-menu-ui/pull/12 (and if not, I'll fix it when I switch to those anyway). The no lazy load in Firefox I'll address independently as well, since it's distinct from these changes, and may require further digging (what I thought would be the obvious solution wasn't).

For the pin/unpin alert, are we okay to let this one go for now? The reason I ask is that I've put a concerted effort into getting it to work, with no luck. From what I can see, screen readers have mixed results with `aria-live` and `aria-announce`, so I figure if it's tracked in #125 then we can come back to it if somebody gets a brainwave, otherwise I suspect it's an NVDA + FF issue 😢

trevorjonesd2l:
This looks good probably once you can load the semesters when the dropdown is opened with the keyboard

ryantmer:
Fixed dat, missed a thing.

trevorjonesd2l:
I see this here but not on master:

In the filter menu
1. Search for something (Fall)
2. Check it off (Fall Test Semester)
3. Clear the search

Fall Test Semester won't be checked off anymore. The tab will indicate there's one active filter. If I check off Fall Test Semester it will indicate that there are two active filters

ryantmer:
Hm, odd. Was caused by a `.toString`, but I didn't add it here... not sure why that is, but fixed now!

trevorjonesd2l:
👍

PR #140, opened 2016-10-11
Manually update the search results. Fixes #137.

### LINKED ISSUE #137: Empty dropdown search results showing up
Unclear why this is occurring, but sometimes empty results are shown in the dropdown on the search box in the all courses overlay.

ryantmer:
For some inexplicable reason (closest possibly-related bug is [this one](https://github.com/Polymer/polymer/issues/3682)), using the proper Polymer.splice array method would result in empty results, as it appeared the results were not being notified correctly. To fix, or at least get around, this problem, we can instead use a normal JS splice, then call notifySplices to notify Polymer of the changes, once we know all the changes have finished being applied (i.e. all /organizations requests are finished).

### DIFF
--- a/d2l-search-widget-custom.html
+++ b/d2l-search-widget-custom.html
@@ -296,6 +296,8 @@
 				document.body.removeEventListener('click', this._handleClickBound, true);
 			},
 
+			_pageSize: 5,
+
 			/*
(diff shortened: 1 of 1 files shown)


apalaniuk:
LGTM

PR #143, opened 2016-10-11
Readjust the margin

trevorjonesd2l:
Fix the margin so the grid grids again. Should fix this issue:

image

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -53,7 +53,7 @@ in that organization - student, teacher, TA, etc.
 				</div>
 				<div class="course-text d2l-heading-4">{{_organization.properties.name}}</div>
 				<!-- Uncomment this line to get a temporary choose image button -->
-				<!-- <button on-click="_launchCourseTileImageSelector">Choose Image</button> -->
+				<button on-click="_launchCourseTileImageSelector">Choose Image</button>
 			</a>
 
 			<div class="hover-menu no-tap-interaction">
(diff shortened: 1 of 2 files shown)


d2l-course-tile.html
```diff
@@ -53,7 +53,7 @@
 				</div>
 				<div class="course-text d2l-heading-4">{{_organization.properties.name}}</div>
 				<!-- Uncomment this line to get a temporary choose image button -->
-				<!-- <button on-click="_launchCourseTileImageSelector">Choose Image</button> -->
+				<button on-click="_launchCourseTileImageSelector">Choose Image</button>
```
trevorjonesd2l: - [ ] TODO: Remove after tests and stuff

jstefaniuk-d2l:
Code LGTM (I'd run this by Bob as well)

ryantmer:
Is this more durable than the previous solution?

jstefaniuk-d2l:
@ryantmer Not actually sure why this broke, it probably has something to do with a change here: https://github.com/Brightspace/d2l-my-courses-ui/pull/129 (since that's the only other PR)

Since this grid is pretty much the same as the course tiles (just simplified and resized), I'm not sure we need to worry that much about robustness (Once we hook it up to the actual ui, anything breaking it will be caught with testing)

ryantmer:
Hm, maybe - thought it was in master before merging that, granted 😮

ryantmer:
Poked around with this, seems to work a-okay.

PR #144, opened 2016-10-12
Update to latest d2l-dropdown and d2l-menu.  

dbatiste:
Updating "select" event accordingly.


dbatiste:
fyi: @jstefaniuk-d2l , @apalaniuk , @ryantmer : 

The `select` event name changed for the `d2l-menu` component.  This was the only spot I found that needed updating.  Let me know if you think there should be others.

ryantmer:
Yep, LGTM - we're not using a lot of `d2l-menu` directly :)

ryantmer:
@dbatiste mind if I merge this? We're lagging behind in `d2l-menu` versions in particular, causing BSI version bummers.

dbatiste:
go for it.  `d2l-navigation` has a similar PR that will need to be merged before updating BSI since this is breaking change.

PR #145, opened 2016-10-12
Fetch actual department/semester type IDs

ryantmer:
What the actual organization type ID of a semester or department is can be set in the LMS (not sure if it ever really changed, but can be). Safest route is to hit the /semester and /department routes in Valence, which give us back the ID we know to be correct, and use that as the organizationTypeId for the department/semester list/search requests.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -319,15 +319,9 @@
 		</style>
 
 		<d2l-ajax
-			id="departmentTypeRequest"
-			url="[[_orgUnitTypeDepartmentUrl]]"
-			on-iron-ajax-response="_onOrgUnitTypeDepartmentResponse">
-		</d2l-ajax>
-
-		<d2l-ajax
-			id="semesterTypeRequest"
-			url="[[_orgUnitTypeSemesterUrl]]"
-			on-iron-ajax-response="_onOrgUnitTypeSemesterResponse">
+			id="organizationsRequest"
+			url="[[_organizationsUrl]]"
+			on-iron-ajax-response="_onOrganizationsResponse">
 		</d2l-ajax>
 
 		<d2l-ajax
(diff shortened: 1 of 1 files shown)


ryantmer:
This isn't associated with a US or anything - just something that occurred to me.

Doesn't really overly need to be tested in-depth, mostly just sanity-checked, since I'm not sure how many clients actually change what the default department/semester types are...

apalaniuk:
Could this use the values of `organizationTypeIds` from the `add-department-filter` and `add-semester-filter` actions on the `organization` root HM response, or do I not grok this? (probably the latter)

ryantmer:
Well... damn. Nope, you are totally correct, and that's way cleaner. Totally forgot that was there, thanks!

ryantmer:
Yep, much simpler using that. Updated to fetch _/organizations_ instead - could probably leverage another request, but figured that would muddy some lines that maybe shouldn't be muddied.

d2l-all-courses.html
```diff
@@ -534,6 +547,8 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 							{'name':'organizationTypeIds', 'type':'hidden', 'value':'5'}]
 					}
 				},
+				// 203 is the default type ID for semesters
```
apalaniuk: nit: `departments`

apalaniuk:
Looks legit (and a good catch, as eg. daylightdemo *does* change the department type ID!) :shipit:

d2l-all-courses.html
ryantmer: Pfft, I don't copy+paste things, idk what you're talking about, never...

d2l-all-courses.html
```diff
@@ -717,6 +730,43 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+				if (response.detail.status === 200) {
+					var parser = document.createElement('d2l-siren-parser');
+					var responseEntity = parser.parse(response.detail.xhr.response);
+
+					var departmentId = responseEntity.getActionByName('add-department-filter').getFieldByName('organizationTypeIds').value;
+					var semesterId = responseEntity.getActionByName('add-semester-filter').getFieldByName('organizationTypeIds').value;
+					this.set('_searchDepartmentsAction', {
+						name: 'search-course-collection',
+						method: 'GET',
+						href:'/d2l/api/hm/organizations',
```
ryantmer: Already did in fact :P

ryantmer: Already did, in fact! :P

PR #150, opened 2016-10-14
Don't use orgunitid

ryantmer:
This might require some explanation -

The _/organizations_ HM API right now has a property, `id`, that is the org unit ID. This shouldn't be there, since IDs in Hypermedia _should_ really be the self link for the entity, since that is unique. I added the ID since it was available from the `IOrgUnit`, and wasn't thinking, basically.

I've added the `parentOrganization` query parameter to the HM API ([dis](https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/3467/overview)), which is equivalent to `parentOrganizationIds` except that it takes comma-separated URIs rather than comma-separated ints.

This is the only usage of this query parameter (Pulse almost used it, but Eduardo asked about it, and @jopnick noticed that it shouldn't be there and let me know), so once this is merged then there will be another `lp` PR to remove `parentOrganizationIds` and the `id` property entirely.

This isn't... strictly a _bug_, but I'm tagging it as such because it is an incorrect behaviour.

### DIFF
--- a/d2l-utility-behavior.html
+++ b/d2l-utility-behavior.html
@@ -55,14 +55,23 @@
 			return sizes.highMax || sizes.lowMax || sizes.highMid || sizes.lowMid || sizes.highMin || sizes.lowMin;
 		},
 		/*
-		* Creates a unique identifier for an enrollment (really just the self Link href)
-		* @param {Entity} enrollment
+		* Creates a unique identifier for a Siren Entity (really just the self Link href)
+		* @param {Entity}
 		* @return {String}
 		*/
-		getEnrollmentIdentifier: function(enrollment) {
-			// An enrollment's self href should be unique, so use it as an identifier
-			var selfLink = enrollment.getLinkByRel('self');
-			return selfLink.href;
+		getEntityIdentifier: function(entity) {
+			// An entity's self href should be unique, so use it as an identifier
+			if (entity.getLinkByRel) {
+				var selfLink = entity.getLinkByRel('self');
+				return selfLink.href;
+			} else {
+				var links = entity.links || [];
+				links.forEach(function(link) {
+					if (link.rel.indexOf('self') > -1) {
+						return link.href;
+					}
+				});
+			}
 		}
 	};
 </script>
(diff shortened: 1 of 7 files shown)


d2l-utility-behavior.html
```diff
@@ -55,14 +55,23 @@
...
-			// An enrollment's self href should be unique, so use it as an identifier
-			var selfLink = enrollment.getLinkByRel('self');
-			return selfLink.href;
+		getEntityIdentifier: function(entity) {
+			// An entity's self href should be unique, so use it as an identifier
+			if (entity.getLinkByRel) {
+				var selfLink = entity.getLinkByRel('self');
+				return selfLink.href;
+			} else {
+				var links = entity.links || [];
```
jopnick: do you have in mind a case where the entity wont have that function? (also, nit: `'undefined' === typeof entity.getLinkByRel` feels more explicit.  If this is a case for this, is there a corresponding test?

dbatiste:
Sort of unrelated to this change: I noticed that a couple of behaviors are added to the `Polymer` object.  I think we should leave that alone.  Rather, something like:  `window.D2L.MyCourses...`.

ryantmer:
@dbatiste Yeah, when we first started this I don't know if we had come to the `window.D2L.MyCourses` format yet (or if we had, Adam/Joel/myself missed it), but I do agree, should be updated to be consistent with other repos.

d2l-utility-behavior.html
ryantmer: The only case I know of is if the entity isn't `node-siren-parse`-d before going through here; which, no explicit case, but might be possible?

ryantmer: Now that I've said it out loud actually, yagni. If something comes through that isn't `node-siren-parse`-d, there's something wrong so it should be fixed at the source, not here.

dbatiste:
No worries.  :)  

I didn't look at what these functions are doing, but if there's something we definitely see being reusable, it can be added to our [d2l-polymer-behaviors](https://github.com/Brightspace/d2l-polymer-behaviors-ui).  If we're not sure, I'd keep it local to the app until we're sure.

ryantmer:
Ping for potential further feedback? 😀

ryantmer:
@dbatiste #157 addresses the moving of behaviours, btw :)

apalaniuk:
LGTM!

jopnick:
LGTM2

cgalvind2l:
LGTM

PR #151, opened 2016-10-14
Remove custom d2l-menu items

ryantmer:
Removing the custom menu item for the sorting menu, and applying `d2l-menu-item-selectable-styles` to the custom item for the filter menu.

After some discussion in https://github.com/Brightspace/d2l-menu-ui/pull/17 and offline, it was decided that the filter menu checkboxes aren't really _menu items_ per se, so they're remaining as custom items. However, their use is limited to the all courses overlay, hence keeping them there for now. If there is a use case to use them elsewhere, they should probably be broken out into their own repo.

PR #154, opened 2016-10-18
US76123 UI fixes

ryantmer:
This addresses Bob's UI feedback in US76123, namely:

- Remove space between menu caret and text
- Vertically align checkbox and text
- Search input too wide - should be 5/12 width
- Don't show recent searches dropdown if there are none
- Space between search and "Pinned" should be 50px
- Smaller search in dropdowns (45px tall)
- Clicking a checkbox's label doesn't click the checkbox
- Selecting a value in the sort dropdown should apply it and close the dropdown
- Clicking dropdown text should toggle dropdown, not always open it

Some of these were likely already addressed by the switch to using "proper" `d2l-menu` custom components in #129. (Which will be even further improved by #151.)

There is also a note about weird scrolling in the dropdowns, but I cannot reproduce this. I suspect it was a Mac/Safari thing, and is likely fixed now with the new `d2l-menu` components.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -83,6 +83,7 @@
 			:host {
 				display: block;
 				padding: 0.9rem 1rem;
+				cursor: pointer;
 			}
 			:host span {
 				line-height: 1.2rem;
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -338,14 +338,15 @@
 			<div id="filterAndSort">
 				<div id="filterSection">
 					<span id="filterText" class="dropdown-opener-text" on-tap="_toggleFilterDropdown">{{localize('filtering.filter')}}</span>
-					<d2l-dropdown id="filterDropdown">
+					<d2l-dropdown id="filterDropdown" no-auto-open>
 						<button
 							class="d2l-dropdown-opener dropdown-button"
 							on-focus="_focusFilterText"
 							on-blur="_blurFilterText"
 							on-mouseenter="_focusFilterText"
 							on-mouseleave="_blurFilterText"
-							aria-labelledby="filterText">
+							aria-labelledby="filterText"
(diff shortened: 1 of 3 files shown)


d2l-all-courses.html
```diff
@@ -305,16 +285,27 @@
 				width: 80%;
 				margin: auto;
 			}
-			button:focus,
-			button:hover,
-			.highlight {
-				text-decoration: underline;
+			d2l-dropdown-content d2l-search-widget {
+				--d2l-search-widget-height: 45px;
+				margin-left: 20px;
```
jstefaniuk-d2l: `margin: 10px 20px`

ryantmer: Ah, yeah. I just moved it in this case, but I think we talked about this previously. I shall oopdate.

jstefaniuk-d2l:
Code LGTM as far as I can tell

ryantmer:
Side note: since I was "in the area", I was going to move the styling out, but removing `d2l-menu-item-sort` and `d2l-menu-item-filter` (#151) will remove ~150 lines, so meh.

jstefaniuk-d2l:
Took a look, All the notes seem to be addressed

I would love to see something other than the text select cursor in the sort menu, but that doesn't really need to be done in this PR (Although it is a pretty short fix)

None of the sort options (Or filters) seem to be working for me, but as far as I can tell this PR is just the ui stuff (Maybe it's a quirk of the test data?)

LGTM

ryantmer:
Good catch - I'll update to make that a `cursor: pointer`.

As for things seeming not to work, the sort/filter only apply to the live search options currently - search isn't reflected in the course tiles, so it seems like nothing is happening visually. [US77177](https://rally1.rallydev.com/#/51191528503d/detail/userstory/63425703478?fdp=true) is the story to make your search apply to the tiles (which will also apply sorting and filtering there).

PR #153, opened 2016-10-18
US77856 - Image select success/failure animation

jstefaniuk-d2l:
✨ ✨

PR #155, opened 2016-10-18
Us78203 - Fix the image change not reflecting when the back button is hit

jstefaniuk-d2l:
Based off: https://github.com/Brightspace/d2l-my-courses-ui/pull/153

Refetch the image after the browser response so that we don't cache the wrong value

- [x] To fix: Currently won't work if they click faster than a second after the response


CodeBaboon:
Code and Testing LGTM

PR #156, opened 2016-10-18
The overlay shouldn't glitch out if there is no title provided

jstefaniuk-d2l:
This helps solve some of the strange behaviour we were seeing with overlays on @CodeBaboon's computer (Turns out that something was leaving his overlay without a title)

### DIFF
--- a/d2l-simple-overlay-styles.html
+++ b/d2l-simple-overlay-styles.html
@@ -28,9 +28,14 @@
 			margin: 0;
 			padding: 0 0 0.2rem 0;
 			color: var(--d2l-color-tungsten);
-			width: 24px;
-			height: 24px;
 			cursor: pointer;
+		}
+
+		.container {
+			display: flex;
+			justify-content: space-between;
+			align-items: center;
+			min-height: 105px;
 		}
 
 		.close-button d2l-icon {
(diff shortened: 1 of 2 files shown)


d2l-simple-overlay-styles.html
```diff
@@ -26,11 +26,19 @@
...
+			top: 0;
+			bottom: 0;
+			margin-top: auto;
+			margin-bottom: auto;
+			position: absolute;
+			right: 0;
+		}
+
+		.container {
+			position: relative;
```
ryantmer: Could this same result be achieved (in perhaps a simpler way) using `display: flex` along with `justify-content: space-between`?

jstefaniuk-d2l: Probably, will try!

d2l-simple-overlay-styles.html
```diff
@@ -26,19 +26,17 @@
...
-			right: 0;
 		}
 
 		.container {
-			position: relative;
+			display: flex;
+			justify-content: space-between;
+			align-items: center;
+			min-height: 105px;
+			flex-direction: row-reverse;
```
ryantmer: Could also just swap the order in HTML, I think? (I've never actually used `row-reverse` before though, so if it does something else as well, then neat!)

jstefaniuk-d2l:
@ryantmer good to merge?

ryantmer:
:+1:

PR #157, opened 2016-10-18
Move behaviour namespace to window.D2L.MyCourses

ryantmer:
While using Polymer.X.Y is recommended in some Polymer documentation, other D2L projects are using window.D2L.X.Y. This includes d2l-polymer-behaviors, which includes shared behaviours.

Just a straight-up replacement, so a cursory check is all that's really needed here.


jstefaniuk-d2l:
code lgtm

PR #158, opened 2016-10-19
Remove the tooltip

jstefaniuk-d2l:
The tooltip was showing up because we were using the `title` property on the overlay

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -108,7 +108,7 @@
 
 		<d2l-simple-overlay
 			id="all-courses"
-			title$="{{localize('allCourses')}}"
+			title-name="{{localize('allCourses')}}"
 			locale="[[locale]]"
 			with-backdrop>
 
@@ -131,7 +131,7 @@
 
 		<d2l-simple-overlay
 			id="basic-image-selector-overlay"
-			title$="{{localize('changeImage')}}"
+			title-name="{{localize('changeImage')}}"
 			locale="[[locale]]"
 			with-backdrop>
 			<d2l-basic-image-selector
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -19,21 +19,23 @@ The overlay supports using different animations and transitions for desktop and 
 <dom-module id="d2l-simple-overlay">
 	<style include="d2l-simple-overlay-styles"></style>
 	<template>
-		<div class="max-width container">
-			<h1 class="d2l-heading-2">{{title}}</h1>
-			<button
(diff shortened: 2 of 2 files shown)


d2l-my-courses.html
```diff
@@ -108,7 +108,7 @@
 
 		<d2l-simple-overlay
 			id="all-courses"
-			title$="{{localize('allCourses')}}"
+			title-name="{{localize('allCourses')}}"
```
ryantmer: `header` might make more sense, or `header-text` or something

d2l-simple-overlay.html
```diff
@@ -19,21 +19,23 @@
...
-				on-tap="_handleClose"
-				aria-label$="{{localize('closeSimpleOverlayAltText')}}"
-				alt$="{{localize('closeSimpleOverlayAltText')}}">
-				<d2l-icon icon="d2l-tier1:close-large-thick"></d2l-icon>
-			</button>
-		</div>
-		<div class="scrollable">
-			<div class="max-width">
-				<content></content>
+		<span role="dialog" aria-label$="[[titleName]]">
```
ryantmer: I'm not super familiar with `aria-label` - is it common for it to be applied to a span? I would think that generally the text content of the span is used to describe it (which there isn't any in this case, leading me to wonder why we need it at all).

d2l-my-courses.html
jstefaniuk-d2l: This is essentially acting as the title of the dialog (for screen reader as well)

d2l-simple-overlay.html
jstefaniuk-d2l: Aria-label is here so that the dialog has a name for screen readers, it's the label for "role=dialog" rather than the span

jstefaniuk-d2l: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role

d2l-my-courses.html
ryantmer: Yarr, I was just referring to the name of the property is all. Seems a bit awks to me.

d2l-simple-overlay.html
ryantmer: Ah, gotcha gotcha. I think for the label since it's set by a property you have to do it this way, but for the role you could also use `hostAttributes`. Not better/worse, but just in case you didn't know about `hostAttributes`! :D

ryantmer:
LGTM :shipit:

PR #159, opened 2016-10-19
US77513 - Utilize the appendMore querystring to fill out default image result pages

jstefaniuk-d2l:
Requires: https://github.com/Brightspace/course-image-catalog/pull/48/files

### DIFF
--- a/image-selector/d2l-basic-image-selector.html
+++ b/image-selector/d2l-basic-image-selector.html
@@ -65,7 +65,6 @@
 				this._images = this._defaultImages;
 				this.$$('d2l-search-widget').addEventListener('search-results-changed', function(response) {
 					var searchWidget = this.$$('d2l-search-widget');
-					this._searchImages = searchWidget.searchResults.entities || [];
 
 					var delimiter = '$$$DELIMITER???',
 						searchRegex = /search=([^&]*)/,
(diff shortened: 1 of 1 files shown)


trevorjonesd2l:
👀 👍

image-selector/d2l-basic-image-selector.html
```diff
@@ -75,6 +74,16 @@
 						userSearchText = (match.length > 1) ? match[1] : null,
 						noResultsSplit = noResultsText.split(delimiter);
 
+					// Throw out empty searches
+					// TODO: d2l-search-widget property that sends events instead of making network calls
+					if (match[1] === "") {
+						this._searchImages = [];
+						this._showGrid = true;
+						this._updateImages();
```
CodeBaboon: Both this section and the part below both end by executing `this._updateImages();`. Wonder if perhaps an `if/else` construct wouldn't be better than an early return, then stick the `this._updateImages();` in after so that both paths run it and then exit. Might want to also then split the insides of each condition into their own methods? So perhaps ideally in the end you'd have something like:

```
if (userSearchText) {
    DisplaySearchResults(); // all the noResults stuff and the searchWidget thing
} else {
    DisplayDefaultResults(); // the 'throw out empty' stuff
}

this._updateImages();
```

cgalvind2l:
LGTM

PR #160, opened 2016-10-20
DE23046: Update typography to version 5.

njostonehouse:


d2l-all-courses.html
```diff
@@ -204,7 +204,7 @@
 				flex: 1
 			}
 			.dropdown-text {
-				@apply(--d2l-small-text);
+				@apply(--d2l-body-small-text);
 				/* overrides to d2l-small-text */
```
ryantmer: Mind updating this comment (and the other similar ones?) There may also be no-longer-required properties in here if the new stuff sets fewer/different things :)

njostonehouse:
Nice, a legitimate failure because I haven't release search widget .18 yet! Excellent.

PR #164, opened 2016-10-24
Update focusable override stuff

apalaniuk:
No story, just removes some dead code I noticed (`focusableNodesOverride` is never called as `_focusableNodes` does the same thing in a much simpler and more generic way directly). Also updates the `iron-overlay-behavior` dependency to the release with the updates in the pinned commit we're currently using.


ryantmer:
`+1 -46` aw yiss.

Code LGTM, I shall also test right meow.

PR #166, opened 2016-10-24
Reset search/image tiles when opening image selector

apalaniuk:
Re-opening the image selector won't clear the search if the same course is selected twice in succession due to how Polymer property observers work. Changed observer to a public setter that's called when the overlay is opened, as the organization should only change while the overlay is closed.

### DIFF
--- a/d2l-simple-overlay.html
+++ b/d2l-simple-overlay.html
@@ -48,6 +48,16 @@ The overlay supports using different animations and transitions for desktop and 
 			* @event recalculate-columns
 			*/
 
+			/*
+			* Fired on distributed children when overlay is about to open
+			* @event d2l-simple-overlay-opening
+			*/
+
+			/*
+			* Fired on distributed children when overlay has finished closing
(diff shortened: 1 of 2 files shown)


jstefaniuk-d2l:
I think we also have problems with fields not clearing out on the all-courses overlay (@ryantmer?)

I'm wondering if it would be a good idea to have `d2l-simple-overlay` call some kind of `overlayOpen` and `overlayClose` functions on it's children (if those functions exist on the child), and then the overlay can run these open and close functions when the overlay actually opens and closes rather than relying on some kind of "Set up" call to be made from d2l-my-courses like we're doing here

apalaniuk:
@jstefaniuk-d2l Good idea. This implementation is still coupled in that the image selector explicitly knows that the overlay is a thing that it can get events from, but that exists elsewhere in the same component already.

jstefaniuk-d2l:
Code LGTM

cgalvind2l:
LGTM

PR #167, opened 2016-10-25
Responsive image resolutions

trevorjonesd2l:
TODO:
- [x] Fix tests
- [x] Write tests
- [x] Picturefill

Use `srcset` to set the best image based on the image width, approximate width of the image container and the density of the screen. Each browser incorporates `srcset` a little different by how they determine the best image to display - chrome seem so optimize it better for network calls (with cache and lower quality images), but from my testing images will never be distorted (lower resolution than the container).

`srcset` is kind of tricky in how it's implemented with `sizes` being in units of `vw` - I did my best to approximate this based on the media query breakpoints and number of columns in the all courses overlay. The approximations are more accurate on mobile, where I feel the image resolution is a greater concern (mobile data impact), where as on desktop it becomes a bit more approximated (over approximated rather than under).

`srcset` isn't support by IE, in which case it will use the default image instead (note the default image could be reconfigured to default to low-max rather than high-max if that would make sense). Here is the support of it as a whole
image

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -51,7 +51,7 @@
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
     "polymer": "Polymer/polymer#^1.4.0",
-    "picturefill": "https://github.com/scottjehl/picturefill.git#3.0"
+    "picturefill": "https://github.com/scottjehl/picturefill.git#^3.0.0"
   },
   "devDependencies": {
     "web-component-tester": "^4.2.2"
--- a/bower.json
+++ b/bower.json
@@ -50,7 +50,8 @@
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
-    "polymer": "Polymer/polymer#^1.4.0"
+    "polymer": "Polymer/polymer#^1.4.0",
+    "picturefill": "https://github.com/scottjehl/picturefill.git#master"
   },
   "devDependencies": {
     "web-component-tester": "^4.2.2"
--- a/bower.json
+++ b/bower.json
@@ -51,7 +51,7 @@
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
     "polymer": "Polymer/polymer#^1.4.0",
-    "picturefill": "https://github.com/scottjehl/picturefill.git#master"
+    "picturefill": "https://github.com/scottjehl/picturefill.git#3.0"
   },
   "devDependencies": {
     "web-component-tester": "^4.2.2"
--- a/d2l-course-tile-grid.html
+++ b/d2l-course-tile-grid.html
@@ -24,8 +24,7 @@
 					class="grid-child"
 					enrollment-id="[[getEntityIdentifier(item)]]"
 					delay-load="[[delayCourseTileLoad]]"
-					all-courses="[[allCourses]]"
-					num-cols="[[_numCols]]">
+					tile-sizes="[[tileSizes]]">
 				</d2l-course-tile>
 			</template>
 		</div>
@@ -56,11 +55,6 @@
 				* Specifies whether or not the tile grid is for the all courses overlay
 				* @type {Boolean}
 				*/
-				allCourses: Boolean,
-				/*
-				* Sets the delay-load property on the course tiles
-				* @type {Boolean}
-				*/
 				delayCourseTileLoad: Boolean,
 				/*
 				* Set of enrollment entities for which to display course tiles
@@ -83,7 +77,12 @@
 					value: function() {
 						return [];
 					}
-				}
+				},
+				/*
+				* Size the tile should render with respect to vw
+				* @type {String}
+				*/
+				tileSizes: String
 			},
 			behaviors: [
 				window.D2L.MyCourses.CourseTileResponsiveGridBehavior,
--- a/d2l-course-tile-grid.html
+++ b/d2l-course-tile-grid.html
@@ -23,7 +23,9 @@
 					animate-insertion="[[_removeEnrollmentFromTransitionList(item)]]"
 					class="grid-child"
 					enrollment-id="[[getEntityIdentifier(item)]]"
-					delay-load="[[delayCourseTileLoad]]">
+					delay-load="[[delayCourseTileLoad]]"
+					all-courses="[[allCourses]]"
+					num-cols="[[_numCols]]">
 				</d2l-course-tile>
 			</template>
 		</div>
@@ -50,6 +52,11 @@
 			*/
 
 			properties: {
+				/*
+				* Specifies whether or not the tile grid is for the all courses overlay
+				* @type {Boolean}
+				*/
+				allCourses: Boolean,
 				/*
 				* Sets the delay-load property on the course tiles
 				* @type {Boolean}
--- a/d2l-course-tile-grid.html
+++ b/d2l-course-tile-grid.html
@@ -52,7 +52,7 @@
 
 			properties: {
 				/*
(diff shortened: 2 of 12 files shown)


d2l-course-tile-grid.html
```diff
@@ -23,7 +23,9 @@
 					animate-insertion="[[_removeEnrollmentFromTransitionList(item)]]"
 					class="grid-child"
 					enrollment-id="[[getEntityIdentifier(item)]]"
-					delay-load="[[delayCourseTileLoad]]">
+					delay-load="[[delayCourseTileLoad]]"
+					all-courses="[[allCourses]]"
+					num-cols="[[_numCols]]">
```
apalaniuk: Just a thought: I wonder if this could be simplified by passing in the `sizes` string directly, rather than having to pass down extra information to `d2l-course-tile` and resolve it there. Then the tile wouldn't have to know anything explicit about being in an overlay, or grid layout, etc.

CodeBaboon:
We should probably use [PictureFill](https://scottjehl.github.io/picturefill/) (v3) to resolve the compatibility issues.

trevorjonesd2l:
I'll look at plugging in PictureFill now 👍

d2l-utility-behavior.html
```diff
@@ -39,6 +39,33 @@
...
+			});
+			if (imageClass === 'tile') {
+				srcset += sizes.lowMin ? sizes.lowMin + ' 145w, ' : '';
+				srcset += sizes.highMin ? sizes.highMin + ' 290w, ' : '';
+				srcset += sizes.lowMid ? sizes.lowMid + ' 220w, ' : '';
+				srcset += sizes.highMid ? sizes.highMid + ' 440w, ' : '';
+				srcset += sizes.lowMax ? sizes.lowMax + ' 540w, ' : '';
+				srcset += sizes.highMax ? sizes.highMax + ' 1080w, ' : '';
+			}
+			if (imageClass === 'banner') {
```
CodeBaboon: What about `narrow` v `wide`, do we need to account for that?

trevorjonesd2l: Good catch - using the `narrow` class for the banner images instead of `banner` could eliminate any chance of `wide` banner images showing up for these.

d2l-course-tile.html
```diff
@@ -510,6 +530,23 @@
 					dropdown.close();
 					this._setCourseTileHovered(false);
 				}
+			},
+			setSizes: function() {
+				if (this.allCourses) {
+					switch(this.numCols) {
```
CodeBaboon: This feels like it could be less switchy and more mathy.

```
var pctWidth = Math.floor(100 / this.numCols);
this._sizes = pctWidth + "vw";
```

d2l-course-tile.html
```diff
@@ -11,6 +11,7 @@
 <link rel="import" href="d2l-course-tile-styles.html">
 <link rel="import" href="localize-behavior.html">
 <link rel="import" href="d2l-utility-behavior.html">
+<script src="bower_components/picturefill/src/picturefill.js"></script>
```
CodeBaboon: I'm curious if this de-dupes when it goes through the web component shard process the way our other common libraries will. Anyone know?

apalaniuk: @CodeBaboon Vulcanize claims this, so we should be okay?

> The Vulcanize tool follows HTML Imports and <script> tags to inline these external assets into a single page, to be used in production.

d2l-course-tile.html
```diff
@@ -218,6 +224,11 @@
 				_courseSettingsLabel: String,
 				_canChangeCourseImage: Boolean,
 				/*
+				* Size property to be used for srcset responsive images
+				* @type {String}
+				*/
+				_sizes: String,
```
ryantmer: Doesn't appear to be used (at least not in the diff)

bower.json
```diff
@@ -50,7 +50,8 @@
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
-    "polymer": "Polymer/polymer#^1.4.0"
+    "polymer": "Polymer/polymer#^1.4.0",
+    "picturefill": "https://github.com/scottjehl/picturefill.git#master"
```
ryantmer: Can this go to a specific release instead? Just a bit safer than going straight to master. (Currently `#3.0.2`)

ryantmer:
Testing notes:

- W8.1
  - Chrome ☀️ 
  - FF ☀️ 
  - IE11 ☀️ 
- W10
  - IE11 ☀️ 
  - Edge ⛅️ Works, but seems to "stick" to low density max for longer than other browsers. Not a deal breaker I'd say, as get it large enough and it does bump up, just do so at a wider res than other browsers
- OSX
  - Safari :cloud_with_lightning: Only selects image on page load - resizing the window has no effect on the image used

d2l-utility-behavior.html
```diff
@@ -39,19 +39,37 @@
 
 			return queryString ? action.href + '?' + queryString : action.href;
 		},
+		getImageSrcset: function(image, imageClass) {
+			if (!image) {
+				return '';
+			}
+			var srcset = '',
+				sizes = this._getImageLinks(image, imageClass);
+			if (imageClass === 'tile') {
```
jstefaniuk-d2l: you could then go and do something like:

```
['lowMin', 'highMin', 'lowMid', ...].forEach(function(size){
     srcset += sizes[size] ? sizes[size]+ imageSets[imageClass][size] : '';
});
```

jstefaniuk-d2l:
I'm thinking that we should move to make all this resizing stuff to it's own polymer component which can easily be thrown anywhere. (just passing in an image size and the image json or organization)

This new requirement isn't necessarily part of this PR, but I do feel that it's something we need to have at some point

d2l-utility-behavior.html
```diff
@@ -39,19 +39,32 @@
 
 			return queryString ? action.href + '?' + queryString : action.href;
 		},
+		getImageSrcset: function(image, imageClass) {
```
jstefaniuk-d2l: we should check somewhere if passed in imageClass exists

d2l-utility-behavior.html
```diff
@@ -39,19 +39,32 @@
...
+				imageWidths = {
+					tile: {
+						lowMin: 145, lowMid: 220, lowMax: 540,
+						highMin: 290, highMid: 440, highMax: 1080
+					},
+					narrow: {
+						lowMin: 320, lowMid: 375, lowMax: 767,
+						highMin: 640, highMid: 750, highMax: 1534
+					}
+				};
```
jstefaniuk-d2l: Maybe not a thing for this PR, but we should fill in the rest of the image sizes (Maybe a good idea for when we split it out into a polymer component)

bower.json
```diff
@@ -50,7 +50,8 @@
     "iron-pages": "^1.0.8",
     "neon-animation": "PolymerElements/neon-animation#^1.2.2",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
-    "polymer": "Polymer/polymer#^1.4.0"
+    "polymer": "Polymer/polymer#^1.4.0",
+    "picturefill": "https://github.com/scottjehl/picturefill.git#3.0"
```
ryantmer: Pretty sure `^3.0.0` works with git addresses (and would mean we can get patches and minor updates for free, yey)

d2l-utility-behavior.html
```diff
@@ -39,19 +39,32 @@
 
 			return queryString ? action.href + '?' + queryString : action.href;
 		},
+		getImageSrcset: function(image, imageClass) {
+			if (!image) {	return ''; }
```
ryantmer: nit time: weird spacing

ryantmer:
Yeah, seems legit. Still doesn't really have an effect on Safari, but since we're defaulting to larger-than-required images, that's not the end of the world I don't think. The only situation where it would be noticeable is if a user loads the page small, then expands it to just the right size, which is a pretty weird edge case to waste any time on (especially since it's only in Safari).

ryantmer:
:+1:

Issue #99, opened 2016-08-10
Unpin Firefox (again)


ryantmer:
Because reasons, https://github.com/Polymer/wct-local/commit/c3bfc75013a2d0b42700c805f3da57162e4f2017#diff-a450911996050a8e4251c5c56f7428d4R129 is a thing, so `wct-local` doesn't currently work with any versions of FF > 46. So, we're pinned to 46.0 in Travis CI until this gets fixed. The bug for this is being tracked at https://github.com/Polymer/wct-local/issues/38.

PR #170, opened 2016-10-26
trying out fixes for wct

jstefaniuk-d2l:


ryantmer:
Not surprisingly, we're not the only ones - https://github.com/Polymer/web-component-tester/issues/417

jstefaniuk-d2l:
Seems to work as far as I can tell!

ryantmer:
Sweet! Merge away :D

Might also want to make a GH issue to track this (like #99) so that we don't forget about it once things are fixed upstream.

PR #172, opened 2016-10-27
blue the upload button (on hover & focus)

jstefaniuk-d2l:
image

### DIFF
--- a/image-selector/d2l-basic-image-selector.html
+++ b/image-selector/d2l-basic-image-selector.html
@@ -40,6 +40,12 @@
 			.upload:focus .upload-text,
 			.upload:hover .upload-text {
 				text-decoration: underline;
+				color: var(--d2l-color-celestine);
+			}
+
+			.upload:focus .upload-icon,
+			.upload:hover .upload-icon {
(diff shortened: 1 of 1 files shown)


trevorjonesd2l:
Code looks 🔵K

jstefaniuk-d2l:
Meets Bob's approval

image-selector/d2l-basic-image-selector.html
```diff
@@ -40,6 +40,12 @@
 			.upload:focus .upload-text,
 			.upload:hover .upload-text {
 				text-decoration: underline;
+				color: var(--d2l-color-celestine);
+			}
+
+			.upload:focus .upload-icon,
```
dbatiste: Minor thing: looks like the class is probably not needed on the icon, and this could just be `.upload:focus d2l-icon`.

jstefaniuk-d2l: I don't mind it being a bit more specific, now I can easily ctrl-f to all instances of the class, and don't have to go back and rename all the references of `.upload d2l-icon` if we add a second icon inside upload.

PR #176, opened 2016-10-27
ids are no longer a thing

jstefaniuk-d2l:


ryantmer:
Oh crap! This is totally my fault, missed that somehow. LGTM, thanks!

apalaniuk:
Mayhaps this is a good time (or, better late than never) to update the test data, [since it still uses IDs](https://github.com/Brightspace/d2l-my-courses-ui/blob/700396e1b46ae9849b1f83fb923363c48867fc12/test/d2l-all-courses/d2l-all-courses.js#L55).

jstefaniuk-d2l:
Missed your edit, although I only found 2 and there's probably more

Anyone else tried this out?

apalaniuk:
Code LGTM as well, I can test.

apalaniuk:
Tested in Chrome 53/FF46/IE11 in Win 8.1, Edge 38 in Win10 VM, and Safari 9.1 in OSX 10.11 over RDP.

Tested changing first, last and random middle tile image in main My Courses view - works without issue.

Tested changing first, last and middle tile in each of the pinned/unpinned course lists in the All Courses view - animation/new image displayed on selected tile as expected.

PR #179, opened 2016-11-02
Us77220 telemetry

cgalvind2l:
Has:

- [x] event on launch of change image UI
- [x] event on selection of an image in the change image UI
- [x] event on lazy load of next page of images
- [x] event on press of the upload button

Needs still:
- [x] event on search
- [x] add user information to the events
- [x] de-hard-code the telemetry URL so that we can also switch to prod when necessary
- [x] tests

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -46,8 +46,7 @@ in that organization - student, teacher, TA, etc.
 			id="telemetryOpenRequest"
 			url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
 			body="[[_telemetryOpenEvent]]"
-			method="POST"
-			on-iron-ajax-response="_onTelemetryResponse">
+			method="POST">
 		</d2l-ajax>
 
 		<div class="tile-container"
@@ -438,9 +437,6 @@ in that organization - student, teacher, TA, etc.
 					dropdown.close();
 					this._setCourseTileHovered(false);
 				}
-			},
-			_onTelemetryResponse: function(response) {
-				console.log(response);
 			}
 		});
 	</script>
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -46,6 +46,7 @@ in that organization - student, teacher, TA, etc.
 			id="telemetryRequest"
 			url="[[telemetryEndpoint]]"
 			body="[[_telemetryEvent]]"
+			headers='{ "Content-Type": "application/json" }'
 			method="POST">
 		</d2l-ajax>
 
@@ -322,8 +323,8 @@ in that organization - student, teacher, TA, etc.
 			},
 			_launchCourseTileImageSelector: function(e) {
 				this._telemetryEvent = JSON.stringify({
-					type: 'LaunchChangeImage',
-					ts: Date.now(),
+					name: 'LaunchChangeImage',
+					ts: Math.round(Date.now() / 1000),
 					userId: this.userId,
 					tenantId: this.tenantId
 				});
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -42,6 +42,14 @@ in that organization - student, teacher, TA, etc.
 			on-iron-ajax-error="_onEnrollmentPinError">
 		</d2l-ajax>
 
+		<d2l-ajax
+			id="telemetryOpenRequest"
+			url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
+			body="[[_telemetryOpenEvent]]"
+			method="POST"
+			on-iron-ajax-response="_onTelemetryResponse">
+		</d2l-ajax>
+
 		<div class="tile-container"
 			on-mouseover="_hoverCourseTile"
 			on-mouseout="_onTileMouseOut">
@@ -154,7 +162,8 @@ in that organization - student, teacher, TA, etc.
 						className: '',
 						iconName: ''
 					}
-				}
+				},
+				_telemetryOpenEvent: String
 			},
 			behaviors: [
 				window.D2L.MyCourses.LocalizeBehavior,
@@ -305,6 +314,13 @@ in that organization - student, teacher, TA, etc.
 				}
 			},
 			_launchCourseTileImageSelector: function(e) {
+				this._telemetryOpenEvent = JSON.stringify({
+					type: 'LaunchChangeImage',
+					ts: Date.now()
+				});
+
+				this.$.telemetryOpenRequest.generateRequest();
+
 				e.preventDefault();
 				e.stopPropagation();
 
@@ -417,6 +433,9 @@ in that organization - student, teacher, TA, etc.
 					dropdown.close();
 					this._setCourseTileHovered(false);
 				}
+			},
+			_onTelemetryResponse: function(response) {
+				console.log(response);
 			}
 		});
 	</script>
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -43,9 +43,9 @@ in that organization - student, teacher, TA, etc.
 		</d2l-ajax>
 
 		<d2l-ajax
-			id="telemetryOpenRequest"
-			url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
-			body="[[_telemetryOpenEvent]]"
+			id="telemetryRequest"
+			url="[[telemetryEndpoint]]"
+			body="[[_telemetryEvent]]"
 			method="POST">
 		</d2l-ajax>
 
@@ -129,6 +129,7 @@ in that organization - student, teacher, TA, etc.
 					type: Boolean,
 					observer: '_pinnedChanged'
 				},
+				telemetryEndpoint: String,
 				tenantId: String,
 				// Size the tile should render with respect to vw
 				tileSizes: Object,
@@ -169,7 +170,7 @@ in that organization - student, teacher, TA, etc.
 				* on the opening of the change image UI
 				* @type {String}
 				*/
-				_telemetryOpenEvent: String
+				_telemetryEvent: String
 			},
 			behaviors: [
 				window.D2L.MyCourses.LocalizeBehavior,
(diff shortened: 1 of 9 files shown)


jstefaniuk-d2l:
Just a note, rather than merging master into your branch, rebase your branch on master (`git rebase master`), this is often a bit more work, but helps keep the commit log nice and readable. (Merging is sometimes ok for super ugly diffs though)

image-selector/d2l-basic-image-selector.html
```diff
@@ -80,6 +80,20 @@
...
+			method="POST">
+		</d2l-ajax>
+
+		<d2l-ajax
+			id="telemetryUploadRequest"
+			url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
+			body="[[_telemetryUploadEvent]]"
+			method="POST">
+		</d2l-ajax>
+
```
trevorjonesd2l: Since we're making calls to the same telemetry service maybe here it'd be better to have one `d2l-ajax` that can be referred to by each event?
```html
<d2l-ajax
    id="telemetryRequest"
    url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
    body="[[_telemetryEvent]]"
    method="POST">
 </d2l-ajax>
```
So for each event you could send a request with:
```js
this._telemetryEvent = JSON.stringify({
    ...
});
this.$.telemetryRequest.generateRequest();
```

image-selector/d2l-basic-image-selector.html
```diff
@@ -91,7 +105,7 @@
 			</d2l-search-widget>
 
 			<template is="dom-if" if="{{_organizationChangeImageHref}}" restamp="true">
-				<a class="upload" href$=[[_organizationChangeImageHref]]>
+				<a class="upload" href$=[[_organizationChangeImageHref]] on-tap="_uploadTelemetry">
```
trevorjonesd2l: `on-tap` won't fire with keyboard interaction. You will also need `on-keydown` mapped to the enter button (since it's a link).

```html
<... on-keydown="_uploadKeydown">
```

```js
_uploadKeydown: function(e) {
    if (e.keyCode === 13) { // Enter keycode
        this._uploadTelemetry();
    }
}
```

d2l-course-tile.html
```diff
@@ -44,6 +44,13 @@
 			on-iron-ajax-error="_onEnrollmentPinError">
 		</d2l-ajax>
 
+		<d2l-ajax
+			id="telemetryOpenRequest"
+			url="https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7"
```
CodeBaboon: Would be cool if this URL were a constant or passed-down-property or something like.

image-selector/d2l-basic-image-selector.html
```diff
@@ -214,6 +230,17 @@
 				if (loadMore) {
 					newImages = (isDefault ? this._defaultImages : this._searchImages) || [];
 					newImages = newImages.concat(parsedResponse.entities);
+					var searchRegex = /search=([^&]*)/,
```
CodeBaboon: Probably would be cleaner to wrap this whole chunk of stuff into it's own `_doNextPageTelemetryEventThing()`.

image-selector/d2l-basic-image-selector.html
```diff
@@ -214,6 +230,17 @@
...
+						match = searchRegex.exec(this._searchString);
+					var searchWidget = this.$$('d2l-search-widget');
+					this._telemetryNextPageEvent = JSON.stringify({
+						ts: Date.now(),
+						type: 'NextImagePageNavigation',
+						resultsPageNumber: parsedResponse.properties.pagingInfo.page,
+						totalResults: parsedResponse.properties.pagingInfo.total,
```
CodeBaboon: The paranoid part of me gets worried about 'what if `properties` or `pagingInfo` is undefined?'.

image-selector/d2l-image-selector-tile.html
```diff
@@ -105,6 +112,13 @@
 			_selectImage: function() {
 				this._setImageUrl = this._getSetImageUrl();
 				this.$.setImageRequest.generateRequest();
+				this._telemetrySetImageEvent = JSON.stringify({
```
CodeBaboon: This chunk could be it's own method too, just to keep the functions more succinct.

d2l-course-tile.html
cgalvind2l: That's on the TO-DO list

CodeBaboon: Doh! My bad #reading-fail

image-selector/d2l-basic-image-selector.html
ryantmer: `totalResults: parsedResponse && parsedResponse.properties && parsedResponse.properties.pagingInfo && parsedResponse.properties.pagingInfo.total || someDefaultValue`? :D

CodeBaboon: I don't know if it's better, but I was thinking maybe something like:

```
var pagingInfo = getPagingInfo(parsedResponse);
this._telemetryNextPageEvent = JSON.stringify({
    ts: Date.now(),
    type: 'NextImagePageNavigation',
    resultsPageNumber: pagingInfo.page,
    totalResults: pagingInfo.total
    search: searchstuff
});

function getPagingInfo(parsedResponse) {
    const pagingInfo = {
        page = 'error retrieving page info',
        total = 'error retrieving total info'
    };
    if (parsedResponse.properties && parsedResponse.properties.pagingInfo) {
        pagingInfo = parsedResponse.properties.pagingInfo;
    }
    return pagingInfo;
}
```

test/d2l-image-selector-tile/d2l-image-selector-tile.js
```diff
@@ -1,22 +1,40 @@
...
 describe('<d2l-image-selector-tile>', function() {
 	var widget,
 		testPath = 'http://testPath',
 		testImageHref = 'http://test.com/images/10',
-		testBadImageHref = 'http://test.com/MUAHAHAHA/10';
+		testBadImageHref = 'http://test.com/MUAHAHAHA/10',
+		telemetryUrl = 'https://telemetry-dev.cloud.desire2learn.com/api/events/c4d46116-d70c-41cc-99f0-607bc86424a7',
+		server;
+
+	before(function() {
```
jstefaniuk-d2l: I try not to use `before` when `beforeEach` works as well, since you usually want everything to be reinitialized between tests (You can run into issues where tests modify something and start colliding with eachother)

trevorjonesd2l:
For the constant `telemetryUrl` I'm under the impression (from @ryantmer) you'll need to set a config variable in the LMS, similar to what's done with `d2l.Tools.Quizzes.TelemetryEndpoint`. I'm not sure how this part works but you'll have to set it up so that is uses the `dev-telemetryUrl` for dev instances and the `prod-telemetryUrl` for prod instances.

With the `telemetryUrl` as a config variable you can pass it to the `d2l-my-courses` FRA as a parameter. You can access it in the project from `d2l-my-courses.html` and add it as a `property`. From there you'll have to cascade it down to the components that use telemetry.

cgalvind2l:
Now you need a new branch in the core for this: https://git.dev.d2l/projects/CORE/repos/le/pull-requests/6115/overview

So close to finishing this story...so...close...

image-selector/d2l-basic-image-selector.html
```diff
@@ -328,6 +343,62 @@
...
+			_uploadTelemetry: function(e) {
+				if((e.type === 'keydown' && e.keyCode === 13) || (e.type === 'tap')) {
+					console.log(e.type);
+					this._telemetryEvent = JSON.stringify({
+						ts: Date.now(),
+						type: "CourseImageUpload",
+						userId: this.userId,
+						tenantId: this.tenantId
+					});
+					this.$.telemetryRequest.generateRequest();
```
ryantmer: Question: I'm assuming the answer is yes, but - does this work as intended? I'm just imagining that since this is triggered by a link, the network call might not actually happen before the page navigates, thereby cancelling it. If it works in practice, then good enough for me!

image-selector/d2l-basic-image-selector.html
```diff
@@ -328,6 +343,62 @@
...
+        			page: 'error retrieving page info',
+        			total: 'error retrieving total info'
+    			};
+    			if (parsedResponse.properties && parsedResponse.properties.pagingInfo) {
+        			pagingInfo = parsedResponse.properties.pagingInfo;
+    			}
+    			return pagingInfo;
+			},
+			_doTelemetrySearchRequest: function(userSearchText) {
+				var searchWidget = this.$$('d2l-search-widget'),
```
ryantmer: Thought: looking at this, I'm like, this would actually be super legit to have in the actual search widget. But, certainly not a thing for here.

d2l-course-tile.html
```diff
@@ -391,6 +406,15 @@
...
 			},
 			_launchCourseTileImageSelector: function(e) {
+				this._telemetryOpenEvent = JSON.stringify({
+					type: 'LaunchChangeImage',
+					ts: Date.now(),
+					userId: this.userId,
+					tenantId: this.tenantId
+				});
+
+				this.$.telemetryOpenRequest.generateRequest();
```
cgalvind2l: Yeah I don't think that the widget particularly cares if the telemetry call succeeds or fails. It shouldn't affect anything.

image-selector/d2l-basic-image-selector.html
cgalvind2l: This does work, it's just hard to catch in the console unless you preserve the log.

cgalvind2l:
The merge instead of rebase was because of a confusing conflict.

jstefaniuk-d2l:
Code LGTM,

If you wanted, you could take the functions in `image-selector/d2l-basic-image-selector.html` and throw them in their own behaviour so that they're hidden away a bit, but that's not really necessary

(At some point we should do a bit of a refactoring sweep and put a bunch of stuff into behaviours/separate components)

jstefaniuk-d2l:
I volunteer to test this tomorrow morning

ryantmer:
Oh whoops - I have actually already checked this and the LE PR out and done a build - had an offline discussion about it with @cgalvind2l. I believe there's a small change still coming, but all that was here tested a-okay.

ryantmer:
Hokay! LP PR for this LGTM, but unfortunately I'm at home and that's causing auth issues, so I can't E2E test :(

ryantmer:
E2E check done - I'd say this is good to merge! (Although the LP one should probably be merged first)

PR #180, opened 2016-11-03
Remove picturefill

ryantmer:
This was causing some BSI build errors, and the performance benefit only really applies to IE11 - and only on small screens, at that. We can revisit this in the future if we so decide (although unlikely there's much point).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -7,4 +7,3 @@ globals:
   Polymer: false
   WCT: false
   D2L: false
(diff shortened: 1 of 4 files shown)


trevorjonesd2l:
Code looks good, this should be tested though

ryantmer:
Ran through some basic testing, nothing broke.

@ryantmer: LGTM.

PR #181, opened 2016-11-03
Fixed some nasty spacing on IE in a way Im not proud of

trevorjonesd2l:


jstefaniuk-d2l:
Any reason you're using Rem? Does everything stay in place if you increase the font-size on the body?

jstefaniuk-d2l:
That annoying moment when you go to search "rem body" and google corrects it to "rem body pillow" right as you hit enter

ryantmer:
Yep, tried it in a bunch of browsers (most notably IE11 and Edge) and looks good. Applies in IE11, nowhere else since they know how2flexbox properly.

PR #187, opened 2016-11-04
Don't use filters with Edge. Fixes #124.

### LINKED ISSUE #124: Animations not smooth on Edge
1. When you hover over a course tile it will flash white.
2. When you hover over the pin/unpin the course tile will be white

ryantmer:
Edge supports filters, but they act funny (flickering when hovering over the course tiles). To prevent this, just `filter:none;` when we're in Edge - means Edge users don't get the colourization effect, but that's better than the flickering by far (and Edge users don't get the colourization anyway right now).

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -118,6 +118,13 @@
 			filter: contrast(1.15);
 			transform: scale(1.1);
 		}
+		@supports (-ms-ime-align:auto) {
+			.course-image-container.hover,
+			.course-image-container.hover > .course-image img {
+				/* See https://github.com/Brightspace/d2l-my-courses-ui/issues/124 */
+				filter: none;
+			}
+		}
 		.course-text {
 			width: 100%;
 			max-height: 75px;


ryantmer:
Checked this in Edge, IE11, Chrome and Firefox. No effect in Chrome/FF, IE11 doesn't support CSS filters anyway (so no colourization, no flickering), and Edge now does the zoom without the flicker (no colourization, but that wasn't there before either).

d2l-course-tile-styles.html
```diff
@@ -118,6 +118,13 @@
 			filter: contrast(1.15);
 			transform: scale(1.1);
 		}
+		@supports (-ms-accelerator:true) {
```
apalaniuk: Apparently this [doesn't work for Edge 14](https://github.com/4ae9b8/browserhacks/issues/167), which I just tried (and didn't work). I did try `-ms-ime-align:auto` which does work on 14, but didn't test elsewhere.

ryantmer: Updated to use `-ms-ime-align:auto`, confirmed that it works in 13, and doesn't have any effect elsewhere.

apalaniuk: @ryantmer My personal laptop with Windows 10!

PR #188, opened 2016-11-07
MS Teams testing Travis integration

ryantmer:


ryantmer:
Going to actually merge this - we can remove it if we don't end up using Teams, but I'm curious as to what else this will do.

PR #189, opened 2016-11-07
US78151 - Don't fetch filter items for a small number of enrollments

ryantmer:
The call to fetch filters is expensive, and is largely pointless if a user only has a handful of enrollments. If a user only has a small number of enrollments, there's no need to do the (expensive) calls to fetch filters - if absolutely required, the search functionality should be enough.

To do:

- [x] Check with Bob to make sure 20 is the right magic number

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -625,6 +625,12 @@
 				this._onMoreResponse(response, '_semesterOrganizations', '_moreSemestersUrl', '_hasMoreSemesters');
 			},
 			_onOrganizationsResponse: function(response) {
+				if (this.pinnedEnrollments.length + this.unpinnedEnrollments.length <= 20) {
+					// Small optimization - hide filter menu and don't fetch contents if user doesn't have a lot of enrollments
+					this.toggleClass('hidden', true, this.$.filterAndSort);
+					return;
+				}
+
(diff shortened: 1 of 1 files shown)


apalaniuk:
Code LGTM. I wonder if we should wrap `filterSection` in a `dom-if` as well just so we don't initialize it for 99% of people, but I haven't measured any performance impact or anything.

ryantmer:
Ah, right - gave that a shot, but was reminded of [this issue](https://www.google.ca/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=dom-if+Invalid+value+used+as+weak+map+key) once again, wherein `dom-if` doesn't work well with dynamic content.

Did update it a bit to add the `_hasManyEnrollments` property though, as that's a bit nicer to look at.

apalaniuk:
:+1:

ryantmer:
@apalaniuk Redid it a bit, again. Back to the original way, since sometimes things weren't happening in quite the right order, causing the menus to disappear when they shouldn't.

Also changed to hide both filter and sort menus, as per Bob.

apalaniuk:
Tested with accounts over/under admin enrollment limit, looks good.

ryantmer:
Noice, thanks! :)

PR #193, opened 2016-11-11
Use caching of search responses

ryantmer:
This is pretty naive/shitty caching (just storing it within the component), so it won't persist across page loads/dom rebuild/refresh/anything really. The main advantage is for the empty query result, so if the search field is cleared then it won't re-fetch that result.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -37,7 +37,7 @@
     "d2l-menu": "^0.2.3",
     "d2l-offscreen": "^2.1.0",
     "d2l-polymer-behaviors": "<1.0.0",
-    "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#0.1.0",
+    "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#0.2.0",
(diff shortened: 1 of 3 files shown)


jstefaniuk-d2l:
I'm thinking this would be valuable in the images ui

ryantmer:
Pretty easy to turn on - just add `cache-responses` as an attribute to the search widget :D

jstefaniuk-d2l:
Code LGTM

ryantmer:
@jstefaniuk-d2l Mind giving a quick test? Should just not redo any network request for anything that has been previously searched for (including the empty search result set).

jstefaniuk-d2l:
Will test in 5 mins

ryantmer:
Schweet, thanks. Just added it to the image selector as well, as recommended.

jstefaniuk-d2l:
Network calls seem to be caching nicely, had to rebase on master to get the image selector working (just a note)

LGTM

jstefaniuk-d2l:
actually, one sec

jstefaniuk-d2l:
LGTM again

ryantmer:
I will leave your moment of doubt as a mystery 😛

PR #196, opened 2016-11-15
Fix some flickering in ie

jstefaniuk-d2l:
Looks like ie bugs out a bit with animation unless the `overflow: hidden` is directly above the element (or something else, but putting this here fixes the problem)


ryantmer:
Dibs.

PR #197, opened 2016-11-15
US79418 - Break course images out into their own component

jstefaniuk-d2l:
- [x] Fix tests

This will allow our nice srcset, fade-in, and different-aspect-ratio parsing behaviour to be used by others

Bob wanted the fade-in for course tile images as well, so I figured that it was a good time to factor out the common behaviour

Success Criteria: Almost exactly the same, course tiles will now fade in like the catalog images do (srcset and aspect ratios should be identical to before)

Later we should probably make a component which you pass in the organizationId/size and it gets the image by itself

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -58,7 +58,7 @@ in that organization - student, teacher, TA, etc.
 								<d2l-icon class$="[[_iconDetails.className]]" icon="[[_iconDetails.iconName]]"></d2l-icon>
 							</div>
 						</div>
-						<d2l-course-image aria-hidden="true" image="[[_image]]" sizes="[[tileSizes]]" size-type="tile"></d2l-course-image>
+						<d2l-course-image aria-hidden="true" image="[[_image]]" sizes="[[tileSizes]]" type="tile"></d2l-course-image>
 					</div>
 				</div>
 				<div class="course-text d2l-heading-4">{{_organization.properties.name}}</div>
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -58,7 +58,7 @@ in that organization - student, teacher, TA, etc.
 								<d2l-icon class$="[[_iconDetails.className]]" icon="[[_iconDetails.iconName]]"></d2l-icon>
 							</div>
 						</div>
-						<d2l-course-image image="[[_image]]" sizes="[[tileSizes]]" size-type="tile"></d2l-course-image>
+						<d2l-course-image aria-hidden="true" image="[[_image]]" sizes="[[tileSizes]]" size-type="tile"></d2l-course-image>
 					</div>
 				</div>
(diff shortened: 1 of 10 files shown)


d2l-course-tile.html
```diff
@@ -57,7 +58,7 @@
 								<d2l-icon class$="[[_iconDetails.className]]" icon="[[_iconDetails.iconName]]"></d2l-icon>
 							</div>
 						</div>
-						<img aria-hidden="true" sizes="[[tileSizes]]"/>
+						<d2l-course-image image="[[_image]]" sizes="[[tileSizes]]" size-type="tile"></d2l-course-image>
```
jstefaniuk-d2l: whoops, putting aria-hidden back in

d2l-simple-overlay.html
```diff
@@ -32,7 +32,7 @@ <h1 class="d2l-heading-2">{{titleName}}</h1>
 			</div>
 			<div class="scrollable">
 				<div class="max-width">
-					<content></content>
+					<content id="content"></content>
```
ryantmer: Question: this needed? If so, I'd kinda recommend against using an ID - while currently there's only the one `d2l-simple-overlay` on a page, that may not always be the case.

ryantmer:
General question: would having this completely separate and in its own BrightspaceUI repo make sense? Maybe not yet, but if we want to use course images elsewhere (like homepage banner), would be nice to have it fully broken out. (I guess YAGNI(yet))

d2l-simple-overlay.html
jstefaniuk-d2l: I don't actually know why it added that in there... Will remove!

jstefaniuk-d2l: Also, we have multiple simple overlays on the page and no issues, we also have other components which have id's which get posted many times, so I think polymer does something shady to avoid the id problem

ryantmer: Yeah I mean, in the case of proper actual Shadow DOM I think that is the case, so maybe? I've avoided it either way, just to be safe. (But as you say, it's entirely possible there's no need to be safe in this case.)

ryantmer:
One minor thing - should the `img` element have an `aria-describedby` or something?

jstefaniuk-d2l:
I don't think it's needed for now, course images aria-hides the image, and I think that when you tab around the image selector there's a label somewhere else that gets read

ryantmer:
:+1: Coolio!

jstefaniuk-d2l:
I'm not sure if we've ever done a good "make sure the image selector is accessible" sweep though, so it's possible that we may not even be reading out image names or anything (maybe just "choose this image")

I can't even think of what we would read out tbh

ryantmer:
Hm, fair point. I'll add it to the existing accessibility GH issue (#125).

PR #198, opened 2016-11-15
Trying to fix some things

### LINKED ISSUE #99: Unpin Firefox (again)

### LINKED ISSUE #171: (Eventually) Remove wct-local dependency
The newest version of `wct-local` doesn't seem to work with the version of java that our travis is using.We now pull the newest working version in as a dependency, so that when `wct` looks for `wct-local`, it finds the package and doesn't pull the most recent.

Once `wct-local` starts working for us, we should remove the dependency

ryantmer:
This removes the override for wct-local (fixes #171), and the override for firefox (fixes #99). To do the latter, had to specify that Java 8 should be used, which increases build times a bit, but I think worth it to get things back to most-recent versions.

This works for me locally, and Travis seems to like it, but it would be super excellent if somebody could check this out and verify that everything does indeed work for them.

(Also updated to node 5, while I was here - shouldn't have any real effect, since it's only used for build/test anyway.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 language: node_js
 node_js:
-- 4
+- 5
 sudo: false
 addons:
-  firefox: 46.0
+  firefox: latest
+  apt:
+    packages:
+      - oracle-java8-installer
+      - oracle-java8-set-default
 before_script:
 - npm run test:lint:js
 script:
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "polylint": "^2.10.1",
     "polymer-cli": "^0.12.0",
     "rimraf": "^2.5.2",
-    "wct-local": "2.0.12",
     "web-component-tester": "^4.2.2"
   }
 }


CodeBaboon:
This LGTM. I rimraf'd my `node_modules` and `bower_components` and then rebuilt everything and re-ran the tests and all seems good. Widget still seems to work. Only caveat is that I'm still on node v4.5 but Travis seems cool with v5 so I don't think that's a problem.

ryantmer:
Coolio. I think that's about as good as testing this requires - pretty sure if anything broke, it would super-break.

PR #199, opened 2016-11-16
Remove iron-component-page

ryantmer:
Yeah this wasn't really serving much purpose, and was making docstrings annoyingly long. Also, the demo was quite broken and is really not worth the effort since the full widget really needs to be in the LMS to work coherently anyway.

Note that this is like 99% just comment changes (some of which I just fully removed since they were obvious). I also removed a few unused properties though, I'll highlight those.


d2l-simple-overlay.html
```diff
@@ -111,14 +86,6 @@ <h1 class="d2l-heading-2">{{titleName}}</h1>
 						};
 					}
 				},
-				/* TODO: No longer required? */
-				_viewportContent: {
```
ryantmer: Removed these, didn't see them being used anywhere.

d2l-touch-menu-item.html
```diff
@@ -26,132 +23,75 @@
...
-				*/
+				// Icon for menu item
 				_displayBackgroundImage: {
 					type: String
 				},
-				/*
-				* TODO: Remove? Doesn't seem to be used anywhere
-				* @type {Object}
-				*/
-				_hostNode: {
```
ryantmer: Removed this as well.

jstefaniuk-d2l:
Code LGTM

PR #200, opened 2016-11-16
make it so changing the body font size doesn't throw our max height off

jstefaniuk-d2l:
To test:

- Open the page
- Change the font size on <body> to `16px` (anything but 20px really)


ryantmer:
Checkin' dis out.

ryantmer:
Things that happened before:

- wonky half-line of text showing up with different font sizes

Things that don't happen anymore:

- wonky half-line of text showing up with different font sizes

Whoop!

PR #201, opened 2016-11-16
Fix blank images issue

jstefaniuk-d2l:
fix issue where image would permanently hide if you swapped out the image object but the new one had the same data

To trigger the old error:

- Checkout master
- Search for "paint"
- Search for "paints"
- Notice how everything is nice and grey

### DIFF
--- a/components/d2l-course-image.html
+++ b/components/d2l-course-image.html
@@ -77,16 +77,16 @@
 				this._updateImageSource();
 			},
 			_updateImageSource: function(newValue, oldValue) {
-				this._imageClass = 'hidden';
 
 				this._srcset = this.getImageSrcset(this.image, this.type);
 				this._tileSizes = this._generateSizes(this.sizes);
 
-				if (this.getDefaultImageLink(newValue, this.type) === this.getDefaultImageLink(oldValue, this.type) &&
-					this._imageClass.indexOf('shown') > -1
-				) {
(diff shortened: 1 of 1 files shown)


ryantmer:
Taking a look!

ryantmer:
Hmm... seeing the same behaviour with and without this change. Request is definitely happening, but still grey boxes :(

jstefaniuk-d2l:
By any chance does it work if you clear your cache first

ryantmer:
Newp, same thing. It's definitely hitting the new code and everything, but still just grey.

jstefaniuk-d2l:
Right you are, one sec, it was working yesterday, perhaps I screwed something up when pushing

components/d2l-course-image.html
```diff
@@ -76,11 +76,17 @@
...
 				this._updateImageSource();
 			},
-			_updateImageSource: function() {
+			_updateImageSource: function(newValue, oldValue) {
 				this._imageClass = 'hidden';
 
 				this._srcset = this.getImageSrcset(this.image, this.type);
 				this._tileSizes = this._generateSizes(this.sizes);
+
+				if (this.getDefaultImageLink(newValue, this.type) === this.getDefaultImageLink(oldValue, this.type) &&
```
ryantmer: Is this condition necessary? Don't we always want to set the `shown` class whenever the source changes?

jstefaniuk-d2l: Think I found the error

ryantmer:
Coolio, all good now! LGTM.

PR #202, opened 2016-11-16
fix up the call to moreSemestersRequest

CodeBaboon:
This was causing some tests to fail for me in the `Filter list content` section of `d2l-all-courses.js`.


ryantmer:
Woah, how was this even working? Or... was it even. Most test instances have fewer semesters than departments, so maybe it never was! 

LGTM, good catch!

PR #203, opened 2016-11-17
Determine if can change image by existence of set-catalog-image action

CodeBaboon:
The `edit-image-page` is actually the page where you can upload an image, but on the course tile we want to determine if the "Change Image" menu item is shown based on the existence of the `set-catalog-image` action, which is what is called when a user selects a new image from the catalog.

This fixes an issue whereby a user who has CHANGE_IMAGE permissions was unable to see the "Change Image" menu item unless they also had SEE_COURSE_INFO permission.

### DIFF
--- a/test/d2l-course-tile/d2l-course-tile.js
+++ b/test/d2l-course-tile/d2l-course-tile.js
@@ -293,6 +293,24 @@ describe('<d2l-course-tile>', function() {
 			widget.getDefaultImageLink = sinon.stub().returns(href);
 		});
 
+		it('should have a change-image-button if the set-catalog-image action exists on the organization', function() {
+			var orgWithSetCatalogImageAction = JSON.parse(JSON.stringify(organization));
+			orgWithSetCatalogImageAction.actions = [{
+				name: 'set-catalog-image',
+				method: 'POST',
+				href: ''
+			}];
+			var parser = document.createElement('d2l-siren-parser');
+
+			var result = widget._getCanChangeCourseImage(parser.parse(orgWithSetCatalogImageAction));
+			expect(!!result).to.equal(true);
+		});
+
+		it('should not have a change-image-button if the set-catalog-image action does not exist on the organization', function() {
+			var result = widget._getCanChangeCourseImage(organizationEntity);
+			expect(!!result).to.equal(false);
+		});
+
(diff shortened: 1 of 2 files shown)


test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -293,6 +293,24 @@ describe('<d2l-course-tile>', function() {
...
+			var orgWithSetCatalogImageAction = JSON.parse(JSON.stringify(organization));
+			orgWithSetCatalogImageAction.actions = [{
+				name: 'set-catalog-image',
+				method: 'POST',
+				href: ''
+			}];
+			var parser = document.createElement('d2l-siren-parser');
+
+			var result = widget._getCanChangeCourseImage(parser.parse(orgWithSetCatalogImageAction));
+			expect(!!result).to.equal(true);
```
jstefaniuk-d2l: Just a note that you can use `ok` for truthy checks

`expect(result).to.be.ok`
http://chaijs.com/api/bdd/

No need to go in and change this though, ok isn't too clear of a keyword anyways

ryantmer: Similarly, can also use `expect(result).to.be.true`.

jstefaniuk-d2l: `expect(result).to.be.ok` === `expect(!!result).to.be.true`

dougmoscrop: similarly, `expect(result).to.be.ok` !== `expect(result).to.be.true`

That's kind of a smell - if the code should return something "truthy" and that's good enough - using `expect().to.be.ok` or `expect().to.be.truthy` says that. saying `expect().to.be.true` says "it should return a boolean". what exists here is a sort of ambiguous middleground where you coerece a value to a boolean and then check that it definitely is a boolean.

ryantmer: Yeah, fair. I was mostly just referring to readability - I didn't actually check what `_getCanChangeCourseImage` returns :)

jstefaniuk-d2l:
I can test this after my build runs

jstefaniuk-d2l:
Tested, works well

PR #207, opened 2016-11-21
US78151 - Improve filter fetching

ryantmer:
Previously, the filter dropdown was being populated by calls to the /organizations HM API - however, this meant all users would see all departments/semesters in those lists, which isn't correct (it was an easier first slice is all).

This changes the fetching of those items to be via the /enrollments domain - meaning you will only receive semesters and departments you are enrolled in. This works, as admins and instructors are generally directly enrolled in semesters and departments, whereas students are not. This is also a much faster API call, even considering the fact that we're now fetching enrollments, then fetching all the organizations - so much faster that we could probably even get away with only fetching them when the user actually clicks the "Filter" menu (but for now, it's just whenever the all courses overlay is opened).

I also reduced the number of API calls we're doing - the `d2l-my-courses` component does the initial /enrollments call, then does the `search-my-enrollments` action. The response from that is then passed to `d2l-all-courses` (this is what is new), so we don't have to re-fetch that.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -270,8 +270,8 @@
 				sort-field="[[_sortField]]">
 			</d2l-search-widget-custom>
 
-			<div id="filterAndSort">
-				<div id="filterSection">
+			<div id="filterAndSort" class="hidden">
+				<div id="filterSection" class="hidden">
 					<span id="filterText" class="dropdown-opener-text" on-tap="_toggleFilterDropdown">{{localize('filtering.filter')}}</span>
 					<d2l-dropdown id="filterDropdown">
 						<button
@@ -521,8 +521,13 @@
 				// Load course tile and filter contents. Called from d2l-my-courses.
 				this.delayCourseTileLoad = false;
 
-				this.$.departmentSearchWidget.search();
-				this.$.semesterSearchWidget.search();
+				// Only make the call to load the filter menu contents if user has a
+				// substantial number of enrollments - otherwise, hide the filter menu
+				if (this.pinnedEnrollments.length + this.unpinnedEnrollments.length >= 20) {
+					this.$.departmentSearchWidget.search();
+					this.$.semesterSearchWidget.search();
+					this.toggleClass('hidden', false, this.$.filterAndSort);
+				}
 			},
 			setCourseImage: function(details) {
 				this._removeAlert('setCourseImageFailure');
@@ -595,9 +600,16 @@
 			_focusFilterText: function() {
 				this.toggleClass('focus', true, this.$.filterText);
 			},
-			_updateMoreParameters: function(entity, moreUrlPath, hasMorePath) {
+			_onSearchResults: function(entity, moreUrlPath, hasMorePath) {
 				this.set(hasMorePath, entity.hasLinkByRel('next'));
 				this.set(moreUrlPath, this[hasMorePath] ? entity.getLinkByRel('next').href : '');
+
+				// If user has >1 departments or >1 semesters, signal to d2l-all-courses to show filters
+				if (this._departments && this._departments.length > 1
+					|| this._semesters && this._semesters.length > 1
+				) {
+					this.toggleClass('hidden', false, this.$.filterSection);
+				}
 			},
 			_onFilterDropdownClose: function() {
 				// If the dropdown is closed by clicking elsewhere, we need to manually update _filterDropdownOpen
@@ -611,12 +623,12 @@
 				this._selectSemesterList();
 			},
 			_onDepartmentSearchResults: function(e) {
-				this._updateMoreParameters(e.detail, '_moreDepartmentsUrl', '_hasMoreDepartments');
 				this.set('_departments', e.detail.entities);
+				this._onSearchResults(e.detail, '_moreDepartmentsUrl', '_hasMoreDepartments');
 			},
 			_onSemesterSearchResults: function(e) {
-				this._updateMoreParameters(e.detail, '_moreSemestersUrl', '_hasMoreSemesters');
 				this.set('_semesters', e.detail.entities);
+				this._onSearchResults(e.detail, '_moreSemestersUrl', '_hasMoreSemesters');
 			},
 			_onMoreResponse: function(response, organizationsPath, moreUrlPath, hasMorePath) {
 				if (response.detail.status === 200 || response.detail.status === 304) {
@@ -626,7 +638,7 @@
 						this.push(organizationsPath, entity);
 					}.bind(this));
 
-					this._updateMoreParameters(responseEntity, moreUrlPath, hasMorePath);
+					this._onSearchResults(responseEntity, moreUrlPath, hasMorePath);
 				}
 			},
 			_onMoreDepartmentsResponse: function(response) {
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -36,6 +36,11 @@
 				margin-right: 0.5rem;
 			}
 		</style>
+		<d2l-ajax
+			url="[[_url]]"
+			headers='{ "Accept": "application/vnd.siren+json" }'
+			on-iron-ajax-response="_onResponse">
+		</d2l-ajax>
 		<d2l-icon icon="d2l-tier2:check-box-unchecked" aria-hidden="true"></d2l-icon>
 		<span>[[text]]</span>
 	</template>
@@ -43,11 +48,30 @@
 		'use strict';
 		Polymer({
 			is: 'd2l-list-item-filter',
+			properties: {
+				entity: {
+					type: Object,
+					observer: '_onEntityChanged'
+				},
+				_url: String
+			},
 			behaviors: [
 				window.D2L.PolymerBehaviors.MenuItemSelectableBehavior
 			],
 			listeners: {
 				'd2l-menu-item-select': '_onSelect'
+			},
+			_onEntityChanged: function(entity) {
+				this.set('_url', entity.links[1].href);
+				this.$$('d2l-ajax').generateRequest();
+			},
+			_onResponse: function(response) {
+				if (response.detail.status === 200 ) {
+					var parser = document.createElement('d2l-siren-parser');
+					var entity = parser.parse(response.detail.xhr.response);
+
+					this.set('text', entity.properties.name);
+				}
 			},
 			_onSelect: function(e) {
 				this.set('selected', !this.selected);
@@ -209,12 +233,6 @@
 		</style>
 
 		<d2l-ajax
-			id="organizationsRequest"
-			url="[[_organizationsUrl]]"
-			on-iron-ajax-response="_onOrganizationsResponse">
-		</d2l-ajax>
-
-		<d2l-ajax
 			id="moreDepartmentsRequest"
 			url="[[_moreDepartmentsUrl]]"
 			headers='{ "Accept": "application/vnd.siren+json" }'
@@ -290,8 +308,7 @@
 									<template is="dom-repeat" items="[[_semesterOrganizations]]">
 										<d2l-list-item-filter
 											value="[[getEntityIdentifier(item)]]"
-											text="[[item.properties.name]]"
-											selected="[[item.properties.selected]]">
+											entity="[[item]]">
 										</d2l-list-item-filter>
 									</template>
 								</d2l-menu>
@@ -307,8 +324,7 @@
 									<template is="dom-repeat" items="[[_departmentOrganizations]]">
 										<d2l-list-item-filter
 											value="[[getEntityIdentifier(item)]]"
-											text="[[item.properties.name]]"
-											selected="[[item.properties.selected]]">
+											entity="[[item]]">
 										</d2l-list-item-filter>
 									</template>
 								</d2l-menu>
@@ -416,38 +432,8 @@
 					value: '/d2l/api/hm/organizations',
 					readOnly: true
 				},
-				_searchSemestersAction: {
-					type: Object,
-					value: {
-						'name':'search-course-collection',
-						'method':'GET',
-						'href': '/d2l/api/hm/organizations',
-						'fields':[
-							{'name':'search', 'type':'search', 'value':''},
-							{'name':'page', 'type':'number', 'value':'1'},
-							{'name':'pageSize', 'type':'number', 'value':'20'},
-							{'name':'embedDepth', 'type':'number', 'value':'1'},
-							// 5 is the default type ID for semesters
-							// organizationTypeIds is updated by organizationsRequest
-							{'name':'organizationTypeIds', 'type':'hidden', 'value':'5'}]
-					}
-				},
-				_searchDepartmentsAction: {
-					type: Object,
-					value: {
-						'name':'search-course-collection',
-						'method':'GET',
-						'href':'/d2l/api/hm/organizations',
-						'fields':[
-							{'name':'search', 'type':'search', 'value':''},
-							{'name':'page', 'type':'number', 'value':'1'},
-							{'name':'pageSize', 'type':'number', 'value':'20'},
-							{'name':'embedDepth', 'type':'number', 'value':'1'},
-							// 203 is the default type ID for departments
-							// organizationTypeIds is updated by organizationsRequest
-							{'name':'organizationTypeIds', 'type':'hidden', 'value':'203'}]
-					}
-				},
+				_searchSemestersAction: Object,
+				_searchDepartmentsAction: Object,
 				_semesterOrganizations: {
 					type: Array,
 					value: function() {
@@ -518,8 +504,12 @@
 				var itemCount = Math.max(this.pinnedEnrollments.length, this.unpinnedEnrollments.length);
 				return itemCount;
 			},
-			loadCourseTiles: function() {
+			load: function() {
+				// Load course tile and filter contents. Called from d2l-my-courses.
 				this.delayCourseTileLoad = false;
+
+				this.$.departmentSearchWidget.search();
+				this.$.semesterSearchWidget.search();
 			},
 			setCourseImage: function(details) {
 				this._removeAlert('setCourseImageFailure');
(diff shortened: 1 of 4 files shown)


ryantmer:
Looking into the test failures - they pass on Chrome, but give weird errors on FF... odd.

d2l-all-courses.html
```diff
@@ -35,23 +35,59 @@
...
+					type: Object,
+					observer: '_onEnrollmentEntityChanged'
+				},
+				selected: {
+					type: Boolean,
+					value: false,
+					observer: '__onSelectedChanged',
+					reflectToAttribute: true
+				},
+				_url: String
```
jstefaniuk-d2l: `_url` is pretty non-descriptive

ryantmer: Yeah, I was trying to make it fairly abstract, but may have over-abstracted :P I shall fix.

d2l-all-courses.html
```diff
@@ -35,23 +35,59 @@
...
 	<script>
 		'use strict';
 		Polymer({
 			is: 'd2l-list-item-filter',
+			properties: {
+				enrollmentEntity: {
+					type: Object,
+					observer: '_onEnrollmentEntityChanged'
+				},
+				selected: {
```
jstefaniuk-d2l: I feel like this could also use some context

ryantmer: This one just directly overrides the selectable menu item's `selected`, so has to be the same name =/

jstefaniuk-d2l: Hmm, perhaps `menu-item-selected` would have been a more descriptive name, ohh well probably too late to change now

d2l-all-courses.html
```diff
@@ -35,23 +35,59 @@
...
 			listeners: {
 				'd2l-menu-item-select': '_onSelect'
 			},
+			_onEnrollmentEntityChanged: function(entity) {
+				this.set('_url', entity.getLinkByRel(/\/organization$/).href);
+				this.$$('d2l-ajax').generateRequest();
+			},
+			_onResponse: function(response) {
+				if (response.detail.status === 200 ) {
+					var parser = document.createElement('d2l-siren-parser');
```
jstefaniuk-d2l: can this use `this._parser`?

d2l-all-courses.html
ryantmer: Eh, yeah - in a normal context, you'd be looking at a `<d2l-menu-item-checkbox selected="true">`, so in that context it's fairly self-explanatory. Admittedly here it's a bit weirder just because the component name doesn't necessarily imply that it's selectable.

d2l-all-courses.html
ryantmer: Not the existing one (different component), but yeah I'll use the same approach here :

d2l-all-courses.html
jstefaniuk-d2l: Makes sense, maybe this is just an odd case

ryantmer: Yeah, pretty much :(

d2l-all-courses.html
```diff
@@ -35,23 +35,59 @@
...
+					var entity = parser.parse(response.detail.xhr.response);
+
+					this.set('text', entity.properties.name);
+					this.set('value', entity.getLinkByRel('self').href);
+				}
+			},
 			_onSelect: function(e) {
 				this.set('selected', !this.selected);
 				this.$$('d2l-icon').setAttribute('icon', this.selected ? 'd2l-tier2:check-box' : 'd2l-tier2:check-box-unchecked');
```
ryantmer: Yeah - `_onSelectedChanged` is just for when `selected` is set directly. Normal operation (tap/click) triggers `_onSelect`.

jstefaniuk-d2l: I was wondering because `_onSelect` sets selected directly on the line before: `this.set('selected', !this.selected);`

d2l-all-courses.html
jstefaniuk-d2l: Oh shoot, there's two components in this file, never noticed that

Is there any reason to have it this way as opposed to putting `d2l-list-item-filter` in it's own component?

ryantmer: Reasoning was mostly because unlike most components, this really _shouldn't_ be broken out and used elsewhere. My rationale was that this is "internal" enough that it won't be tempting to use elsewhere?

d2l-all-courses.html
ryantmer: Oh hah, yeah derp. Good call :P

d2l-all-courses.html
jstefaniuk-d2l: I'm a big fan of encapsulation for code clarity,

Now that we have a 'components' folder, we could throw everything that we intend to actually be reused into there, like `d2l-simple-overlay`, or we could have a folder for things that we never want to be reused, `internal` or something

CodeBaboon: Yep, alert is moving, just need to get word on which org to put it in and get the repo created.

ryantmer:
Okay fixed tests finally, and addressed the feedback. I feel like shit though, so going to be OOO for the rest of the day - I'd say ignore this until I can come back to it with a clear head to make sure it all works as intended...

ryantmer:
Okay, came back to this with a clear head and self-reviewed/tested a bit. Everything seems to be as expected now, so ready for actual review/testing :D

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
...
+			},
 			_onSelect: function(e) {
 				this.set('selected', !this.selected);
-				this.$$('d2l-icon').setAttribute('icon', this.selected ? 'd2l-tier2:check-box' : 'd2l-tier2:check-box-unchecked');
 				this.__onSelect(e);
-			}
+			},
+			__onSelectedChanged: function(e) {
+				// The selected state can be changed by click/tap or by _checkSelected below - need to update the icon when this happens
+				this.$$('d2l-icon').setAttribute('icon', this.selected ? 'd2l-tier2:check-box' : 'd2l-tier2:check-box-unchecked');
```
trevorjonesd2l: Should this selector be more specific than `d2l-icon` in case another icon is ever added?

d2l-all-courses.html
```diff
@@ -494,6 +489,8 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				this.$.scrollThreshold.scrollTarget = this.$.filterDropdownContent.querySelector('.d2l-dropdown-content-container');
 
 				window.addEventListener('resize', this._onResize.bind(this));
+
+				this._parser = document.createElement('d2l-siren-parser');
```
trevorjonesd2l: Do we need to create the parser both here and on `_onReponse`?

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
 				margin-right: 0.5rem;
 			}
 		</style>
+		<d2l-ajax
+			url="[[_organizationUrl]]"
+			headers='{ "Accept": "application/vnd.siren+json" }'
+			on-iron-ajax-response="_onResponse">
```
CodeBaboon: `_onResponse` feels sort of vague. Maybe `_onOrganizationResponse`?

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
...
+			},
 			behaviors: [
 				window.D2L.PolymerBehaviors.MenuItemSelectableBehavior
 			],
 			listeners: {
 				'd2l-menu-item-select': '_onSelect'
 			},
+			_onEnrollmentEntityChanged: function(entity) {
+				this.set('_organizationUrl', entity.getLinkByRel(/\/organization$/).href);
+				this.$$('d2l-ajax').generateRequest();
```
CodeBaboon: could the `d2l-ajax` be `auto` if it always fetches whenever the `_organizationUrl` is set?

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
...
+			},
+			_onResponse: function(response) {
+				if (response.detail.status === 200 ) {
```
CodeBaboon: Any chance `response` or `detail` are null or undefined? Or is that pretty much guaranteed?

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
...
 		<d2l-icon icon="d2l-tier2:check-box-unchecked" aria-hidden="true"></d2l-icon>
 		<span>[[text]]</span>
 	</template>
 	<script>
 		'use strict';
 		Polymer({
 			is: 'd2l-list-item-filter',
+			properties: {
+				enrollmentEntity: {
+					type: Object,
```
CodeBaboon: According to the [Polymer documentation](https://www.polymer-project.org/1.0/docs/devguide/properties) default values are shared across all instances of a component, and so for objects and arrays it is suggested to use a function to initialize the value:

"When initializing a property to an object or array value, use a function to ensure that each element gets its own copy of the value, rather than having an object or array shared across all instances of the element."

```
data: {
      type: Object,
      notify: true,
    }
```

I feel like this might be important for this component because it will have multiple instances created in the `d2l-all-courses` component.

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -157,66 +163,62 @@ describe('smoke test', function() {
...
-			sinon.assert.calledWith(updateMoreParametersSpy, sinon.match.object, '_moreDepartmentsUrl', '_hasMoreDepartments');
-			sinon.assert.calledWith(onMoreResponseSpy, sinon.match.object, '_departmentOrganizations', '_moreDepartmentsUrl', '_hasMoreDepartments');
+			setTimeout(function() {
+				sinon.assert.calledWith(updateMoreParametersStub, sinon.match.object, '_moreDepartmentsUrl', '_hasMoreDepartments');
+				sinon.assert.calledWith(onMoreResponseSpy, sinon.match.object, '_departments', '_moreDepartmentsUrl', '_hasMoreDepartments');
 
-			sandbox.restore();
+				sandbox.restore();
+				done();
+			}, 100);
```
CodeBaboon: Does it need to be 100 or can it be something lower (as low as even 1)? Not that 100ms is the end of the world.

d2l-all-courses.html
CodeBaboon: Thinking about this with a fresh mind, not an issue unless you are mutating the object rather than replacing it, I would gather. So probably safe to ignore here, though maybe not a terrible habit to form to use a function going forward.

ryantmer: Hm, I think you're right in both comments. I don't recall ever seeing issues because of this, but you're right, better safe than sorry/follow good habits. I shall fix!

test/d2l-all-courses/d2l-all-courses.js
jstefaniuk-d2l: We could always use sinon.clock here if needed as well

d2l-all-courses.html
CodeBaboon: I would suspect that empty Arrays are the thing more likely to result in bugs, because you might default them to the same array and then just start `push`-ing stuff in there, and I guess they would show up in all the instances.

ryantmer: Ah yeah, good call. I didn't actually know the whole "one shared" thing, great to know!

d2l-all-courses.html
ryantmer: The only case I can think of where this would create problems is if there was another icon - but since we control the content (and it's just the organization name), I don't see that really being an issue. Inclined to "meh"?

d2l-all-courses.html
```diff
@@ -35,24 +35,63 @@
...
 				'd2l-menu-item-select': '_onSelect'
 			},
+			_onEnrollmentEntityChanged: function(entity) {
+				this.set('_organizationUrl', entity.getLinkByRel(/\/organization$/).href);
+				this.$$('d2l-ajax').generateRequest();
+			},
+			_onResponse: function(response) {
+				if (response.detail.status === 200 ) {
+					if (!this._parser) {
+						this._parser = document.createElement('d2l-siren-parser');
```
jstefaniuk-d2l: Idea: (not for this PR)
Put `_parser` into a behavior since it never actually has to be different

test/d2l-all-courses/d2l-all-courses.js
ryantmer: Whoops, yep. Not even sure this is required anymore? These tests required some fiddling that I did while half-brain-jimmied, so I'll double-check.

ryantmer: Hah yep, not only is this not necessary, but I also failed to update the response in the following (semester) test. Thanks!

ryantmer:
Just found another optimization I think I can do in the "don't call things if we don't have to" camp, but I'll leave it to a follow-up PR as it's not strictly related to this change.

ryantmer:
(By which I meant, I have now re-added the optimization that I removed during the rework, but there's another one that could possibly be done as well... is what I meant by that.)

jstefaniuk-d2l:
Code LGTM

jstefaniuk-d2l:
`bower ECONFLICT     Unable to find suitable version for d2l-offscreen`

ryantmer:
Yeah - [this](https://github.com/Brightspace/d2l-dropdown-ui/pull/85) should remedy that, I'll re-run Travis once that's merged.

jstefaniuk-d2l:
@ryantmer 
What are some good test cases for this? I may test this later today if no-one else picks it up

ryantmer:
@jstefaniuk-d2l - the changes are all centred around the filter list content. Previously, it was fetching all the things from /organizations, which meant all users saw all semesters and departments in the list, even those that weren't relevant. Now, it hits /enrollments first (fetching just directly-enrolled-semesters, and directly-enrolled-departments), then fetches the information about the enrollments from /organizations.

This is more in line with what the widget does for the course tiles (except there its directly-enrolled-course-offerings instead), is much faster (despite there being more calls, they are hella fast in comparison), and means we only present relevant semesters now.

I can give you a specific rundown of the before/after calls if you'd like as well!

jstefaniuk-d2l:
Will test first thing tomorrow

jstefaniuk-d2l:
I'm seeing weirdness when clicking on filters, the boxes aren't checking off

Not a problem with this PR, this seems to be a thing on master as well, anyone else seeing this? (If it's just me then maybe I'll wipe my bsi's node/bower folder)

jstefaniuk-d2l:
I can confirm that the network requests are going to enrollments, and then organizations, going to do a couple more quick checks before giving the almighty thumbs up

Side note:
Has there been any talk about batching HM calls somehow? That's a lot of calls to organizations (Or maybe we could graft the organizations on to the enrollments call with a querystring?)

ryantmer:
Regarding the weirdness, is that when checking items manually? If so then yeah, I suspect something is an old version. There is a problem with the clear button (it clears the internal state, but doesn't update the checkboxes in the UI), but I've fixed that in the follow-up PR to this one.

Regarding the side note, I've definitely thought it, but I don't know of a good way to do it. The batching is very much possible under HTTP2 (without needing to change how we do things, much), but we're not there yet obviously. There is the `embedDepth` parameter, but it doesn't apply to links (the organizations are links on an enrollment, not an embedded entity). So, yeah - a good idea for sure, but thus far no obvious immediate solution, and no serious investigation into it 😢

jstefaniuk-d2l:
Could we _make_ organizations an embedded entity? (just if the querystring is provided)

ryantmer:
We can, but I think the issue is that you're embedding things from a different domain - not that it isn't possible, more that it's architecturally strange, since it fuzzes lines between separation of concerns.

This is definitely a good conversation to have, though - would want to get others involved though, so I'll move it to Flowdock.

d2l-all-courses.html
```diff
@@ -519,8 +517,18 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
 				return itemCount;
 			},
-			loadCourseTiles: function() {
+			load: function() {
+				// Load course tile and filter contents. Called from d2l-my-courses.
 				this.delayCourseTileLoad = false;
+
+				// Only make the call to load the filter menu contents if user has a
+				// substantial number of enrollments - otherwise, hide the filter menu
+				if (this.pinnedEnrollments.length + this.unpinnedEnrollments.length > 20) {
```
jstefaniuk-d2l: `>=`

jstefaniuk-d2l: We need to also hide filterAndSort if there is less than 2 departments or 2 semesters

jstefaniuk-d2l: There's also this: `A user enrolled with (semesters + departments) <= 7 will not see the Department/Semester tabs and search bars in the filter dropdown, and will instead just see a flat list of departments and semesters combine`

Although that seems like it could be split into a separate PR (I haven't looked all that much into your other PR, so not sure if it's there)

ryantmer: Oh, yeah that kinda got split out - there's a design that just shows a flat list (non-tabbed), but that was going to add even more complexity (hence the splitting out the component in the other PR). It'll actually be another follow-up PR probably :P

ryantmer: I'm assuming you mean "2 departments and 2 semesters", right? (otherwise we'd hide the menu for a user with 1 department and 345 semesters) But yarr, will fix.

jstefaniuk-d2l: Oddly enough, according to the story, that case would be hidden as well: 
```
A user will not see the Filter dropdown at all if:
They have <= 20 enrollments
They are enrolled in only 1 semester
They are enrolled in only 1 department
They are enrolled in only 1 department and 1 semester
```

ryantmer: Oh, sorry - that's me not being clear enough. That should actually be:

```
A user will not see the Filter dropdown at all if:
They have <= 20 enrollments
They are enrolled in only 1 semester and 0 departments
They are enrolled in only 1 department and 1 semester
They are enrolled in only 1 department and 1 semester
```

jstefaniuk-d2l:
Testing LGTM

PR #208, opened 2016-11-21
Add in error message langterms now so that they are included in translation

jstefaniuk-d2l:


ryantmer:
Seems good - where are these langterms being used?

jstefaniuk-d2l:
One will be used when the course image catalog is down, the other will be when you try to set an image but it doesn't work

jstefaniuk-d2l:
I just want to include them because I'm sending in the file for translation today and figured that we could save some time this way

ryantmer:
Ohh, gotcha gotcha. So they're not actually in use yet, I was confused. Yep, LGTM!

PR #209, opened 2016-11-22
Fix grey box that sometimes shows when setting an image

jstefaniuk-d2l:
fix race condition where if the server is slow responding it can cause the tile to show a null image instead of the right thing


jstefaniuk-d2l:
This was caused because the image selector wipes out the images when closed, when this happened before the server responds, it was causing issues

CodeBaboon:
LGTM, code and test (Chrome/Safari for Mac)

PR #210, opened 2016-11-22
remove duplicate line

jstefaniuk-d2l:
This was so totally worth it's own PR


ryantmer:
best PR ever. adding test and code review labels.

ryantmer:
nah jk lgtm

trevorjonesd2l:
Should we add any unit tests?

ryantmer:
Good point. Let's get Sauce up and running for this.

jstefaniuk-d2l:
I could test this by using an es6 proxy object to make sure that 'set' is only called once on the property when clear is called

😉

PR #211, opened 2016-11-22
DE23354 - Add padding top

jstefaniuk-d2l:
For some reason the widget is super fidgety with margin (or padding as well if at a lower level)

https://rally1.rallydev.com/#/51191528503d/detail/defect/77048466184?fdp=true


CodeBaboon:
LGTM

PR #213, opened 2016-11-22
Working cut of alert on image set failure

CodeBaboon:
This will display a warning alert when the change image action fails. Also changes the color of the failure icon to the warning yellow (as per Bob). Allows for multiple alerts to be displayed at the same time (such as if you get an image failure on the all courses page while no images are pinned, both the no pinned images call-to-action and the change image failure warning are displayed simultaneously).

TODO:

- [x] Rebase
- [x] Tests
- [x] Use the localized text for the set image failure message  
- [x] Change `alert-type` to `type`
- [x] Add duplicate logic into a behavior

### DIFF
--- a/d2l-alert-behavior.html
+++ b/d2l-alert-behavior.html
@@ -1,0 +1,52 @@
+<link rel="import" href="../polymer/polymer.html">
+
+<script>
+	'use strict';
+
+	(function() {
+		window.D2L = window.D2L || {};
+		window.D2L.MyCourses = window.D2L.MyCourses || {};
+
+		/*
+		* Behavior for my courses alert messages
+		*
+		* This contains utility functions for adding and removing alerts
+		*
+		* @polymerBehavior window.D2L.MyCourses.AlertBehavior
+s		*/
+		window.D2L.MyCourses.AlertBehavior = {
+			properties: {
+				// Array containing alert objects for display
+				_alerts: {
+					type: Array,
+					value: function() { return []; }
+				}
+			},
+			_addAlert: function(type, name, message) {
+				if (this._alerts) {
+					this.push('_alerts', {
+						alertType: type,
+						alertName: name,
+						alertMessage: message
+					});
+				}
+			},
+			_hasAlert: function(name) {
+				if (!this._alerts || this._alerts.length <= 0) {
+					return false;
+				}
+				return this._alerts.some(function(alert) {
+					return alert.alertName === name;
+				});
+			},
+			_removeAlert: function(name) {
+				if (this._alerts && this._alerts.length > 0 && this._hasAlert(name)) {
+					this.splice('_alerts', this._alerts.map(function(alert) { return alert.name; }).indexOf(name), 1);
+				}
+			},
+			_clearAlerts: function() {
+				this._alerts = [];
+			}
+		};
+	})();
+</script>
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -22,12 +22,26 @@
 				left: 0;
 				bottom: 0;
 				width: 6px;
-				background-color: #1598cb;
 				border-top-left-radius: 3px;
 				border-bottom-left-radius: 3px;
 				border-top-right-radius: 0;
 				border-bottom-right-radius: 0;
 				margin: -1px;
+			}
+			:host[alert-type="error"] .message-highlight {
+				background-color: #f0533f;
+			}
+			:host[alert-type="warning"] .message-highlight {
+				background-color: #ffce51;
+			}
+			:host[alert-type="call-to-action"] .message-highlight {
+				background-color: #1598cb;
+			}
+			:host[alert-type="reinforcement"] .message-highlight {
+				background-color: #7c8695;
+			}
+			:host[alert-type="confirmation"] .message-highlight {
+				background-color: #00ae81;
 			}
 			:host-context([dir='rtl']) .message-highlight {
 				left: auto;
@@ -38,8 +52,7 @@
 				border-bottom-right-radius: 3px;
 			}
 		</style>
-
-		<div class="message-wrapper" hidden$="{{!visible}}">
+		<div class="message-wrapper">
 			<div class="message-highlight"></div>
 			<content></content>
 		</div>
@@ -51,9 +64,10 @@
 		Polymer({
 			is: 'd2l-alert',
 			properties: {
-				visible: {
-					type: Boolean,
-					value: false
+				'alert-type': {
+					type: String,
+					value: 'call-to-action',
+					reflectToAttribute: true
 				}
 			}
 		});
--- a/d2l-alert.html
+++ b/d2l-alert.html
@@ -28,19 +28,19 @@
 				border-bottom-right-radius: 0;
 				margin: -1px;
 			}
-			:host[alert-type="error"] .message-highlight {
+			:host[type="error"] .message-highlight {
 				background-color: #f0533f;
 			}
-			:host[alert-type="warning"] .message-highlight {
+			:host[type="warning"] .message-highlight {
 				background-color: #ffce51;
 			}
-			:host[alert-type="call-to-action"] .message-highlight {
+			:host[type="call-to-action"] .message-highlight {
 				background-color: #1598cb;
 			}
-			:host[alert-type="reinforcement"] .message-highlight {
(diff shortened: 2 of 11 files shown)


d2l-alert.html
```diff
@@ -22,13 +22,27 @@
...
 				width: 6px;
-				background-color: #1598cb;
 				border-top-left-radius: 3px;
 				border-bottom-left-radius: 3px;
 				border-top-right-radius: 0;
 				border-bottom-right-radius: 0;
 				margin: -1px;
 			}
+			:host[alert-type="error"] .message-highlight {
+				background-color: #f0533f;
```
jstefaniuk-d2l: if these are valence colors, then `var(--d2l-color-celestine);`

CodeBaboon: Nope, the alert colours are unique from the standard colors (with the exception of 'reinforcement' which maps back to Galena). I tried to add them to the colors repo in https://github.com/BrightspaceUI/colors/pull/22/ but it was determined they should not be part of the common set. When d2l-alert gets moved to its own repo I'll close that PR, pull these colors into variables and then maybe even reference the alert warning color from d2l-course-tile-styles.

d2l-all-courses.html
```diff
@@ -727,6 +745,38 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
 				this.$.sortDropdown.toggleOpen();
+			},
+			_updateEnrollmentAlerts: function(hasPinnedEnrollments) {
+				this._alerts = [];
+
+				if (!hasPinnedEnrollments) {
+					this._addAlert('call-to-action', 'noPinnedCourses', this.localize('noPinnedCoursesMessage'));
+				}
+			},
+			_addAlert: function(type, name, message) {
```
ryantmer: Thought: Trying to think of a way to make these a bit nicer... but I think since it's using `dom-repeat` now, it has to be an array :(

jstefaniuk-d2l: could _addAlert, _hasAlert, _removeAlert, and this._alerts be refactored into a behaviour? (since you repeat these functions in multiple elements)

d2l-alert-behavior.html
```diff
@@ -0,0 +1,52 @@
...
+		*
+		* This contains utility functions for adding and removing alerts
+		*
+		* @polymerBehavior window.D2L.MyCourses.AlertBehavior
+s		*/
+		window.D2L.MyCourses.AlertBehavior = {
+			properties: {
+				// Array containing alert objects for display
+				_alerts: {
+					type: Array,
```
ryantmer: Thought/opinion: Hmm... I think this would work arguably the same way as an `Object` instead, using the alert name as a key.

Right now, you can add multiple alerts with the same `name`, but the behaviour of `_hasAlert` and `_removeAlert` is a bit wonky/undefined-ish if you do so:

```
_addAlert('warning', 'foo', 'messages yay');
_addAlert('warning', 'foo', 'other messages yay');
/*
  _alerts is now [{
    alertType: 'warning',
    alertName: 'foo',
    alertMessage: 'messages yay'
  }, {
    alertType: 'warning',
    alertName: 'foo',
    alertMessage: 'other messages yay'
  }]
*/

_hasAlert('foo'); // true
_removeAlert('foo'); // removes the first one
_hasAlert('foo'); // still true
_removeAlert('foo'); // removes the second one
_hasAlert('foo'); // now false
```

I'm assuming you wouldn't usually add more than one alert with the same name, in which case I think an object would make more sense in some respects:

```
_addAlert('warning', 'foo', 'messages yay');
_addAlert('warning', 'foo', 'other messages yay');
/*
_alerts is now {
  'foo': {
    alertType: 'warning',
    alertMessage: 'other messages yay'
}
*/

_hasAlert('foo'); // true
_removeAlert('foo'); // removes alert
_hasAlert('foo'); // now false
```

I dunno... the latter seems to make more sense to me, but it's also EOD Friday, so maybe not. This does go against the inherent "this is a list of alerts" part of it, granted (instead now requiring `Object.keys(_alerts)` or something). So I'm uncertain.

d2l-my-courses.html
```diff
@@ -198,16 +200,22 @@
...
-				window.D2L.MyCourses.CourseManagementBehavior
+				window.D2L.MyCourses.CourseManagementBehavior,
+				window.D2L.MyCourses.AlertBehavior
 			],
 			listeners: {
 				'open-change-image-view': '_openChangeImageView',
 				'set-course-image': '_setCourseImageEvent',
-				'clear-image-scroll-threshold': '_clearImageScrollThreshold'
+				'clear-image-scroll-threshold': '_clearImageScrollThreshold',
+				'simple-overlay-closed': '_onSimpleOverlayClosed'
```
ryantmer: Opinion: It looks like `iron-overlay-behavior` will already generate an event when the overlay is closed - [`opened-changed`](https://github.com/PolymerElements/iron-overlay-behavior/blob/master/iron-overlay-behavior.html#L75), which you could `listen` to in `ready`. It's perhaps not quite as nice, granted.

ryantmer:
Tested this quickly on Windows/Chrome, I'll check other browsers later.

d2l-alert-behavior.html
CodeBaboon: Jeff and I had this discussion yesterday as well. I originally thought we would want to have multiple alerts with the same name potentially, which was why I used an array, but now I'm less certain. Ultimately something would probably need to distinguish them I would imagine, so my original thinking was probably flawed and an object would be better for that scenario for sure.

The downside to an object is, can you `dom-repeat` on it? So the change would require another way of dealing with the situation of showing multiple alerts at a time (of which there is actually a scenario on `d2l-all-courses` where this can occur).

Now, to add a little more complexity to all this thinking, I know that Bob invented a design for handling multiple alerts, some of the same type (in the sense of 2 warnings, not the same name like what I have), some of different types, etc. I think only one displayed at a time and they sorted by priority in some way and you could navigate through them with actions... it was all super cool but was outside the scope of this PR. Bob was cool with the current behaviour of just stacking them in the one situation where it can actually happen right now.

Sooo... here's the vision I sort of see for the alerts:

1) We need to pull `d2l-alert` into its own repo, which is simple enough once we know which org it should go in (Brightspace or BrightspaceUI... UI seems to make more sense but the React alert repos are over in Brightspace so ? ).
2) We should take the color definitions out of `d2l-alert.html`. I created a PR in `d2l-colors` for them but there was concern that if they are only to be used in alert scenarios that they shouldn't be part of the 'base' color package. Speaking with Bob after work today it sounds like he isn't 100% certain what the intent for those colors is, so he is going to discuss it (hopefully early next week) with the other designers and try to get consensus on what those colors will ultimately mean. If they are truly only for alerts then I imagine we should define them as variables and move them into a `d2l-alert-colors.html` in the same repo as `d2l-alert.html`. Then if someone needs to show an alert color independently from our alert control they can do so. If the decision is made that these colors are part of the 'base' colors package we can just merge in the outstanding PR in d2l-colors and reference it from d2l-alert.
3) Moving into the more long-term range, we should consider if the behavioural aspects of the alerts as implemented in this repo should be split out as something re-usable, either in the same package as d2l-alert.html or even something different, like a d2l-alert-manager.html or something. And when we do so, then we should get serious about implementing the behavior as it was designed, rather than just this simple stacking behavior.

ryantmer: Yeah, the `dom-repeat` does make it less-nice... I guess you could set it up on a `calculated` value that uses `Object.keys(theThing)` or something? That might be super ugly granted. I think either way, a change like that can probably wait until this is in its own repo - it's not like we're working with million-element arrays here anyway, so despite the operations being a bit expensive "compared to what they could be", I can't imagine it'll have any serious performance impact 😛 

And, since this has a lovely add/delete/has API, we can change the internals whenever we want and it wouldn't be a breaking change necessarily 😄

d2l-my-courses.html
ryantmer: Yeah, sounds good. I think using `opened-changed` is more Polymer-pragmatic, but I've found that the auto-named-events-created-by-`notify` is also fairly non-obvious as a thing 😕

ryantmer:
Tested! Checked Chrome/IE/FF on Windows 8, Edge W10, and Safari 9 on OSX. LGTM :+1:

PR #214, opened 2016-11-23
fix the squished user-uploaded images

jstefaniuk-d2l:
I noticed that on daylightDemo user-uploaded images were in a rough shape, so here's a fix for that

Upload an image and verify that what's shown isn't distorted and makes sense, ideally someone with safari could test


CodeBaboon:
I be testin'

CodeBaboon:
LGTM

PR #215, opened 2016-11-23
sVal

jstefaniuk-d2l:
This is what I was doodling down on notepads during infusion, thought I made it so I may as well put it out there

Think that it could be useful? (If so I can go and add it into more places, if not I can silently kill this PR)

### DIFF
--- a/d2l-sval-behavior.html
+++ b/d2l-sval-behavior.html
@@ -1,0 +1,48 @@
+<link rel="import" href="../polymer/polymer.html">
+
+<script>
+	'use strict';
+
+	window.D2L = window.D2L || {};
+	window.D2L.MyCourses = window.D2L.MyCourses || {};
+
+	/*
+	* A utility for accessing the inner properties of an object without worrying about whether something is undefined or not
+	* @polymerBehavior window.D2L.MyCourses.sVal
+	*/
+	window.D2L.MyCourses.sVal = {
+		sVal: function(obj, path) {
+			function _retrieveArray(objects, start) {
+				var cur = start;
+				objects.forEach(function(i) {
+					cur = cur.sVal(i);
+				});
+				return cur;
+			}
+
+			var newSVal = {
+				sVal: function(pName) {
+					var nextVal = (this.value || {})[pName];
+					return window.D2L.MyCourses.sVal.sVal(nextVal)
+				},
+				sGet: function(toRetrieve) {
+					if (Array.isArray(toRetrieve)) {
+						return _retrieveArray(toRetrieve, this);
+					} else if (typeof toRetrieve === 'string') {
+						var objects = toRetrieve.split('.');
+						return _retrieveArray(objects, this);
+					} else {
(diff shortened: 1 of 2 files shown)


image-selector/d2l-basic-image-selector.html
```diff
@@ -385,12 +389,8 @@
 			},
 			_doTelemetrySearchRequest: function(userSearchText) {
 				var searchWidget = this.$$('d2l-search-widget'),
-					total = searchWidget.searchResults
-							&& searchWidget.searchResults.properties
-							&& searchWidget.searchResults.properties.pagingInfo
-							&& searchWidget.searchResults.properties.pagingInfo.total
-							? searchWidget.searchResults.properties.pagingInfo.total
-							: 0;
+					total = this.sVal(searchWidget, 'searchResults.properties.pagingInfo.total').value || 0;
```
CodeBaboon: this is hella nice in comparison

d2l-sval-behavior.html
```diff
@@ -0,0 +1,48 @@
...
+			function _retrieveArray(objects, start) {
+				var cur = start;
+				objects.forEach(function(i) {
+					cur = cur.sVal(i);
+				});
+				return cur;
+			}
+
+			var newSVal = {
+				sVal: function(pName) {
```
CodeBaboon: Same here, `pName` might be stronger as `propertyPath` or something.

image-selector/d2l-basic-image-selector.html
ryantmer: This is hella nice for sure, but I think the same functionality can be achieved by just using `reduce`, e.g.

```
var searchWidget = { searchResults: { properties: { pagingInfo: { total: 12 } } } };

total = 'searchResults.properties.pagingInfo.total'.split('.').reduce(function(a, b) {
    if (!a) return;
    return a[b];
}, searchWidget) || 0;
// total is 12

total = 'searchResults.property.pagingInfo.total'.split('.').reduce(function(a, b) {
    if (!a) return;
    return a[b];
}, searchWidget) || 0;
// total is 0
```

(not that this wouldn't be nice to have in a behaviour, granted, since it took me like 10 minutes to get all that right 😜 )

jstefaniuk-d2l: I never actually knew about reduce, nifty! Is there any way to make it return null/undefined rather than 0 if those values turn up?

jstefaniuk-d2l: Whoops, forget that XD, it's too early in the morning

ryantmer: Oh, it's only returning zero because of the `|| 0` (as that's there in the original use case here). It returns undefined normally :)

jstefaniuk-d2l: After some testing, I think I'm sold, when I get a chance I'll make this thing do it the reduce way

PR #216, opened 2016-11-25
Break out filter content

ryantmer:
As I've been adding stuff to filtering, I keep thinking "I should make that it's own component". Well, now I have. I wish I had done it sooner. I am sorry 😭 

This is _mostly_ just removing the `d2l-filter-list-item` out, but there are some other changes too, done due to (really small) problems found in redoing the `d2l-all-courses` tests.

To do:

- [x] Merge #207 
- [x] Change base for this PR
- [x] Write tests for new broken-out component
- [x] Get milk

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -1,103 +1,19 @@
 <link rel="import" href="../polymer/polymer.html">
-<link rel="import" href="../iron-scroll-threshold/iron-scroll-threshold.html">
-<link rel="import" href="../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="../d2l-colors/d2l-colors.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown-content.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown-menu.html">
 <link rel="import" href="../d2l-icons/d2l-icons.html">
-<link rel="import" href="../d2l-menu/d2l-menu-item-selectable-behavior.html">
 <link rel="import" href="../d2l-menu/d2l-menu-item-radio.html">
-<link rel="import" href="../d2l-menu/d2l-menu-item-selectable-styles.html">
-<link rel="import" href="../d2l-search-widget/d2l-search-widget.html">
+<link rel="import" href="../d2l-siren-parser/d2l-siren-parser.html">
 <link rel="import" href="d2l-alert.html">
 <link rel="import" href="d2l-alert-behavior.html">
 <link rel="import" href="d2l-course-management-behavior.html">
 <link rel="import" href="d2l-course-tile-grid.html">
 <link rel="import" href="d2l-course-tile-responsive-grid-behavior.html">
+<link rel="import" href="d2l-filter-menu-content.html">
 <link rel="import" href="d2l-search-widget-custom.html">
 <link rel="import" href="localize-behavior.html">
-
-<dom-module id="d2l-list-item-filter">
-	<template>
-		<style include="d2l-menu-item-selectable-styles">
-			:host {
-				padding: 0.5rem 1rem;
-			}
-
-			:host > span {
-				font-size: 0.8rem;
-			}
-
-			:host d2l-icon {
-				--d2l-icon-height: 1rem;
-				--d2l-icon-width: 1rem;
-				visibility: visible;
-				margin-right: 0.5rem;
-			}
-		</style>
-		<d2l-ajax
-			auto
-			url="[[_organizationUrl]]"
-			headers='{ "Accept": "application/vnd.siren+json" }'
-			on-iron-ajax-response="_onOrganizationResponse">
-		</d2l-ajax>
-		<d2l-icon icon="d2l-tier2:check-box-unchecked" aria-hidden="true"></d2l-icon>
-		<span>[[text]]</span>
-	</template>
-	<script>
-		'use strict';
-		Polymer({
-			is: 'd2l-list-item-filter',
-			properties: {
-				enrollmentEntity: {
-					type: Object,
-					value: function() { return {}; },
-					observer: '_onEnrollmentEntityChanged'
-				},
-				selected: {
-					type: Boolean,
-					observer: '_onSelectedChanged',
-					reflectToAttribute: true
-				},
-				_organizationUrl: String
-			},
-			behaviors: [
-				window.D2L.PolymerBehaviors.MenuItemSelectableBehavior
-			],
-			listeners: {
-				'd2l-menu-item-select': '_onSelect'
-			},
-			_onEnrollmentEntityChanged: function(entity) {
-				if (entity.getLinkByRel && entity.getLinkByRel(/\/organization$/)) {
-					this.set('_organizationUrl', entity.getLinkByRel(/\/organization$/).href);
-				}
-			},
-			_onOrganizationResponse: function(response) {
-				if (response.detail.status === 200 ) {
-					if (!this._parser) {
-						this._parser = document.createElement('d2l-siren-parser');
-					}
-
-					var entity = this._parser.parse(response.detail.xhr.response);
-
-					this.set('text', entity.properties.name);
-					this.set('value', entity.getLinkByRel('self').href);
-				}
-			},
-			_onSelect: function(e) {
-				this.set('selected', !this.selected);
-				this.__onSelect(e);
-			},
-			_onSelectedChanged: function(e) {
-				// The selected state can be changed by click/tap or by _checkSelected below - need to update the icon when this happens
-				this.$$('d2l-icon').setAttribute('icon', this.selected ? 'd2l-tier2:check-box' : 'd2l-tier2:check-box-unchecked');
-				this.__onSelectedChanged(e);
-			},
-			_parser: null
-		});
-	</script>
-</dom-module>
 
 <dom-module id="d2l-all-courses">
 	<template>
@@ -122,9 +38,6 @@
 			}
 			.hidden {
 				display: none !important;
-			}
-			.invisible {
-				visibility: hidden;
 			}
 			.search-and-filter {
 				display: flex;
@@ -165,6 +78,14 @@
 				padding: 0;
 				color: var(--d2l-color-tungsten);
 			}
+			.dropdown-content-header {
+				display: flex;
+				align-items: center;
+				justify-content: space-between;
+				border-bottom: 1px solid var(--d2l-color-gypsum);
+				width: 100%;
+				padding: 20px;
+			}
 			@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
 				#filterText {
 					padding-top: 4px;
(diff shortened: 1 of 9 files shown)


d2l-all-courses.html
```diff
@@ -238,39 +106,23 @@
...
-			id="moreSemestersRequest"
-			url="[[_moreSemestersUrl]]"
-			headers='{ "Accept": "application/vnd.siren+json" }'
-			on-iron-ajax-response="_onMoreSemestersResponse">
-		</d2l-ajax>
-
 		<div class="search-and-filter">
 			<d2l-search-widget-custom
-				hidden$="[[_hasEnrollments]]"
+				hidden$="[[!_hasEnrollments]]"
```
ryantmer: sneaky af

d2l-all-courses.html
```diff
@@ -375,11 +167,9 @@
 
 		<h2 class="d2l-heading-3">{{localize('pinnedCourses')}}</h2>
 
-		<template is="dom-if" if="[[!_hasPinnedEnrollments]]">
```
ryantmer: Since `d2l-alert` has its own visibility, figured we didn't also need the `dom-if` here... I think?

d2l-all-courses.html
```diff
@@ -525,69 +272,26 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-					this.set('_searchSemestersAction', myEnrollmentsEntity.getActionByName('add-semester-filter'));
-				}
-			},
 			_focusFilterText: function() {
 				this.toggleClass('focus', true, this.$.filterText);
 			},
-			_updateMoreParameters: function(entity, moreUrlPath, hasMorePath) {
-				this.set(hasMorePath, entity.hasLinkByRel('next'));
-				this.set(moreUrlPath, this[hasMorePath] ? entity.getLinkByRel('next').href : '');
+			_onFilterChanged: function(e) {
```
ryantmer: This is the crux of the communication between the filter menu content and the search bar (via `_parentOrganizations`), and the filter menu content and the text that displays on the filter menu opener (via `_filterText`).

d2l-all-courses.html
```diff
@@ -597,104 +301,28 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-				button.setAttribute('aria-pressed', selected);
-			},
 			_toggleFilterDropdown: function() {
 				// This is somewhat gross, but is a side effect of using the filter text span as a secondary opener
 				if (this._filterDropdownOpen) {
 					this.$.filterDropdownContent.close();
 					this.$.filterDropdownOpener.focus();
 				} else {
 					this.$.filterDropdownContent.open();
+					this.$.filterMenuContent.load();
```
ryantmer: While it may not seem like it, this is okay to do here. This triggers both searches in the menu (one for semesters, one for departments), but this will only result in a network call the first time. Every other time you open the menu, it just returns immediately thanks to `cacheResponses` in [`d2l-search-widget`](https://github.com/Brightspace/d2l-search-widget-ui/blob/master/d2l-search-widget.html#L130).

d2l-all-courses.html
```diff
@@ -597,104 +301,28 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-					: this.localize('filtering.department');
-			},
 			_updateSortBy: function(e) {
 				this.set('_sortField', e.detail.value);
-
 				this.$.sortText.textContent = this.localize(this._sortTextOptions[this._sortField] || '');
-
 				this.$.sortDropdown.toggleOpen();
+			},
+			_updateTileSizes: function() {
```
ryantmer: Didn't modify this, just moved it down because everything was almost alphabetical anyway 😜

d2l-course-management-behavior.html
```diff
@@ -76,11 +76,32 @@
 				'enrollment-unpin-complete': '_onEnrollmentUnpinComplete'
 			},
 			observers: [
-				'_onPinnedEnrollmentsChanged(pinnedEnrollments.*)'
+				'_enrollmentsChanged(pinnedEnrollments.length)',
```
ryantmer: I think this is a simpler way to do this - we don't actually care if the _contents_ of either array changes, only if the number of elements does (for the purposes of `_hasEnrollments` and `_hasPinnedEnrollments`, that is).

d2l-filter-menu-content.html
```diff
@@ -0,0 +1,479 @@
...
+
+				if (this._currentFilters.length === 0) {
+					this.filterText = this.localize('filtering.filter');
+				} else if (this._currentFilters.length === 1) {
+					this.filterText = this.localize('filtering.filterSingle');
+				} else {
+					this.filterText = this.localize('filtering.filterMultiple', 'num', this._currentFilters.length);
+				}
+
+				this.fire('d2l-filter-menu-content-filters-changed', {
```
ryantmer: Changed the workflow within this component as well (to something I think is easier):

Clicking a filter item now just updates `_numSemesters` or `_numDepartments`, both of which trigger this observer. In this observer, we show/hide the Clear button, set the semester/department tab button text(s), then fire the event with `filterText` ("the menu opener text should be this" text) and `_currentFilters` (which becomes `_parentOrganizations` for the search in `d2l-all-courses`).

d2l-my-courses.html
```diff
@@ -284,11 +284,6 @@
 						type: 'splice'
 					}]);
 
-					this.set('enrollments', enrollmentEntities);
```
ryantmer: Yeah this just wasn't used anywhere.

d2l-all-courses.html
CodeBaboon: Worth noting that my alerts PR changes all of this (including removing the `visible` property from `d2l-alert` completely because why?

ryantmer: Ah, cool cool yeah no worries. I'll update as required. This is blocked on my other one anyway, so yours will defs be merged before this one.

d2l-all-courses.html
```diff
@@ -476,32 +234,22 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-				this.listen(this.$.filterDropdown, 'd2l-menu-item-change', '_updateFilter');
-
-				this.listen(this.$.filterDropdownContent, 'd2l-dropdown-open', '_onFilterDropdownOpen');
 				this.listen(this.$.filterDropdownContent, 'd2l-dropdown-close', '_onFilterDropdownClose');
-
-				this.listen(this.$.semesterSearchWidget, 'd2l-search-widget-results-changed', '_onSemesterSearchResults');
-				this.listen(this.$.departmentSearchWidget, 'd2l-search-widget-results-changed', '_onDepartmentSearchResults');
-
-				this.$.scrollThreshold.scrollTarget = this.$.filterDropdownContent.querySelector('.d2l-dropdown-content-container');
+				this.listen(this.$.filterMenuContent, 'd2l-filter-menu-content-filters-changed', '_onFilterChanged');
```
dbatiste: Perf optimization: preferred approach to wiring up listeners (including the resize one below) is:

```
attached: function() {
	Polymer.RenderStatus.afterNextRender(this, function() {
		this.listen( ... );
	}.bind(this));
}
```

Aside: wiring up via `listeners` happens even earlier than `attached`, and while less code, should be avoided for high-use components.  If the wire-up is going to happen in `attached` anyway, we should defer to `afterNextRender` when the handler is not needed immediately.  This is true for observers too.

Related note: I think `bind(this)` may not be necessary, but the linter complains without it.  I did a perf measurement and I did not measure any significant difference including the `bind` call.

d2l-filter-menu-content.html
```diff
@@ -0,0 +1,475 @@
...
+			}
+			.invisible {
+				visibility: hidden;
+			}
+			.hidden {
+				display: none !important;
+			}
+			button:hover,
+			button:focus,
+			.focus {
```
dbatiste: I think we should be careful using generic class names (ex. `focus`, `hidden`).  While these styles are scoped to the component with Shady-DOM, elements that have these classes may be unexpectedly styled is those classes are defined in the page (now or in the future) by something else (ex styles from legacy, MVC, or third-party libraries) due to Shady being leaky.

d2l-course-management-behavior.html
jstefaniuk-d2l: could you do something like: `'_enrollmentsChanged(pinnedEnrollments.length, unpinnedEnrollments.length)',`

d2l-all-courses.html
ryantmer: Not sure I fully follow this - is `afterNextRender` preferred over using `this.listen`? Just can't find much information about it, so - :confused:

Yeah, the `listeners` thing I think @CodeBaboon and I both stumbled across recently. If I understand correctly, the general best-practice is to just set up all listeners with `this.listen` inside the `afterNextRender` callback, in that case? (And also set up observers in `afterNextRender` as well?)

d2l-filter-menu-content.html
ryantmer: Yeah, fair - I'll rename this to something a bit more specific, thanks :)

d2l-course-management-behavior.html
ryantmer: Ah, I believe you are correct. (Below is the reasoning that was initially why I thought we needed only the one, but it turned out to support what you're proposing by the time I got to the end of it 😛 )

On the first fetch

- If the whole first page is pinned enrollments, then `pinnedEnrollments.length` changes -> `_hasPinnedEnrollments == _hasEnrollments == true`
- If the first page contains both, then `pinnedEnrollments.length` changes -> `_hasPinnedEnrollments == _hasEnrollments == true`
- If the user has no pinned enrollments, `pinnedEnrollments.length` doesn't change, so the handler doesn't get called, so `_hasEnrollments` doesn't get set to true (`_hasPinnedEnrollments` will be correct, since it defaults to false

After the first fetch we're okay just observing `pinnedEnrollments.length`, since a change to one array's length is 1:1 to a change with the other array's length.

But yeah, the no-pinned-enrollments-first-fetch case means we need to observe both. Good catch! 😄

d2l-filter-menu-content.html
```diff
@@ -0,0 +1,475 @@
...
+			is: 'd2l-list-item-filter',
+			properties: {
+				enrollmentEntity: {
+					type: Object,
+					value: function() { return {}; },
+					observer: '_onEnrollmentEntityChanged'
+				},
+				selected: {
+					type: Boolean,
+					value: false,
```
ryantmer: Hm, I don't believe it is, in fact. I'll make sure, and remove it if so, since this is a hella-repeated item (sometimes). Thanks!

d2l-all-courses.html
ryantmer: Looked into this a bit, as mentioned offline there were some issues in testing with this method. I'll make a GH issue so I don't forget about this, and will come back to it after this (since it's not strictly related anyway).

ryantmer:
Yay, this does what it's supposed to now! Pretty sure I haven't broken anything in the process of doing this - it was mostly just moving the filter menu content out into its own component and doing some minor tweaks/improvements along the way (+ tests).

d2l-filter-menu-content.html
```diff
@@ -0,0 +1,474 @@
...
+	<script>
+		'use strict';
+
+		Polymer({
+			is: 'd2l-filter-menu-content',
+			properties: {
+				filterText: String,
+				myEnrollmentsEntity: {
+					type: Object,
+					value: function() { return {}; },
```
jstefaniuk-d2l: if we're removing default values for perf, and check for null on all these anyway (I think), would it be good to remove these `function()` default values?

ryantmer: Oh, for sure - I've got https://github.com/Brightspace/d2l-my-courses-ui/issues/217 up for that.

Er, well actually that's for `afterNextRender`, but I was considering it to be included in that (I'll update it to explicitly mention it there).

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -1,119 +1,42 @@
...
-describe('smoke test', function() {
-	var server,
-		widget,
-		pinnedEnrollment = {
+describe('d2l-all-courses', function() {
+	var widget,
+		pinnedEnrollmentEntity,
+		unpinnedEnrollmentEntity;
+
+	before(function() {
```
jstefaniuk-d2l: I think we should ditch the "before", and throw this in "beforeEach" (in the other file as well)

PR #220, opened 2016-11-30
Update d2l-menu

ryantmer:
I do believe this is what's causing CI failures right now - I [override `_onSelectedChanged`](https://github.com/Brightspace/d2l-my-courses-ui/commit/bab7e658ece32cacdb72e2fccc9283c87ab45f3f#diff-98931a3c4fd43464bdaaf7d5c18ce360R88) in `d2l-list-item-filter`, but [it's been changed to `_onSelectedChanged`](https://github.com/Brightspace/d2l-menu-ui/commit/2947d3ddd1a97fcba5ac46c4bb4f42f6e3ba4094).

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -70,7 +70,9 @@
 				'd2l-menu-item-select': '_onSelect'
 			},
 			_onEnrollmentEntityChanged: function(entity) {
-				this.set('_organizationUrl', entity.getLinkByRel(/\/organization$/).href);
+				if (entity.getLinkByRel && entity.getLinkByRel(/\/organization$/)) {
+					this.set('_organizationUrl', entity.getLinkByRel(/\/organization$/).href);
+				}
 			},
 			_onOrganizationResponse: function(response) {
 				if (response.detail.status === 200 ) {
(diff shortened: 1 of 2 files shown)


ryantmer:
...or not. I will find zis bug.

ryantmer:
Theeere we go. This will have to go into a new release, I'll do that now.

d2l-all-courses.html
```diff
@@ -88,10 +90,10 @@
 				this.set('selected', !this.selected);
 				this.__onSelect(e);
 			},
-			__onSelectedChanged: function(e) {
+			_onSelectedChanged: function(e) {
```
dbatiste: It looks like the observer is still wired up to the old function name.  Does that need to get updated above?  Also, looking at that, is it necessary to set the default value of `selected` to `false`.  Setting default values on properties with observers should probably be avoided.

Apologies for the rename of the handler in d2l-menu behavior.  I incorrectly assumed it was private.  :(

ryantmer: Nah all good, I'm kind of doing a naughty thing here (and your change to the name is definitely the correct naming convention as well).

And yep, missed the observer whoops. I'll remove the default value while I'm at it too.

PR #222, opened 2016-12-02
Us77969 sync enrollments

apalaniuk:
Allows syncing enrollment state with an external component (eg, the course selector in the nav bar)

### DIFF
--- a/d2l-course-management-behavior.html
+++ b/d2l-course-management-behavior.html
@@ -1,5 +1,6 @@
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../d2l-siren-parser/d2l-siren-parser.html">
+<link rel="import" href="../d2l-polymer-behaviors/d2l-subscribe.html">
 <link rel="import" href="d2l-utility-behavior.html">
 
 <script>
@@ -70,6 +71,7 @@
 					value: false
 				}
 			},
+			_enrollmentIdMap: {},
 			listeners: {
 				'tile-remove-complete': '_onTileRemoveComplete',
 				'enrollment-pin-complete': '_onEnrollmentPinComplete',
@@ -96,6 +98,16 @@
 				// Remove enrollment from unpinned list, add to pinned
 				var enrollmentId = this.getEntityIdentifier(enrollment);
 
+				this._moveEnrollmentIdToPinnedList(enrollmentId);
+			},
+			_moveEnrollmentToUnpinnedList: function(enrollment) {
+				// Remove enrollment from pinned list, add to unpinned
+				var enrollmentId = this.getEntityIdentifier(enrollment);
+
+				this._moveEnrollmentIdToUnpinnedList(enrollmentId);
+			},
+			_moveEnrollmentIdToPinnedList: function(enrollmentId) 
+			{
 				for (var index = 0; index < this.unpinnedEnrollments.length; index++) {
 					var pinnedEnrollmentId = this.getEntityIdentifier(this.unpinnedEnrollments[index]);
 					if (pinnedEnrollmentId === enrollmentId) {
@@ -115,10 +127,7 @@
 					}
 				}
 			},
-			_moveEnrollmentToUnpinnedList: function(enrollment) {
-				// Remove enrollment from pinned list, add to unpinned
-				var enrollmentId = this.getEntityIdentifier(enrollment);
-
+			_moveEnrollmentIdToUnpinnedList: function(enrollmentId) {
 				for (var index = 0; index < this.pinnedEnrollments.length; index++) {
 					var unpinnedEnrollmentId = this.getEntityIdentifier(this.pinnedEnrollments[index]);
 					if (unpinnedEnrollmentId === enrollmentId) {
@@ -163,6 +172,16 @@
 					entity.actions[0].fields[0].value = true;
 					entity._actionsByName['pin-course'] = entity.actions[0];
 				}
+			},
+			_pinEnrollment: function(message) {
+				console.log('d2l-my-courses received notification that pin state for ' + message + ' was pinned');
+				console.log('I think the URL for this ID is ' + this._enrollmentIdMap[message]);
+				this._moveEnrollmentIdToPinnedList(this._enrollmentIdMap[message]);
+			},
+			_unpinEnrollment: function(message) {
+				console.log('d2l-my-courses received notification that pin state for ' + message + ' was unpinned');
+				console.log('I think the URL for this ID is ' + this._enrollmentIdMap[message]);
+				this._moveEnrollmentIdToUnpinnedList(this._enrollmentIdMap[message]);
 			}
 		};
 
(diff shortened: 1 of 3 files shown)


d2l-course-management-behavior.html
```diff
@@ -70,10 +70,9 @@
 					value: false
 				}
 			},
+			_enrollmentIdMap: {},
```
jstefaniuk-d2l: Should this be:

```
function() { return {}; },
```

apalaniuk: I think that's specific to properties that are part of the `properties` object

jstefaniuk-d2l: 👍 I think you're right

d2l-my-courses.html
```diff
@@ -358,6 +371,27 @@
 			_clearImageScrollThreshold: function() {
 				this.$['image-selector-threshold'].clearTriggers();
 			},
+			_onEnrollmentPinAction: function(e) {
+				var isPinned = e.type === 'enrollment-pinned';
+				var orgUnitId = /organizations\/([0-9]+)/
```
ryantmer: Ah, yeah - something like this is what I mean by the above comment. Only extracting the `orgUnitId` right before we actually need it.

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -409,6 +438,104 @@ describe('d2l-my-courses', function() {
 			widget._setCourseImageEvent(setCourseImageEvent);
 			expect(widget._alerts).not.to.include({ alertName: 'setCourseImageFailure', alertType: 'warning', alertMessage: 'Sorry, we\'re unable to change your image right now. Please try again later.' });
 		});
+
+		it('should bubble the correct d2l-course-pinned-change event when an enrollment is pinned', function(done) {
+			widget.fire = sinon.stub();
```
ryantmer: I... did not know you could do this. That is so much easier than `var stub = sinon.stub(widget, 'fire');`!

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -409,6 +438,104 @@ describe('d2l-my-courses', function() {
...
+				expect(widget.fire.calledWith('d2l-course-pinned-change',
+					sinon.match({
+						detail: {
+							orgUnitId: 1,
+							isPinned: true
+						}
+					})
+				));
+				done();
+			}, 0);
```
ryantmer: Nit: default `setTimeout` value is 0 anyway.

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -62,6 +62,26 @@ describe('d2l-my-courses', function() {
 					rel: ['self'],
 					href: '/enrollments/users/169/organizations/1'
 				}]
+			}, {
+				class: ['unpinned', 'enrollment'],
```
ryantmer: Looks like this was added, but not used.

d2l-my-courses.html
```diff
@@ -224,9 +227,14 @@
 					this.$.enrollmentsRootRequest.generateRequest();
 				}
 
+				document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
```
jstefaniuk-d2l: It seems to make the event go to the registered listener first before bubbling up the dom

test/d2l-my-courses/d2l-my-courses.js
apalaniuk: Ah, it's implicitly used by the "external pin event" tests, so that there's a pinned and unpinned enrollment returned by the same request.

d2l-my-courses.html
```diff
@@ -262,15 +270,20 @@
...
-							var enrollmentId = this.getEntityIdentifier(enrollment);
 							if (!this._pinnedCoursesMap.hasOwnProperty(enrollmentId)) {
 								newPinnedEnrollments.push(enrollment);
 								this._pinnedCoursesMap[enrollmentId] = true;
 							}
 						} else {
 							newUnpinnedEnrollments.push(enrollment);
 						}
+
+						this._enrollmentIdMap[orgUnitId] = enrollmentId;
```
ryantmer: Yeah, I had a set of PRs a while ago where I actually _removed_ the orgUnitID from the response ([here](https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/3614/diff#framework/tools/D2L.LP.Tools.Extensibility/Hypermedia/Organizations/Serializers/SirenOrganizationSerializer.cs) and [here](https://github.com/Brightspace/d2l-my-courses-ui/pull/150)), since it's not something the client really needs to know about. In this case we're kind of forced to, just by how the course selector works, but we can pretend for as long as possible that we don't.

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -357,6 +377,15 @@ describe('d2l-my-courses', function() {
 	});
 
 	describe('With enrollments', function() {
+		var pinnedEnrollmentEntity,
+			unpinnedEnrollmentEntity;
+
+		before(function() {
+			var parser = document.createElement('d2l-siren-parser');
```
jstefaniuk-d2l: may as well just throw this stuff in the beforeEach and avoid any potential error from someone modifying one of these values during a test

d2l-my-courses.html
dbatiste: It is the [useCapture](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) arg.

ryantmer: For sure - still not entirely sure of the implications of it, granted. Wasn't saying it's wrong or anything, I've never used it myself :)

ryantmer:
Rebased this, and had to slightly modify the tests (using `sinon.useFakeTimers` was introduced in the meantime, and the tests were using `setTimeout`). Going to address some of the code review and test, as this has been open for a while.

d2l-my-courses.html
ryantmer: Ah, in hindsight (and in attempting to do this), I've realized that this doesn't work particularly well - works fine for dispatching events, since we can trim down the `self` link to just the orgUnitId in that case, but when receiving events from the course selector, we would have to build up the `self` link, which is hella gross. Going to leave this as-is, since it works, and isn't a major issue in any sense anyway.

apalaniuk:
@ryantmer "No animations" is the intended behavior right now, as per Bob.

ryantmer:
@apalaniuk Cool, works for me then :) As for the needing to fetch the enrollment/organization for "new" org units, I would say that would be a fairly required thing for this functionality to be considered complete. Not sure if you want to do it in a new PR or this one, granted, whatever works for you I'd say!

apalaniuk:
@ryantmer I'll do it in a new PR if that's ok with you =)

ryantmer:
@apalaniuk Yep sgtm. I left some testing notes on the LP PR [here](https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/3978/overview?commentId=202830), LGTM overall.

PR #223, opened 2016-12-02
Lint more things

ryantmer:
Was missing the image-selector and components folders. Now just does everything except the demos!


ryantmer:
I am already going to abuse the "requires formal approval" thing I just turned on yesterday 😛

PR #225, opened 2016-12-06
animate the alerts and delay image failure

CodeBaboon:
Bob requested two things:
1) the warning alert should not show up until the animation on the tile also does (and that one waits one second before turning into the X); and
2) the alerts should animate in, sort of 'dropping' into place similar to how the dropdown content does if you click on the little bell up in the main brightspace nav bar.


jstefaniuk-d2l:
Code LGTM

CodeBaboon:
Rebased from master.

PR #226, opened 2016-12-07
US79521 Translations

ryantmer:
These were copied in from the translations that were attached to the Rally story. Also cleaned up bower.json - updated some minor/patch versions.


bower.json
```diff
@@ -37,18 +37,17 @@
...
     "iron-a11y-keys": "^1.0.6",
-    "iron-ajax": "PolymerElements/iron-ajax#^1.2.0",
     "iron-input": "^1.0.10",
     "iron-menu-behavior": "^1.1.9",
-    "iron-overlay-behavior": "https://github.com/PolymerElements/iron-overlay-behavior.git#^1.9.0",
+    "iron-overlay-behavior": "^1.10.1",
     "iron-pages": "^1.0.8",
-    "neon-animation": "PolymerElements/neon-animation#^1.2.2",
+    "neon-animation": "^1.2.4",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
```
ryantmer: Annoyingly, the `iron-scroll-threshold` in bower is _not_ `PolymerElements/iron-scroll-threshold`. So, have to reference the git repo here.

localize-behavior.html
```diff
@@ -77,19 +77,530 @@
 								'sorting.lastAccessed': 'Last Accessed',
 								'images.noResults': 'No results found for {search}.',
 								'images.pleaseModify': 'Please check your spelling or modify your search.',
-								'error.displayingImages': "Oh no, we're having problems displaying images. Please check back soon.",
-								'error.settingImage': "Sorry, we're unable to change your image right now. Please try again later."
+								'error.displayingImages': 'Oh no, we\'re having problems displaying images. Please check back soon.',
```
CodeBaboon: I think because the double quotes on the outside allowed the single quotes inside to not be escaped.

CodeBaboon: Ah I see, because linting. Hmm, that is interesting...

ryantmer: It is a mystery

PR #227, opened 2016-12-07
Change back to Brightspace/app-localize-behavior

ryantmer:
...

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,7 @@
     "/demo/"
   ],
   "dependencies": {
-    "app-localize-behavior": "^0.10.1",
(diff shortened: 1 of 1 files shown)


CodeBaboon:
Why back to the repo?

ryantmer:
The Brightspace one has [this commit](https://github.com/Brightspace/app-localize-behavior/commit/8b45c60c3800ba8df46373150a94138be377b876) in particular, which addresses some IE concerns. I'm not sure if those concerns have been addressed in the PolymerElements version of it, but if so, then everything should be updated. (Right now, causing bower conflicts when I try to update BSI.)

PR #228, opened 2016-12-07
Update to latest d2l-colors (no include required).

dbatiste:
This change updates to the latest d2l-colors, where the palette is defined in a custom-style element (so no include needed).  This change removes the `include="d2l-colors"` to eliminate browser console warnings.


dbatiste:
@ryantmer : are you able to review?

CodeBaboon:
I will take a look

Issue #191, opened 2016-11-11
Remove hardcoded advanced search URL


ryantmer:
Right now, [the advanced search URL](https://github.com/Brightspace/d2l-my-courses-ui/blob/fe922292bfa82a0941f677ec01548a9b74834e53/d2l-search-widget-custom.html#L571) is hardcoded. Pretty sure this is because we're not planning on having that link in there long-term, but if we are then that will have to be generated - for example, 6606 is not used everywhere, which means the link doesn't work.

ctwomey1:
fixed in #241

PR #229, opened 2016-12-07
Search course tiles

ctwomey1:
Enables live search for course tiles

d2l-search-enrollment-response event may be moved to only occur on search button press/enter key.


d2l-search-widget-custom.html
```diff
@@ -344,7 +344,7 @@
 			},
 
 			_onSearchButtonClick: function() {
-				this._advancedSearch(this._searchString);
```
jstefaniuk-d2l: Looks like we should be removing advancedSearch and computePreviousSearchLink

jstefaniuk-d2l: Essentially updating or removing anything using those 2

ryantmer: eyyy that would fix #191 as well

d2l-my-courses.html
```diff
@@ -273,34 +273,29 @@
...
 					this.fire('recalculate-columns');
+					this.toggleClass('hidden', true, this.$.lazyLoadSpinner);
+					this.$['all-courses-scroll-threshold'].clearTriggers();
+					this._showContent();
+					this._lastEnrollmentsSearchResponse = enrollments;
 				}
+			},
+			_updateEnrollmentEntities: function(e, detail) {
+				this._pinnedCoursesMap = {};
+				this.splice('pinnedEnrollments',  0, this.pinnedEnrollments.length);
```
jstefaniuk-d2l: if we're able to change these back to
```
this.pinnedEnrollments = [];
this.unpinnedEnrollments = [];
```

It's easier to scan

jstefaniuk-d2l: (or use `set`)

d2l-course-tile.html
```diff
@@ -259,6 +258,15 @@
 
 				this._pendingPinAction = this.pinned;
 			},
+			_enrollmentChanged: function() {
+				if (!this.delayLoad) {
```
jstefaniuk-d2l: We should probably verify here that enrollment exists or if someone sets it to null for whatever reason we'll crash

d2l-course-tile.html
```diff
@@ -259,6 +258,15 @@
...
+					var organizationLink = this.enrollment.getLinkByRel(/\/organization$/);
+					if (this._organizationUrl !== organizationLink.href + '?embedDepth=1' ) {
```
ctwomey1: To match, it was mostly a last minute "optimization".  If it's preferable to use .indexOf i can do that

CodeBaboon: I just think we don't want to have to update this code every time someone adds something to the query string or changes the embed depth.

jstefaniuk-d2l: We could also do something like:

```
_organizationUrl.slice(organizationsLink.href.length) !== organizationsLink.href
```

but `indexOf === 0` is much easier to read

ctwomey1:
Made changes based on feedback, however due to the way i'm changing the pinnedEnrollments there's a larger problem when closing the dialog.

d2l-my-courses.html
```diff
@@ -273,34 +273,29 @@
 
 					this.pinnedEnrollments = this.pinnedEnrollments.concat(newPinnedEnrollments);
 					this.unpinnedEnrollments = this.unpinnedEnrollments.concat(newUnpinnedEnrollments);
+					this.enrollments = enrollmentEntities;
```
apalaniuk: Is this a thing? I can't seem to find it being used or defined anywhere else..

d2l-my-courses.html
```diff
@@ -273,34 +273,29 @@
...
 					var colNum = this._calcNumColumns(this._getAvailableWidth(Polymer.dom(this.root).node.host), this.pinnedEnrollments.length);
 					this._tileSizes = (colNum === 2) ?
 						{ mobile: { maxwidth: 767, size: 50 }, tablet: { maxwidth: 1243, size: 33 }, desktop: { size: 20 } } :
 						{ mobile: { maxwidth: 767, size: 100 }, tablet: { maxwidth: 1243, size: 67 }, desktop: { size: 25 } };
 
 					this.fire('recalculate-columns');
+					this.toggleClass('d2l-all-courses-hidden', true, this.$.lazyLoadSpinner);
+					this.$['all-courses-scroll-threshold'].clearTriggers();
+					this._showContent();
+					this._lastEnrollmentsSearchResponse = enrollments;
```
apalaniuk: I *think* we should remove the setting of this property [here](https://github.com/Brightspace/d2l-my-courses-ui/blob/64a919ea681e132d820c4d7373d5b6df1483642f/d2l-my-courses.html#L127), as it will always be a parsed Siren entity rather than the raw response. Then we can use it [here](https://github.com/Brightspace/d2l-my-courses-ui/blob/e9021a791e5a0c24efe6f92bd2eadfd0bdbaabc4/d2l-my-courses.html#L332) as an entity directly without parsing.

d2l-search-widget-custom.html
```diff
@@ -252,6 +252,7 @@
 			],
 
 			ready: function() {
+				this._searchUrl = '';
```
apalaniuk: nit: The default value can be specified on the property [here](https://github.com/Brightspace/d2l-my-courses-ui/blob/64a919ea681e132d820c4d7373d5b6df1483642f/d2l-search-widget-custom.html#L234), as in the [default value docs](https://www.polymer-project.org/1.0/docs/devguide/properties#configure-values), which looks a bit cleaner to me (unless there's some perf benefit to delaying it until `ready` that I wouldn't expect but is a real thing)

apalaniuk:
Hmm. Should the dropdown on the search bar still exist at all? If not, I think we can just cull all of that. If it should remain, I think selecting an item under "recent searches" must populate the search bar and force a search (previously, it would have a link attached that would perform the same action as would be done when hitting enter, but the computer function that would do that was removed). 

Then again, I think it would require rethinking what "recent searches" represents, as previously it represented an advanced search (hitting enter) and that was taken out.

d2l-search-widget-custom.html
```diff
@@ -503,8 +491,8 @@
 							if (e.detail.item.dataset.text) {
 								this._addSearchToHistory(e.detail.item.dataset.text);
 							}
-
-							window.location.href = e.detail.item.dataset.href;
+							this._searchString = e.detail.item.dataset.text;
```
apalaniuk: So if the dropdowns should still exist, I think there's two cases now - recent searches, where there's no href as that was a link to the legacy search and has been removed, and live search results, where a link exists. I think you'd want to perform the old (removed) behavior on selecting a live search result from the dropdown (add search to history and navigate to the URL), and use this new behavior (performing a search) when selecting a recent search.

ryantmer:
I believe the general approach was to be Google-search-esque - the search dropdown is a live-updated list of possible results. Clicking on a result (or pressing Enter at any time) does the full search, and the tiles update. I don't know that the Recent Searches list was addressed directly, but I would personally expect it to act the same as the live search - click in field, recent searches show up, click recent search, course tiles update to show results.

d2l-search-widget-custom.html
ctwomey1: Ryan's comment below is accurate i believe.  Bob doesn't want the context to shift away from the screen, however I  will clarify with him

d2l-search-widget-custom.html
ctwomey1: I think Default value only gets called on init of the control, ready gets called every time it opens the all-courses view

d2l-search-widget-custom.html
```diff
@@ -335,16 +336,8 @@
...
-				if (searchString) {
-					var queryString = this._advancedSearchUrl + searchString;
-					this._addSearchToHistory(searchString);
-					window.location.href = queryString;
-				}
-			},
-
 			_onSearchButtonClick: function() {
-				this._advancedSearch(this._searchString);
+				this._search();
```
ryantmer: The live search request is overriding `pageSize` to be 5, just to make the live search dropdown not-that-large. I'm assuming the plan is to not use the same search function anyway though, correct? (As right now, full-course-tile-searching is happening on every keystroke which :fire: :fire: :fire: the servers)

d2l-search-widget-custom.html
```diff
@@ -356,14 +349,15 @@
...
+
 				}
 			},
 
 			_onSearchStringChanged: function(newString) {
 				if (this._listboxes) {
 					if (newString.length > 0) {
 						if (!this.isDebouncerActive('liveSearchJob')) {
-							this.debounce('liveSearchJob', this._liveSearch.bind(this), 250);
+							this.debounce('liveSearchJob', this._search.bind(this), 750);
```
ryantmer: Many :+1:s, many 🚒 🚒 🚒 .

d2l-search-widget-custom.html
```diff
@@ -356,14 +349,15 @@
 					this.$$('iron-pages').select('recent-searches-page');
 				} else {
 					this.$.dropdown.close();
+
```
ryantmer: image

d2l-search-widget-custom.html
```diff
@@ -516,9 +504,6 @@
 				ENTER: 13
 			},
 
-			// URL for advanced search functionality (existing advanced search page from legacy My Courses widget).
-			_advancedSearchUrl: 'le/manageCourses/search/6606?searchTerm=',
```
ryantmer: 🎆 🎆 🎆 

This tangentially fixes #191, whoop

ctwomey1:
Since there were so many issues with this branch i've "started over" on https://github.com/Brightspace/d2l-my-courses-ui/tree/course-live-search.   I'll make a proper pull request when it's in a fully complete state

PR #230, opened 2016-12-08
One missed class, and one missed class change

### LINKED ISSUE #218: Change overly-generic class names
Class names like `hidden` are ripe for problems - should change them to be more specific. (This was mentioned recently, and surprise surprise it came up less than two days later.)

ryantmer:
Follow-up to #224, which itself (mostly) fixed #218.


trevorjonesd2l:
Code looks this I'll test

PR #233, opened 2016-12-09
Rework delayLoad slightly

ryantmer:
Mostly just a rename, but also it wasn't ever being set in d2l-my-courses itself, so removed it from there.

This is a super low priority, QOL kinda thing.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -184,7 +184,7 @@
 			id="all-courses-pinned"
 			enrollments="{{pinnedEnrollments}}"
 			enrollments-queue="{{pinnedEnrollmentsQueue}}"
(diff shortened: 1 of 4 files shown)


CodeBaboon:
LGTM. Seems to work (tiles still load, but not until the all courses overlay is loaded).

CodeBaboon:
In the spirit of the retro, my test plan:
- Launched Brightspace and verified that my pinned tiles loaded.
- Clicked View All Courses and verified that those tiles loaded.
- Closed View All Courses, opened debug tools, and hard-refreshed
- Noted that when displaying the pinned courses only /organizations HM calls were made for the tiles pinned and visible.
- Clicked View All Courses and noted that the network traffic showed /organizations HM calls for all of the tiles being loaded on the All Courses screen.

PR #235, opened 2016-12-09
added new PR process information

cgalvind2l:
AC: the process for testing that we discussed during retro is made clear in the README
Areas of risk: oh geez none I hope

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ In the description of your pull request, please include the following informatio
 
 3. Areas of risk that should be addressed during testing
 
+Before your PR is merged, it should be tested thoroughly, and the person who does this testing should describe their test steps in detail in a comment. These steps should then be reviewed by the developer and assessed for gaps before the PR gets merged.
+
 > Note: To contribute, please create a branch in this repo instead of a fork.
 We are using [Sauce Labs](https://saucelabs.com/) in our CI builds which don't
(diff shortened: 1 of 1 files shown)


CodeBaboon:
I don't see a US# :p

cgalvind2l:
Ummm you'll note that the documentation says IF APPLICABLE.

ryantmer:
I like the level of sass in this PR.

CodeBaboon:
Should we indicate that when testing a PR there should be detailed test steps and that those should be reviewed before being merged? Or is that a separate PR?

cgalvind2l:
I was unsure what level of detail to put in the README for the repo; I was thinking not to put the entire process in (as that is team specific rather than codeline specific). But I think that adding the bit about test steps and reviewing them is probably a good idea.

CodeBaboon:
Docs LLTM. Tested with the following steps:

- used eyeballs to read words
- used mindballs to comprehend words
- used memoryballs to crossreference with earlier discussion
- used wordballs to discuss gaps with developer
- used thumbballs to type results here

cgalvind2l:
Wow, that's way more balls than I would typically expect to be involved in a PR.

I have evaluated your test plan for gaps and have found that they were all filled with balls. So should be good. Just need someone to approve so it can be merged...

PR #236, opened 2016-12-12
US78151 - TA108231 Break out d2l-list-item-filter

ryantmer:
This is a bit of precursor work to creating the simplified filter menu view. The only noticeable effect should be that I un-hid the search/filter stuff.

I moved `d2l-filter-menu-content` into a folder, and broke out `d2l-list-item-filter` from it into its own file (and updated tests/links accordingly). Just wanted to do this large-but-no-real-change diff separately, so that the real change(s) are nicer to code review/test.

Acceptance Criteria/risks:

- Filter/search is visible again
- Filter menu still loads content normally
- Filter menu still has tabbed view, and lists departments and semesters in separate tabs
- Not really much risk 😐 Pretty sure this either works, or a bunch of errors would happen due to misaligned directories/import paths

@jstefaniuk-d2l [called it.](https://github.com/Brightspace/d2l-my-courses-ui/pull/207#discussion_r88941050)

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -12,7 +12,7 @@
 <link rel="import" href="d2l-course-management-behavior.html">
 <link rel="import" href="d2l-course-tile-grid.html">
 <link rel="import" href="d2l-course-tile-responsive-grid-behavior.html">
-<link rel="import" href="d2l-filter-menu-content.html">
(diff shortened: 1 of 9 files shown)


jstefaniuk-d2l:
Code LGTM

PR #237, opened 2016-12-13
US78151 - TA108233 Refactor d2l-filter-menu-content

ryantmer:
This breaks out some of the behaviour of `d2l-filter-menu-content` into `d2l-filter-menu-content-tabbed`. The idea behind this is that a user with a smaller number (<=7) of department/semester enrollments will see a simplified view. Doing so with the current state of things would be a pain, so I've broken out the (_significantly_ more complex) tabbed view into its own component. In the next PR, `d2l-filter-menu-content` will choose which view to present based on the number of department/semester enrollments for the current user; as of right now, this is behaviourally unchanged (always shows the tabbed view). Like #236, this is just a "the diff would make you swear fealty to Satan if I did this all once" kind of PR.

This also hopefully will make a transition to [`d2l-tabs`](https://github.com/Brightspace/d2l-tabs) easier, should we decide to use it (we should, pending some updates to it).

~~WIP-ish since this includes commits in #236, so that'll have to be merged first.~~

Acceptance Criteria:

- No behavioural change - filter menu still shows tabs for department and semester; lazy loading of additional departments/semesters still works; filter menu still hides itself when the user has a small number of enrollments; etc.

Risks:

- This is a refactor, meaning things moved around a fair amount, so it's possible that things are no longer "connected". #216 means that this is a bit less risky, since as long as `d2l-filter-menu-content` is still emitting the two events it normally emits at the right times, then we should be good.
- I've added an extra two network calls. I can't imagine this will have any significant performance impact, but it's worth noting. They're also semi-superfluous, but I haven't yet thought of a way to dedupe them intelligently.

TODO:

- [x] Merge #236 
- [x] rebase dis shiz
- [x] buy a Lamborghini

### DIFF
--- a/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
+++ b/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
@@ -1,0 +1,336 @@
+<link rel="import" href="../../polymer/polymer.html">
+<link rel="import" href="../../iron-scroll-threshold/iron-scroll-threshold.html">
+<link rel="import" href="../../d2l-ajax/d2l-ajax.html">
+<link rel="import" href="../../d2l-colors/d2l-colors.html">
+<link rel="import" href="../../d2l-icons/d2l-icons.html">
+<link rel="import" href="../../d2l-menu/d2l-menu.html">
+<link rel="import" href="../../d2l-search-widget/d2l-search-widget.html">
+<link rel="import" href="../../d2l-siren-parser/d2l-siren-parser.html">
+<link rel="import" href="../localize-behavior.html">
+<link rel="import" href="d2l-list-item-filter.html">
+
+<dom-module is="d2l-filter-menu-content-tabbed">
+	<template>
+		<style>
+			:host {
+				display: block;
+			}
+			.d2l-filter-menu-content-invisible {
+				visibility: hidden;
+			}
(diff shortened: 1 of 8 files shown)


d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -0,0 +1,336 @@
+<link rel="import" href="../../polymer/polymer.html">
```
ryantmer: Sigh... wish the diff was a bit smarter here. This is the previous `d2l-filter-menu-content`, copied into this file, then removed stuff that stayed in the new `d2l-filter-menu-content`, plus some minor changes to make things work. Unfortunately code review for this would suck; testing would probably be the best verification here. (Nothing should change, outwardly-speaking.)

dbatiste:
@ryantmer : the [d2l-tabs](https://github.com/Brightspace/d2l-tabs) is a WIP, and I don't think it's ready for general consumption.  It's not clear whether it follows a Daylight design.  That being said, I think it needs some updates and further review before making it public and available in BrightspaceUI.

ryantmer:
@dbatiste Yeah, noticed that. Definitely not ready for implementing here just yet, but would be legit to get them up to Daylighting and use them here :)

PR #240, opened 2016-12-15
Font size and weight updates for all courses.

njostonehouse:


d2l-course-tile-styles.html
```diff
@@ -127,7 +127,7 @@
 			position: relative;
 			margin-top: 0.4rem;
 			/* overrides to d2l-heading-4 */
-			font-weight: 400;
+			font-weight: 700;
```
ryantmer: Hm, this one should definitely be lighter than 700 - is there another `d2l-heading` we should use instead?

d2l-search-listbox.html
```diff
@@ -49,7 +49,7 @@
 			}
 
 			::content > [data-list-title] {
-				font-weight: 400;
+				font-weight: 700;
```
ryantmer: This one is a big jump too, but I think it's okay since context-wise, it's the only even-somewhat-bold text in the area.

d2l-course-tile-styles.html
njostonehouse: I think just removing this font-weight override will work. From what I can see, `d2l-typography-4` isn't impacting the style anyway, so there's nothing significant to override.

d2l-search-listbox.html
njostonehouse: I think this needs to stand out relative to the other items in the list (weight 300 -> 400), and 700 is currently how we want to do that...

d2l-course-tile-styles.html
ryantmer: Ah, yeah good catch! Removing it is even better, thanks :)

PR #241, opened 2016-12-15
Course live search

ctwomey1:
Take 2 of my last pr.

Some things of note:

- Course image management behavior was causing too much coupling between the two views.  I'd consider partially coupling them together through the behavior, but i feel like the two views are fundamentally different.  Very open to suggestions, this was a "make it work" type of release
- debounce functions weren't setup correctly before, i've made a change that should reduce request spam, as well as introduced more debounce 
- Polymer is really really bad at notifying for arrays with as much change as i'm doing.  As per the documentation i'm doing https://www.polymer-project.org/1.0/docs/devguide/model-data#override-dirty-check to make the courses correctly work.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -14,6 +14,7 @@
 <link rel="import" href="d2l-filter-menu-content.html">
 <link rel="import" href="d2l-search-widget-custom.html">
 <link rel="import" href="localize-behavior.html">
+<link rel="import" href="d2l-simple-overlay.html">
 
 <dom-module id="d2l-all-courses">
 	<template>
@@ -125,7 +126,7 @@
 			id="enrollmentsSearchRequest"
 			url="[[_enrollmentsSearchUrl]]"
 			headers='{ "Accept": "application/vnd.siren+json" }'
-			on-iron-ajax-response="_updateFilteredEnrollments"
+			on-iron-ajax-response="_enrollmentsSearchResponse"
 			last-response="{{lastEnrollmentsSearchResponse}}">
 		</d2l-ajax>
 		<d2l-simple-overlay
@@ -243,7 +244,8 @@
 					value: function() {
 						return [];
 					},
-					notify: true
+					notify: true,
+					observer: '_setFilteredPinnedEnrollments'
 				},
 				// filtered pinned enrollment entities
 				filteredPinnedEnrollments: {
@@ -258,7 +260,8 @@
 					value: function() {
 						return [];
 					},
-					notify: true
+					notify: true,
+					observer: '_setFilteredUnpinnedEnrollments'
 				},
 				filteredUnpinnedEnrollments: {
 					type: Array,
@@ -469,7 +472,10 @@
 				this._pinnedCoursesMap = {};
 				this._updateFilteredEnrollments(detail, false);
 			},
-			_updateFilteredEnrollments: function(response, append = true) {
+			_enrollmentsSearchResponse: function(response) {
+				this._updateFilteredEnrollments(response, true);
+			},
+			_updateFilteredEnrollments: function(response, append) {
 				if (response.detail.status === 200) {
 					var parser = document.createElement('d2l-siren-parser');
 					var enrollments = parser.parse(response.detail.xhr.response);
@@ -540,6 +546,12 @@
 				var itemCount = Math.max(this.pinnedEnrollments.length, this.unpinnedEnrollments.length);
 				return itemCount;
 			},
+			_setFilteredPinnedEnrollments: function() {
+				this.filteredPinnedEnrollments = this.pinnedEnrollments;
+			},
+			_setFilteredUnpinnedEnrollments: function() {
+				this.filteredUnpinnedEnrollments = this.unpinnedEnrollments;
+			},
 			load: function() {
 				// Load course tile and filter contents. Called from d2l-my-courses.
 				this.$$('#all-courses').open();
@@ -547,8 +559,6 @@
 				this.$['all-courses-scroll-threshold'].clearTriggers();
 				this.$['search-widget']._searchString = '';
 				this._pinnedCoursesMap = {};
-				this.filteredPinnedEnrollments = this.pinnedEnrollments;
-				this.filteredUnpinnedEnrollments = this.unpinnedEnrollments;
 				this.set('delayLoad', false);
 
 				// Only make the call to load the filter menu contents if user has a
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -182,8 +182,8 @@
 
 		<d2l-course-tile-grid
 			id="all-courses-pinned"
-			enrollments="{{pinnedEnrollments}}"
-			enrollments-queue="{{pinnedEnrollmentsQueue}}"
+			enrollments="{{filteredPinnedEnrollments}}"
+			enrollments-queue="{{filteredPinnedEnrollmentsQueue}}"
 			delay-load="[[delayLoad]]"
 			tile-sizes="[[_tileSizes]]">
 		</d2l-course-tile-grid>
@@ -192,8 +192,8 @@
 
 		<d2l-course-tile-grid
 			id="all-courses-unpinned"
-			enrollments="{{unpinnedEnrollments}}"
-			enrollments-queue="{{unpinnedEnrollmentsQueue}}"
+			enrollments="{{filteredUnpinnedEnrollments}}"
+			enrollments-queue="{{filteredUnpinnedEnrollmentsQueue}}"
 			delay-load="[[delayLoad]]"
 			tile-sizes="[[_tileSizes]]">
 		</d2l-course-tile-grid>
@@ -205,6 +205,84 @@
 		Polymer({
 			is: 'd2l-all-courses',
 			properties: {
+					// Set of pinned enrollment Entities
+				pinnedEnrollments: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				// filtered pinned enrollment entities
+				filteredPinnedEnrollments: {
+					type: Array,
+					value: function() {
+						return [];
+					}
+				},
+				// Set of unpinned enrollment Entities
+				unpinnedEnrollments: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				filteredUnpinnedEnrollments: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				// Set of enrollments Entities that are being transferred from unpinned to pinned state
+				// Only used by `d2l-all-courses`
+				pinnedEnrollmentsQueue: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				filteredPinnedEnrollmentsQueue: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				// Set of enrollments that are being transferred from pinned to unpinned state
+				// Only used by `d2l-all-courses`
+				unpinnedEnrollmentsQueue: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				filteredUnpinnedEnrollmentsQueue: {
+					type: Array,
+					value: function() {
+						return [];
+					},
+					notify: true
+				},
+				// If enabled, the component will use the pinned/unpinned enrollment queues
+				// In the case of `d2l-all-courses`, the enrollments are switching lists, so this is required.
+				usePendingLists: {
+					type: Boolean,
+					value: false
+				},
+				// True when there are pinned enrollments (i.e. `pinnedEnrollments.length > 0`)
+				_hasPinnedEnrollments: {
+					type: Boolean,
+					value: false
+				},
+				// True when there are any enrollments (pinned or unpinned)
+				_hasEnrollments: {
+					type: Boolean,
+					value: false
+				},
 				// Sets the delay-load property on the course tile grid
 				delayLoad: {
 					type: Boolean,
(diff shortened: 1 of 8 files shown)


ryantmer:
(Side note that this fixes #191, so should close that once this is merged)

ctwomey1:
@ryantmer #191 should stay open because we want to still expose the advanced search link to users.  This will have to be done by exposing a parameter for my courses and updating the renderer in the lms

ryantmer:
Yep, although would be its own thing I'd imagine. #191 was just for removing it, since it was bad form (and eventually, the bad form caught up with us 😿 )

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -377,11 +377,12 @@ describe('d2l-my-courses', function() {
 				});
 
 			widget.$.enrollmentsRootRequest.generateRequest();
-
-			// Wait until the second (search) request finishes before checking things
-			widget.$.enrollmentsSearchRequest.addEventListener('iron-ajax-response', function() {
+			var doneFunc = function() {
 				done();
-			});
+				widget.$.enrollmentsSearchRequest.removeEventListener('iron-ajax-response', doneFunc);
```
jstefaniuk-d2l: For clarity, I'd put this before `done()`

d2l-all-courses.html
```diff
@@ -113,11 +114,37 @@
 				text-decoration: underline;
 				color: var(--d2l-color-celestine);
 			}
+			#lazyLoadSpinner {
```
jstefaniuk-d2l: Why bother having 2 different css entries for the same thing?

`#lazyLoadSpinner` and `d2l-loading-spinner`

d2l-my-courses.html
```diff
@@ -197,27 +187,38 @@
...
 				return this.pinnedEnrollments.length;
 			},
+			_rootEnrollmentRequest: function() {
+				this._pinnedCoursesMap = {};
+				this._unpinnedCoursesMap = {};
+				this.unpinnedEnrollments = [];
+				this.pinnedEnrollments = [];
+				this.debounce('d2l-root-request', this._rootRequest, 1000);
+			},
+			//encapsulates root enrollment request, the debounce method loses context when directly using this
```
jstefaniuk-d2l: Using `bind` may solve the loss of context
```
this.debounce('d2l-root-request', this.$.enrollmentsRootRequest.generateRequest.bind(this), 1000);
```

d2l-my-courses.html
```diff
@@ -288,8 +276,7 @@
 					this.fire('recalculate-columns');
 				}
 
-				this.toggleClass('d2l-my-courses-hidden', true, this.$.lazyLoadSpinner);
-				this.$['all-courses-scroll-threshold'].clearTriggers();
+				//this.toggleClass('d2l-my-courses-hidden', true, this.$.lazyLoadSpinner);
```
jstefaniuk-d2l: leftover comment

d2l-my-courses.html
```diff
@@ -302,7 +289,7 @@
 			_openAllCoursesView: function(e) {
 				e.preventDefault();
 				e.stopPropagation();
-				this.$$('#all-courses').open();
+				this.$$('d2l-all-courses').lastEnrollmentsSearchResponse = this._lastEnrollmentsSearchResponse;
```
jstefaniuk-d2l: This is already bound:

```
+		<d2l-all-courses
 +			pinned-enrollments="{{pinnedEnrollments}}"
 +			unpinned-enrollments="{{unpinnedEnrollments}}"
 +			my-enrollments-entity="[[_lastEnrollmentsSearchResponse]]"
 +			lastEnrollmentsSearchResponse="[[_lastEnrollmentsSearchResponse]]"
 +			locale="[[locale]]">
 +		</d2l-all-courses>
```

So may not be needed

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -432,7 +433,7 @@ describe('d2l-my-courses', function() {
 			widget._addAlert('warning', 'setCourseImageFailure', 'failed to do that thing it should do');
 			clock.tick(1001);
 			expect(widget._alerts).to.include({ alertName: 'setCourseImageFailure', alertType: 'warning', alertMessage: 'failed to do that thing it should do' });
-			widget.$['all-courses']._handleClose();
+			widget.$$('d2l-all-courses').children['all-courses']._handleClose();
```
jstefaniuk-d2l: I think you may be able to just chain the $ calls here:
```
widget.$['d2l-all-courses'].$['all-courses']._handleClose();
```

d2l-my-courses.html
ctwomey1: I've found that the binding doesn't always push down. I'd be open to testing it out deeper though

test/d2l-my-courses/d2l-my-courses.js
ryantmer: Is this necessary anyway? `widget` should be getting reset between tests

d2l-search-widget-custom.html
```diff
@@ -486,28 +515,24 @@
...
-						if (e.detail.item.dataset.href) {
-							if (e.detail.item.dataset.text) {
-								this._addSearchToHistory(e.detail.item.dataset.text);
-							}
-
-							window.location.href = e.detail.item.dataset.href;
-						}
-						break;
+			_onSearchResponseTiles: function(response) {
+				if (response.detail.status === 200 && this._searchString) {
```
jstefaniuk-d2l: Are we able to pull the search string off of the response instead of `this._searchString`?

d2l-search-widget-custom.html
```diff
@@ -335,16 +347,40 @@
 				}
 			},
 
-			_advancedSearch: function(searchString) {
-				if (searchString) {
-					var queryString = this._advancedSearchUrl + searchString;
-					this._addSearchToHistory(searchString);
-					window.location.href = queryString;
+			// Calls the search-my-enrollments action, used to populate live search results in the listbox
+			_search: function() {
```
jstefaniuk-d2l: most of this and `_livesearch` is duplicated

test/d2l-my-courses/d2l-my-courses.js
jstefaniuk-d2l: This is in a beforeEach, so we don't want done getting called twice when there's a second request

d2l-search-widget-custom.html
```diff
@@ -486,28 +515,24 @@
...
+				}
+			},
+			_onIronSelect: function() {
+				var ironPages = this.$$('iron-pages');
+				var pageIndex = ironPages.indexOf(ironPages.selectedItem);
+				this._currentListbox = this._listboxes[pageIndex];
+			},
+			// Handles iron-activate events, which are fired when listbox items are selected and dropdown pages are changed
+			_onIronActivate: function(e) {
+				if (e.detail.item.dataset.text) {
```
jstefaniuk-d2l: Maybe put `e.detail.item.dataset.text` into a variable since we use it 3 times

d2l-search-widget-custom.html
```diff
@@ -239,10 +246,17 @@
 					value: function() {
 						return [];
 					}
+				},
+				_searchResponse: {
```
jstefaniuk-d2l: I don't think this is being used

d2l-all-courses.html
ctwomey1: mostly because it was just ported directly when i moved the elements, I should have combined them at that point.

d2l-search-widget-custom.html
ryantmer: It would be in the URL, but I think we'd have to regex and de-URL-encode :(

d2l-search-widget-custom.html
```diff
@@ -486,28 +515,24 @@
...
-						break;
+			_onSearchResponseTiles: function(response) {
+				if (response.detail.status === 200 && this._searchString) {
+					this.fire('d2l-search-enrollment-response', response);
+				}
+			},
+			_onIronSelect: function() {
+				var ironPages = this.$$('iron-pages');
+				var pageIndex = ironPages.indexOf(ironPages.selectedItem);
+				this._currentListbox = this._listboxes[pageIndex];
```
ryantmer: I'm thinking it might be possible to get rid of all this `_currentListBox` stuff by just using the `hidden` attribute in a way like #239 is doing it for choosing which filter menu to display. Probably a future PR though, unless it makes other things easier to do here.

d2l-search-widget-custom.html
```diff
@@ -356,27 +392,27 @@
...
+						this.debounce('liveSearchJob', this._liveSearch, 750);
 					}
 				}
 			},
 
 			_onSearchInputKeyPressed: function(e) {
 				switch (e.keyCode) {
 					case this._keyCodes.ENTER:
-						this._advancedSearch(this._searchString);
+						this._updateTiles = true;
```
ryantmer: Could just use `this._onSearchButtonClick` here

d2l-search-widget-custom.html
```diff
@@ -356,27 +392,27 @@
 					this.$$('iron-pages').select('recent-searches-page');
 				} else {
 					this.$.dropdown.close();
+
```
ryantmer: 👾

d2l-search-listbox.html
```diff
@@ -149,7 +149,7 @@
 				if (e.detail.event === 'keydown' && this.keyboardEventMatchesKeys(e, 'space')) {
 					e.preventDefault();
 					if (this.focusedItem && !this.focusedItem.hasAttribute('disabled')) {
-						this.fire('iron-select', {item: this.focusedItem});
+						this.fire('iron-activate', {item: this.focusedItem});
```
ryantmer: Reading the docs for this, why the change? Can't figure out what the difference would be in this case 😢

d2l-my-courses.html
ryantmer: Push down? Like it isn't set in `d2l-all-courses`?

d2l-my-courses.html
```diff
@@ -250,36 +251,23 @@
...
+						var enrollmentId = this.getEntityIdentifier(enrollment);
 						if (enrollment.hasClass('pinned')) {
-							var enrollmentId = this.getEntityIdentifier(enrollment);
 							if (!this._pinnedCoursesMap.hasOwnProperty(enrollmentId)) {
 								newPinnedEnrollments.push(enrollment);
 								this._pinnedCoursesMap[enrollmentId] = true;
 							}
 						} else {
-							newUnpinnedEnrollments.push(enrollment);
+							if (!this._unpinnedCoursesMap.hasOwnProperty(enrollmentId)) {
```
ryantmer: Can just `!this._unpinnedCoursesMap[enrollmentId]`, no? (Same, but easier to read imo)

d2l-my-courses.html
```diff
@@ -197,27 +187,38 @@
...
-				'simple-overlay-closed': '_onSimpleOverlayClosed'
+				'simple-overlay-closed': '_onSimpleOverlayClosed',
+				'd2l-root-enrollment-search': '_rootEnrollmentRequest'
 			},
 			observers: [
 				'_updateEnrollmentAlerts(_hasEnrollments, _hasPinnedEnrollments)'
 			],
 			ready: function() {
-				this._pinnedCoursesMap = {};
+
```
ryantmer: 👾

d2l-my-courses.html
```diff
@@ -110,29 +104,13 @@
 			</button>
 		</div>
 
-		<d2l-simple-overlay
```
ryantmer: I take it the main advantage of moving the overlay inside is that the `iron-scroll-threshold` can now be handled in there? The downside would be that all courses can now only ever be shown in an overlay (probably not a big deal, granted).

(If so, we should do the same with the image overlay perhaps)

d2l-my-courses.html
```diff
@@ -110,29 +104,13 @@
...
-			<d2l-loading-spinner
-				id="lazyLoadSpinner"
-				class="d2l-my-courses-hidden"
-				size="100">
-			</d2l-loading-spinner>
-		</d2l-simple-overlay>
+		<d2l-all-courses
+			pinned-enrollments="{{pinnedEnrollments}}"
+			unpinned-enrollments="{{unpinnedEnrollments}}"
+			my-enrollments-entity="[[_lastEnrollmentsSearchResponse]]"
```
ryantmer: Are both of these needed if they're the same anyway?

d2l-course-tile.html
```diff
@@ -289,6 +288,15 @@
 						break;
 				}
 			},
+			_enrollmentChanged: function() {
```
ryantmer: Is there a need to observe this property? We only ever actually do the call when `delayLoad === true`, so doesn't seem like there's a need to potentially trigger the request on both things?

d2l-all-courses.html
```diff
@@ -277,9 +564,30 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				var itemCount = Math.max(this.pinnedEnrollments.length, this.unpinnedEnrollments.length);
 				return itemCount;
 			},
+			_setFilteredPinnedEnrollments: function() {
+				this.filteredPinnedEnrollments = [];
+				this.filteredPinnedEnrollments = this.pinnedEnrollments;
+			},
+			_setFilteredUnpinnedEnrollments: function() {
+				this.filteredUnpinnedEnrollments = [];
```
ryantmer: Does using `this.set` solve the problem that causes the need for this?

d2l-all-courses.html
```diff
@@ -205,6 +238,103 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 		Polymer({
 			is: 'd2l-all-courses',
 			properties: {
+				// URL constructed to fetch a user's enrollments with the enrollments search Action
+				_enrollmentsSearchUrl: String,
+					// Set of pinned enrollment Entities
+				pinnedEnrollments: {
+					type: Array,
+					value: function() {
```
ryantmer: There are some performance concerns with having observers on properties with default values - any way we can avoid doing it?

d2l-search-listbox.html
ctwomey1: We're using select for focus on the lists.  And it seemed to be more sticky, activate just fires an event on a list item without imposing state (which i don't need here)

d2l-my-courses.html
ctwomey1: So my courses handles initial state, then passes state back down to all courses.  Which all courses then has to handle varying search results. However when you load the all courses widget you need to return to the starting state.  Dirty checking in polymer doesn't always work, and the easiest way is to push the new value again.

d2l-my-courses.html
ctwomey1: I had tried it at one point and it didn't seem to work reliably

d2l-my-courses.html
ctwomey1: So my take on this whole thing was, my courses shouldn't know implementation details of all-courses.  Moving the overlay inside gives that responsibility to the view itself instead of my-courses managing it.  

Overall i wanted all-courses to just be an element we drop on the page that opens an overlay.  If this was a generic component i would create a new polymer element so it could be used outside of an overlay though...

d2l-search-widget-custom.html
ctwomey1: It should, good catch.  Semantically onSearchButtonClick isn't the right thing to call it so i'll rename that

d2l-course-tile.html
ctwomey1: If you don't observe the property things don't update when you change them.  

also it calls when delayLoad === false (it reads **!**this.delayLoad).  Perhaps we don't need the delayload property

d2l-my-courses.html
ryantmer: Yeah, I see what you mean. I'm thinking the passing state isn't a great idea all in all - it saves us re-fetching at most the number of pinned courses anyway. Future change though.

d2l-search-widget-custom.html
ryantmer: Haha, yeah for sure :P

d2l-search-listbox.html
ryantmer: Gotcha.

d2l-course-tile.html
ryantmer: Removing `delayLoad` would require a bit more change, since it saves us a bunch of network calls if the user doesn't open the all courses overlay. That said, if we fully decouple `d2l-my-courses` and `d2l-all-courses` then yeah, it could be removed.

I was thinking the property was unnecessary since we were only previously triggering the load by changing `delayLoad` (i.e. didn't matter if the enrollment changed, we ignore that change until we `.load()` anyway), but I'm realizing that that may no longer be the case what with the searching and whatnot.

d2l-search-widget-custom.html
```diff
@@ -335,16 +342,32 @@
 				}
 			},
 
-			_advancedSearch: function(searchString) {
-				if (searchString) {
-					var queryString = this._advancedSearchUrl + searchString;
-					this._addSearchToHistory(searchString);
-					window.location.href = queryString;
+			//
```
ryantmer: This is my favourite comment 😜

d2l-all-courses.html
```diff
@@ -113,11 +114,36 @@
...
+
+
+		<iron-scroll-threshold
+			id="all-courses-scroll-threshold"
+			on-lower-threshold="_loadNextEnrollmentsPage">
+		</iron-scroll-threshold>
 
 		<div class="search-and-filter">
 			<d2l-search-widget-custom
-				hidden$="[[!_hasEnrollments]]"
```
ryantmer: Hahaha, I'm laughing at how dumb my question is now xD Yeah nvm, that makes no sense at all.

#ForestFromTheTrees

d2l-all-courses.html
```diff
@@ -205,6 +237,103 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+				},
+				// Object containing the IDs of previously loaded unpinned enrollments, to avoid duplicates
+				_unpinnedCoursesMap: {
+					type: Object,
+					value: function() {
+						return {};
+					}
+				},
+				// If enabled, the component will use the pinned/unpinned enrollment queues
+				// In the case of `d2l-all-courses`, the enrollments are switching lists, so this is required.
```
ctwomey1: Yeah, there's some dead code being introduced here.  If you'd like me to remove it now it's probably for the best.  But really I wanted the behavior to be correct before i went cowboy on this

ryantmer: Hm, yeah might as well. Otherwise too easy to forget :( COMMENCE COWBOYING.

ryantmer:
Testing notes!

- Live search works as expected in the text field. As you type, new characters will update the dropdown "suggestions" list, after a delay thanks to the debounce. This seems to have solved the problem where we ended up with >5 entries in that list too, many :+1:s for that!
- When enter is pressed, the course tiles update to display only the courses that match the search term
- When the X is clicked in the search bar, the search reoccurs and goes back to showing all courses
- Course tiles are still divided into pinned and unpinned when searching
- Searching obeys filtering and sorting (mostly, see below)

Things that should probably be addressed in this PR (unless they prove to be fairly large, in which case new PRs!):

- Searching with an empty search term doesn't obey the filter/sort - I suspect the "reset" code path is just not paying attention to `parentOrganizations`
- The AC for this story currently says the search box dropdown should display only recent results (i.e. no live search dropdown)... I suspect this is out of date, right? (If so, @ctwomey1 can you remove it from the US for clarity?)

Things that I noticed that are either pre-existing behaviours that were brought to my attention here, or further changes (either way, future PRs):

- The only way to clear the search is clicking the X - deleting the search term then hitting Enter doesn't clear it (I would argue that clearing the search field should immediately reset the search, regardless of Enter anyway, but meh future discussion)
- Recent/suggested dropdown opens when any part of the search is focused, including the left-hand button and the X. I think it would make more sense if it only showed up when you're focused on the text editor part of the search.
- Recent/suggested dropdown doesn't go away when you hit Enter - I think it should hide itself when the search is done. Not sure if we should do this by auto-focusing the first search result or what, but worth thinking about.
- At least the empty state should be cached somehow - it's a pretty expensive thing to do every time the user hits the X, so would be good to keep it around somehow.

PR #242, opened 2016-12-19
Dbatiste/update deps

dbatiste:
This PR updates deps for perf opts in d2l-icons, d2l-dropdown.


CodeBaboon:
Code LGTM, I will give this a quick smoke test and report back.

CodeBaboon:
d2l-button-group and d2l-navigation also need to get updated to the 4.0.0 version of d2l-dropdown before this should be merged/released, but otherwise all good.

PR #243, opened 2016-12-20
adding configuration value to my-courses for showing course code

ctwomey1:
Fairly small change to enable showing of course code.

If we start doing more config we should really consider using something like https://github.com/PolymerElements/iron-meta but for the time being i left it as a variable passed down.

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -122,6 +122,10 @@
 				filter: none;
 			}
 		}
+		.course-code-text {
+			display: block;
+			font-size: 14px;
+		}
 		.course-text {
 			width: 100%;
 			position: relative;
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -125,6 +125,7 @@
 		.course-code-text {
 			display: block;
 			font-size: 14px;
+			text-transform: uppercase;
 		}
 		.course-text {
 			width: 100%;
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -67,7 +67,11 @@ the user has in that organization - student, teacher, TA, etc.
 						<d2l-course-image aria-hidden="true" image="[[_image]]" sizes="[[tileSizes]]" type="tile"></d2l-course-image>
 					</div>
 				</div>
-				<div class="course-text d2l-heading-4">{{_organization.properties.name}}</div>
+				<div class="course-text d2l-heading-4">{{_organization.properties.name}}
+					<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]" restamp="true">
+						<span class="course-code-text">{{_organization.properties.code}}</span>
+					</template>
+				</div>
 			</a>
 
 			<div class="hover-menu no-tap-interaction">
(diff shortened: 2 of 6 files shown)


test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -122,7 +122,21 @@ describe('<d2l-course-tile>', function() {
...
+		});
+
+		it('should show the course code if configured', function() {
+			widget.showCourseCode = true;
+			widget.$$('#courseCodeTemplate').render();
+			var courseCode = widget.$$('.course-code-text');
+			expect(courseCode.innerText).to.equal(organizationEntity.properties.code);
+		});
+
+		it('should not show the course code if not configured', function() {
```
CodeBaboon: This test name feels off from the implementation. I would expect 'not configured' to basically mean 'no one ever set it', which would mean not including the `widget.showCourseCode = false;` line (and so it becomes an expectation of the default behavior). Then it might be valuable to additionally have another test with this exact code that ensures it properly respects the false setting when set.

d2l-course-tile-styles.html
```diff
@@ -122,6 +122,10 @@
 				filter: none;
 			}
 		}
+		.course-code-text {
+			display: block;
+			font-size: 14px;
```
CodeBaboon: Needs more `text-transform: uppercase;`, per US77133

CodeBaboon: Also supposed to be `font-size: 16px` per the story and the zepplin.

CodeBaboon: Probably should talk to Bob but there seems to be too much space between the name and the code. Maybe `line-height: 20px` or something (computed the value is 28px, the zepplin shows 30px actually but it is also aligning the course code text right to the top of the span).

ctwomey1: bob mentioned that he liked 14px better, despite it not being reflected in the story

d2l-course-tile.html
```diff
@@ -67,7 +67,11 @@
 						<d2l-course-image aria-hidden="true" image="[[_image]]" sizes="[[tileSizes]]" type="tile"></d2l-course-image>
 					</div>
 				</div>
-				<div class="course-text d2l-heading-4">{{_organization.properties.name}}</div>
+				<div class="course-text d2l-heading-4">{{_organization.properties.name}}
+					<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]" restamp="true">
```
ryantmer: Probably not actually a concern, like, whatsoever (since the course tiles are only rendered once, pretty sure), but what's the rationale behind the `restamp` here? That would destroy and recreate the element, which is a performance hit.

d2l-course-tile-styles.html
CodeBaboon: Cool, please have Bob (or yourself) update the story to reflect the change in specs.

d2l-course-tile.html
ctwomey1: Because this is a global setting i don't anticipate the performance hit from restamp to be significant.  show course code would come in at render time and never change during the life time of a course tile. 

I have no great rational for having restamp=true other than elements that are configured to not be there should stop being there.

CodeBaboon: I removed the `restamp="true"` locally and everything seems to work fine on both Chrome and Safari, fwiw.

PR #244, opened 2016-12-20
Add render-content (needed by new d2l-dropdown version)

ryantmer:
Need to render the dropdown contents manually - the newest version of d2l-dropdown defers rendering until required.

@dbatiste - We've got some weird open/close logic around this particular dropdown due to using a not-normal opener, so it seems that `__openedChanged` doesn't get called. Is this an okay approach here?

AC/Risk:

- No story for this, twas just found and fixed in like 10 minutes
- On master right now, the filter menu doesn't load the semesters and departments (it has the tabs for them, but no entries). With this change, yay, semesters and departments!
- Risk: not having this would mean the filter menu doesn't work properly 😛

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -137,7 +137,7 @@
 							aria-labelledby="filterText">
 							<d2l-icon icon="d2l-tier1:chevron-down" aria-hidden="true"></d2l-icon>
 						</button>
-						<d2l-dropdown-content id="filterDropdownContent" no-padding min-width="350">
+						<d2l-dropdown-content id="filterDropdownContent" no-padding min-width="350" render-content>
 							<d2l-filter-menu-content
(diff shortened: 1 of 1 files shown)


trevorjonesd2l:
Code looks good I'll give er a test

dbatiste:
Yes, although ideally `render-content` is used sparingly.  An alternate solution is still defer, but call `forceRender` on the content element just before it's needed.

One of the cases the deferred rendering is meant to hit is - the dropdown/menus for each of the course tiles  Hopefully that's still deferred and working (it seemed to be when during my testing)?

ryantmer:
@dbatiste Yep, sure is :) Not sure what was actually going on here, but the menu is only rendered once anyway, so hopefully that's not too bad.

PR #249, opened 2016-12-21
Remote filterText padding

ryantmer:
This was never exceptionally consistent across browsers, but it seems that recent typography changes have made that difference more noticeable. Since we're not using a "normal" dropdown opener for the filter menu, it makes the most sense to try to match it to the "normal" one (i.e. the sort menu). The best way to do this is to not add any padding - this does mean there's not a lot of space between the chevron on text, but at least the gap is the same between the two menus for any given browser.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -95,9 +95,6 @@
 			#filterSection {
 				display: flex;
 			}
-			#filterText {
-				padding-right: 4px;
-			}
 			button[aria-pressed="true"] {
 				color: var(--d2l-color-celestine);
 			}


CodeBaboon:
Is it better to have it the same between browsers, or optimized for the most common browser(s) and live with inconsistencies elsewhere? I think we should probably consult Bob.

ryantmer:
Yeah - asked him about it earlier, awaiting response!

ryantmer:
For clarity though, I still think this is the right way to go - we are kind of forced to make the filter text/chevron match the sort text/chevron, since the latter is set by `d2l-dropdown`. It may be a bit more squished than we want, but it'll at least be consistent not just between the two menus, but elsewhere in the LMS as well.

ryantmer:
Chatted with Bob - the spacing on the filter button is correct, so it seems that the better approach would be to look at maybe increasing the spacing in d2l-dropdown rather than here. Going to close this; it's a pretty low-priority thing anyway.

PR #251, opened 2016-12-22
Fix Clear button in IE11

ryantmer:
.forEach is not defined on a NodeList (which is returned by querySelectorAll) in IE11. So, fix that by avoiding it and use a regular for.

### DIFF
--- a/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
+++ b/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
@@ -224,12 +224,15 @@
 				return this._currentFilters.indexOf(id) > -1;
 			},
 			_clearFilters: function() {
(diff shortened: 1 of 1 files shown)


apalaniuk:
@ryantmer LGTM, though we use the pattern of calling `Array.prototype.forEach.call` on NodeLists to use forEach on them elsewhere. Could also create an inner function that takes an Array/NodeList to remove a bit of duplication, but all seems minor.

ryantmer:
@apalaniuk Yeah, there's an interesting article about that [here](https://toddmotto.com/ditch-the-array-foreach-call-nodelist-hack/) - like many such articles a lot of it is case-specific, or a bit overblown, but I do agree with [Problem 3](https://toddmotto.com/ditch-the-array-foreach-call-nodelist-hack/#problem3-separation-of-concerns) in particular.

PR #252, opened 2016-12-22
Fix issue where enrollment arrays were getting aliased

apalaniuk:
Fixes an issue where arrays were getting aliased, causing some unintended results when pinning/unpinning. Somehow arrays in `d2l-all-courses` and `d2l-my-courses` were getting aliased by Polymer as well, which is fixed by this, but not particularly well understood why it happened..

Fixes an issue where course pin changes in the All Courses view that acted on enrollments that My Courses was unaware of would results in pinning not working. Enrollment entities are now always added to the correct list of pinned/unpinned enrollment entities, and only removed from the other list when it exists.

Enforced one-way data binding on a ton of array properties that we currently only want to use one-way.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -209,8 +209,8 @@
 
 		<d2l-course-tile-grid
 			id="all-courses-pinned"
-			enrollments="{{filteredPinnedEnrollments}}"
-			enrollments-queue="{{filteredPinnedEnrollmentsQueue}}"
+			enrollments="[[filteredPinnedEnrollments]]"
+			enrollments-queue="[[filteredPinnedEnrollmentsQueue]]"
 			delay-load="[[delayLoad]]"
 			tile-sizes="[[_tileSizes]]"
 			show-course-code="[[showCourseCode]]">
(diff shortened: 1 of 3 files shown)


d2l-all-courses.html
```diff
@@ -543,23 +539,21 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				return itemCount;
 			},
 			_setFilteredPinnedEnrollments: function() {
-				this.filteredPinnedEnrollments = [];
```
apalaniuk: @ctwomey1 Live search seems to behave the same after multiple searches/clears with/without this, but wanted to point out that I removed these sorts of assignments in case there's something I'm missing here.

ctwomey1: I think ultimately i was masking the real problem by reseting that filtered pinned enrollments array each time.  This should be fine but i'll test

ctwomey1:
This looks good to me, @ryantmer can you take a look too, i fear i have tunnel vision from looking at this for so long

PR #254, opened 2016-12-23
Update d2l-siren-parser-ui

ryantmer:
Addresses a problem that was causing some test failures due to trying to parse nothing. In the context of the tests, it's fine that we're passing nothing (usually just means something wasn't fully mocked out), but without this update the tests would fail for an effectively-unrelated reason.

Should solve build failures like https://travis-ci.org/Brightspace/d2l-my-courses-ui/builds/186197684. This is actually _slightly_ circumventing the issue, since we could avoid it entirely by mocking out everything in tests.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -38,7 +38,7 @@
     "d2l-offscreen": "^2.2.2",
     "d2l-polymer-behaviors": "<1.0.0",
     "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#^0.2.4",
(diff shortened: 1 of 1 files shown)


ryantmer:
That's... a different error. I'll look into that tomorrow. By which I mean, later today.

ryantmer:
Oh yay... it's a transient test failure that I can't seem to reproduce locally... I guess I'll look into it further if it comes up again? :S

PR #255, opened 2016-12-23
Allow fetching of enrollment entity on external pin/unpin event

apalaniuk:
Fetch unknown enrollment entities on external pin/unpin event (eg, a course is pinned in the Course Selector dropdown, but the enrollment entity for that course hasn't been retrieved by My Courses yet). Otherwise, pinning a course in this sitation would result in nothing happening in My Courses, when it should add the pinned course there as well, in all circumstances.

### DIFF
--- a/d2l-course-management-behavior.html
+++ b/d2l-course-management-behavior.html
@@ -176,18 +176,18 @@
 			_fetchEnrollmentEntity: function(orgUnitId, pinned) {
 				if (!this._fetchEnrollmentAjaxPinned) {
 					this._fetchEnrollmentAjaxPinned = document.createElement('d2l-ajax');
-					this._fetchEnrollmentAjaxPinned.onResponse = 
+					this._fetchEnrollmentAjaxPinned.onResponse =
 						this._onFetchEnrollmentEntityPinnedResponse.bind(this);
 				}
 
 				if (!this._fetchEnrollmentAjaxUnpinned) {
 					this._fetchEnrollmentAjaxUnpinned = document.createElement('d2l-ajax');
-					this._fetchEnrollmentAjaxUnpinned.onResponse = 
+					this._fetchEnrollmentAjaxUnpinned.onResponse =
 						this._onFetchEnrollmentEntityUnpinnedResponse.bind(this);
 				}
 
-				var ajax = pinned 
-					? this._fetchEnrollmentAjaxPinned 
+				var ajax = pinned
+					? this._fetchEnrollmentAjaxPinned
 					: this._fetchEnrollmentAjaxUnpinned;
 
 				ajax.url = this.fetchEnrollmentUrl + orgUnitId;
--- a/d2l-course-management-behavior.html
+++ b/d2l-course-management-behavior.html
@@ -173,50 +173,51 @@
 					entity._actionsByName['pin-course'] = entity.actions[0];
 				}
 			},
-			_pinEnrollment: function(orgUnitId) {
-				var enrollmentId = this._enrollmentIdMap[orgUnitId];
-
-				if (enrollmentId) {
-					this._moveEnrollmentToPinnedList(enrollmentId);
-				} else {
-					this._fetchEnrollmentEntity(orgUnitId, true);
-				}
-			},
-			_unpinEnrollment: function(orgUnitId) {
-				var enrollmentId = this._enrollmentIdMap[orgUnitId];
-
-				if (enrollmentId) {
-					this._moveEnrollmentToUnpinnedList(enrollmentId);
-				} else {
-					this._fetchEnrollmentEntity(orgUnitId, false);
-				}
-			},
 			_fetchEnrollmentEntity: function(orgUnitId, pinned) {
-				var ajax = document.createElement('d2l-ajax');
+				if (!this._fetchEnrollmentAjaxPinned) {
+					this._fetchEnrollmentAjaxPinned = document.createElement('d2l-ajax');
+					this._fetchEnrollmentAjaxPinned.onResponse = 
+						this._onFetchEnrollmentEntityPinnedResponse.bind(this);
+				}
+
+				if (!this._fetchEnrollmentAjaxUnpinned) {
+					this._fetchEnrollmentAjaxUnpinned = document.createElement('d2l-ajax');
+					this._fetchEnrollmentAjaxUnpinned.onResponse = 
+						this._onFetchEnrollmentEntityUnpinnedResponse.bind(this);
+				}
+
+				var ajax = pinned 
+					? this._fetchEnrollmentAjaxPinned 
+					: this._fetchEnrollmentAjaxUnpinned;
 
 				ajax.url = this.fetchEnrollmentUrl + orgUnitId;
 				ajax.method = 'GET';
-				ajax.onResponse = function(response) {
-					if (response.detail.status === 200 || response.detail.status === 304) {
-						// Parse siren entity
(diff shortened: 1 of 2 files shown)


d2l-my-courses.html
```diff
@@ -237,6 +237,7 @@
 					}
 
 					var searchAction = enrollmentsRoot.getActionByName('search-my-enrollments');
+					this.fetchEnrollmentUrl = searchAction.href + '/organizations/';
```
apalaniuk: Fetching an enrollment entity by org unit ID doesn't seem part of the workflow, so I think I have to get this directly in a.. less than good way. If there's a better way than this, that would be awesome..

d2l-course-management-behavior.html
```diff
@@ -174,10 +174,49 @@
...
+				ajax.method = 'GET';
+				ajax.onResponse = function(response) {
+					if (response.detail.status === 200 || response.detail.status === 304) {
+						// Parse siren entity
+						var parser = document.createElement('d2l-siren-parser');
+						var enrollmentEntity = parser.parse(response.detail.xhr.response);
+
+						if (pinned) {
+							// Enforce specific pin state
+							this._setEnrollmentPinData(enrollmentEntity, pinned);
```
apalaniuk: Done due to the fact that the enrollment pin event from other sources may be sent before the pin/unpin event to the LMS may be processed, so if we know the expected action, enforce it.

d2l-course-management-behavior.html
```diff
@@ -174,10 +174,49 @@
...
+
+				if (enrollmentId) {
+					this._moveEnrollmentToPinnedList(enrollmentId);
+				} else {
+					this._fetchEnrollmentEntity(orgUnitId, true);
+				}
 			},
 			_unpinEnrollment: function(orgUnitId) {
-				this._moveEnrollmentToUnpinnedList(this._enrollmentIdMap[orgUnitId]);
+				var enrollmentId = this._enrollmentIdMap[orgUnitId];
```
CodeBaboon: The contents of `_pinEnrollment` and `_unpinEnrollment` are pretty much identical, save for the move-to-list function they call and the `pinned` arg value they pass to `_fetchEnrollmentEntity`. Might be room for a factored out function that just takes those two params?

d2l-course-management-behavior.html
```diff
@@ -174,10 +174,49 @@
...
+
+				if (enrollmentId) {
+					this._moveEnrollmentToUnpinnedList(enrollmentId);
+				} else {
+					this._fetchEnrollmentEntity(orgUnitId, false);
+				}
+			},
+			_fetchEnrollmentEntity: function(orgUnitId, pinned) {
+				var ajax = document.createElement('d2l-ajax');
```
CodeBaboon: Is this going to create a new dom element on every call? Is there a way to re-use?

d2l-course-management-behavior.html
```diff
@@ -174,10 +174,49 @@
...
+
+				ajax.url = this.fetchEnrollmentUrl + orgUnitId;
+				ajax.method = 'GET';
+				ajax.onResponse = function(response) {
+					if (response.detail.status === 200 || response.detail.status === 304) {
+						// Parse siren entity
+						var parser = document.createElement('d2l-siren-parser');
```
CodeBaboon: Same thing here, could we create a single parser that we re-use instead of a new one on every response?

d2l-course-management-behavior.html
apalaniuk: The only concert I have with that is if there are several events received in a short period of time.

d2l-course-management-behavior.html
apalaniuk: Yeah, I think this makes sense in any scenario. For some reason I thought that we had reverted some cases where we put a parser on the element, but I think I was mistaken.

apalaniuk: I'll probably actually leave this out until after this release - these definitely make sense, and I think a second pass is required, but may make sense to not rush it.

d2l-course-management-behavior.html
CodeBaboon: But the response doesn't rely on the parameters of the element so I think it should be fine? Pretty sure you could just have a `d2l-ajax` element defined in the template and use `auto` on it and then just trigger the request by setting the url of the element inside this function. The `pinned` argument might be the sticking point, but perhaps the answer there is to have two `d2l-ajax` controls, one for pinned and one for un-pinned, though that seems less ideal. I guess it comes down to how often this code will get hit, on average. The way it is written you get N d2l-ajax components, whereas that change would mean you always have 2.

apalaniuk: Yep, looks like you're right (though it seems `auto` might not be suitable for this due to `iron-ajax`s debouncer`).

CodeBaboon: I think the debouncer would only affect us if we tried to use 1 component and attach the pinned value as an argument. If the only attribute changing is the URL then nothing should get debounced.

d2l-course-management-behavior.html
```diff
@@ -173,11 +173,51 @@
...
+				ajax.method = 'GET';
+				ajax.generateRequest();
+			},
+			_onFetchEnrollmentEntityPinnedResponse: function(response) {
+				this._onFetchEnrollmentEntityResponse(response, true);
 			},
-			_unpinEnrollment: function(orgUnitId) {
-				this._moveEnrollmentToUnpinnedList(this._enrollmentIdMap[orgUnitId]);
+			_onFetchEnrollmentEntityUnpinnedResponse: function(response) {
+				this._onFetchEnrollmentEntityResponse(response, false);
```
apalaniuk: I don't think you can do partial application that ignores leading parameters in ES5 without libraries or writing additional helper code..

d2l-course-management-behavior.html
```diff
@@ -173,11 +173,51 @@
...
+			},
+			_onFetchEnrollmentEntityResponse: function(response, pinned) {
+				if (response.detail.status === 200 || response.detail.status === 304) {
+					this._sirenParser = this._sirenParser || document.createElement('d2l-siren-parser');
+					var enrollmentEntity = this._sirenParser.parse(response.detail.xhr.response);
+
+					if (pinned) {
+						// Enforce specific pin state
```
jstefaniuk-d2l: So we enforce pinned = true, but not pinned = false?

apalaniuk: Bah.

jstefaniuk-d2l: It seems like no-matter what there will be a passed in explicit value here

d2l-course-management-behavior.html
jstefaniuk-d2l: Looks like you can do this with `bind`
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Function/bind

```
this._fetchEnrollmentAjaxPinned.onResponse = _onFetchEnrollmentEntityResponse.bind(this, true)
...
_onFetchEnrollmentEntityResponse: function(pinned, response)
```

jstefaniuk-d2l:
Code LGTM

cgalvind2l:
Tried it out. It does the thing! #tested 

More specifically, I pinned courses in the all courses dropdown before they were ever loaded by the my courses widget and they showed up immediately in pinned courses. (Also verified that this does not happen without these changes.)

Are there any other areas of risk?

cgalvind2l:
Actually, I'm seeing this work in Firefox but not Chrome.

PR #256, opened 2017-01-03
US80997 - Don't render all-courses initially

ryantmer:
No point registering all these WCs (which has a negative performance impact in non-Chrome browsers) if we don't need them. This also just significantly reduces the number of WCs being rendered in general - down from >1000 on the homepage to 267, according to `D2L.Performance.getComponentCounts`.

The disadvantage of this approach is that it now takes a moment for the overlay to load the first time. While this is an improvement overall, going to see if I can find a way to make it a bit less obvious.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -216,12 +216,6 @@
 			_createAllCourses: function() {
 				if (!this._allCoursesCreated) {
 					var allCourses = document.createElement('d2l-all-courses');
-					allCourses.pinnedEnrollments = this.pinnedEnrollments;
-					allCourses.unpinnedEnrollments = this.unpinnedEnrollments;
-					allCourses.myEnrollmentsEntity = this._lastEnrollmentsSearchResponse;
-					allCourses.lastEnrollmentsSearchResponse = this._lastEnrollmentsSearchResponse;
-					allCourses.locale = this.locale;
-					allCourses.showCourseCode = this.showCourseCode;
 					this.$.allCoursesPlaceholder.appendChild(allCourses);
 					this._allCoursesCreated = true;
 				}
@@ -312,9 +306,24 @@
 			_openAllCoursesView: function(e) {
 				this._createAllCourses();
 
+				var allCourses = this.$$('d2l-all-courses');
+				allCourses.open();
+
 				e.preventDefault();
 				e.stopPropagation();
-				this.$$('d2l-all-courses').load();
+
+				this.async(function() {
+					// The opening of the overlay lags slightly if we try to do this while it
+					// is opening - do it after the animation is finished. Animation takes
+					// 500ms, but some lag if we wait exactly that long, so +100ms for safety)
+					allCourses.pinnedEnrollments = this.pinnedEnrollments;
+					allCourses.unpinnedEnrollments = this.unpinnedEnrollments;
+					allCourses.myEnrollmentsEntity = this._lastEnrollmentsSearchResponse;
+					allCourses.lastEnrollmentsSearchResponse = this._lastEnrollmentsSearchResponse;
+					allCourses.locale = this.locale;
+					allCourses.showCourseCode = this.showCourseCode;
+					this.$$('d2l-all-courses').load();
+				}.bind(this), 600);
 			},
 			_refreshPage: function() {
 				document.location.reload(true);
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -53,8 +53,6 @@
 				margin-bottom: 20px;
 				clear: both;
 			}
-
-
 		</style>
 
 		<d2l-ajax
@@ -105,14 +103,8 @@
 			</button>
 		</div>
 
-		<d2l-all-courses
-			pinned-enrollments="[[pinnedEnrollments]]"
-			unpinned-enrollments="[[unpinnedEnrollments]]"
-			my-enrollments-entity="[[_lastEnrollmentsSearchResponse]]"
-			lastEnrollmentsSearchResponse="[[_lastEnrollmentsSearchResponse]]"
-			locale="[[locale]]"
-      		show-course-code="[[showCourseCode]]">
-		</d2l-all-courses>
+		<div id="allCoursesPlaceholder">
+		</div>
 
 		<d2l-simple-overlay
 			id="basic-image-selector-overlay"
@@ -147,6 +139,8 @@
 				imageCatalogLocation: String,
 				// hashed userid to send to telemetry
 				userId: String,
+				// Configuration value passed in to toggle course code
+				showCourseCode: Boolean,
 				// hashed tenantid to send to telemetry
 				tenantId: String,
 				// URL to send telemetry to
@@ -213,6 +207,7 @@
 			detached: function() {
 				document.body.removeEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
 			},
+			_allCoursesCreated: false,
 			getCourseTileItemCount: function() {
 				return this.pinnedEnrollments.length;
 			},
@@ -300,15 +295,21 @@
 				this.$['basic-image-selector-overlay'].open();
 			},
 			_openAllCoursesView: function(e) {
+				if (!this._allCoursesCreated) {
+					var allCourses = document.createElement('d2l-all-courses');
+					allCourses.pinnedEnrollments = this.pinnedEnrollments;
+					allCourses.unpinnedEnrollments = this.unpinnedEnrollments;
+					allCourses.myEnrollmentsEntity = this._lastEnrollmentsSearchResponse;
+					allCourses.lastEnrollmentsSearchResponse = this._lastEnrollmentsSearchResponse;
+					allCourses.locale = this.locale;
+					allCourses.showCourseCode = this.showCourseCode;
+					this.$.allCoursesPlaceholder.appendChild(allCourses);
+					this._allCoursesCreated = true;
+				}
+
 				e.preventDefault();
 				e.stopPropagation();
-				this.$$('d2l-all-courses').lastEnrollmentsSearchResponse = this._lastEnrollmentsSearchResponse;
 				this.$$('d2l-all-courses').load();
-
-				setTimeout(function() {
-					// Slight delay to allow for overlay to get a DOM width before triggering the recalculation
-					this.fire('recalculate-columns');
-				}.bind(this), 50);
 			},
 			_refreshPage: function() {
 				document.location.reload(true);
@@ -326,7 +327,9 @@
 						}.bind(this), 1000); // delay until the tile fail icon animation begins to kick in (1 sec delay)
 					}
 				}
-				this.$$('d2l-all-courses').setCourseImage(e);
+				if (this._allCoursesCreated) {
+					this.$$('d2l-all-courses').setCourseImage(e);
+				}
 				this.$$('d2l-course-tile-grid').setCourseImage(e);
 			},
 			_changeImageLoadMore: function() {
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -98,7 +98,9 @@
 			<button
 				class="all-courses-button"
 				hidden$="{{!_hasEnrollments}}"
-				on-tap="_openAllCoursesView">
+				on-tap="_openAllCoursesView"
+				on-mouseover="_createAllCourses"
+				on-focus="_createAllCourses">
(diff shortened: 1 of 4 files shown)


d2l-my-courses.html
```diff
@@ -299,15 +294,21 @@
 				this.$['basic-image-selector-overlay'].open();
 			},
 			_openAllCoursesView: function(e) {
+				if (!this._allCoursesCreated) {
+					var allCourses = document.createElement('d2l-all-courses');
+					allCourses.pinnedEnrollments = this.pinnedEnrollments;
```
apalaniuk: Since we were originally using one-way data binding, should we explicitly clone objects rather than alias? I haven't looked too hard, and All Courses seems fairly resilient due to it cloning the enrollments arrays, but my Spidey sense is tingling..

ryantmer: Hmm, yep I think your Spidey sense has a point. Cloning is probably the more-explicitly-safe thing to do, I'll update it.

ryantmer: Er, actually - missed your comment about it being resilient, I see what you're saying. I think this is okay, since `d2l-all-courses` is doing the `slice` itself?

ctwomey1:
Few thoughts that have popped into my head while mulling this over.

The number of web components dropping from 1000 to 267 seems like maybe the components could be streamlined somehow?  Is there any chance we can get a list of what web components are being created?  We know that we are creating siren parser too often, or perhaps the tile grids are creating a new component on every change?  Something to think about.

A possible solution to having the simple overlay load fully populated would be to use the polymer [async  ](https://www.polymer-project.org/1.0/docs/api/Polymer.Base#method-async) method to create the overlay in the background.  I haven't tested it, but doing it this way should let you register the elements after mycourses has fired it's attached callback

ryantmer:
The big drop largely comes from the extra course tiles we're rendering (the 25 within all-courses) - each one adds a few `d2l-ajax`es, some icons, etc. Dave B added this information under `D2L.Performance.getComponentCounts()`.

I'll give async a shot too - I suspect its performance will come somewhere in between using a dom-if template and using `createElement`, but it might be a cleaner approach.

CodeBaboon:
I think this PR might go a long way to solving the multiple-image-download issue that Trevor and I were looking into yesterday. I'm going to play around with this and see.

ryantmer:
Were you seeing it occur only on pinned courses, or all courses?

CodeBaboon:
The issue is because all-courses doesn't correctly calculate the `sizes` value on initial load (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-all-courses.html#L641). It incorrectly calculates that 1 column is visible and so pulls down the image corresponding to 100vw.

Simply removing that calculation on Ready didn't quite work because for some reason the logic here (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-all-courses.html#L491) isn't being respected and instead the default size values that are set in the d2l-course-image get used (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/components/d2l-course-image.html#L97). Those sizes don't match up with my-courses (it uses the same ones that all-courses tries to use in that second link of mine), leading to situations where they still pulled different image sizes.

As I type all this I think I realize though that this change, even if the calc works, still won't be perfect, and that I still want to do what Trevor and I originally came up with yesterday. So maybe ignore all this, I guess. Blargh.

ryantmer:
@ctwomey1 Hah nice, looks like `async` is comparable in speed! OTOH, it does mean we're still registering and rendering the whole overlay, even if the user never clicks the "View All Courses" button, which... seems wasteful? Idk. Would be nice to just do it as-required, but if I can't get that to work in a nice way, then yeah `async` seems like the way to go!

@CodeBaboon Oof. Yeah I don't that that will be helped here, necessarily - it might for the course tiles in the overlay, but I don't think it would for the initially-loaded ones :< Maybe things are happening before the elements get a dom width? I know that was happening elsewhere, so we had to `setTimeout` some things.

ryantmer:
Wait, actually - `async` is slightly _faster_ than doing it in the open handler...? I have no idea how that works, but whatever! Even if it's within the margin of error, it's a cleaner way to do it either way. I shall update!

CodeBaboon:
@ryantmer Yeah I think that's part/all of it. I think ideally the all-courses page would use the same image as the my-courses page for all of the pinned tiles, even if that means the image isn't 100% optimal, because better to download one image than two. Unpinned tiles don't load until the overlay is opened and at that point the column calculations all work great, so for any of those images we would download the optimal size (and that might include downloading a second size for an image that is used by one of the pinned courses, but that seems like an edge case that isn't going to be a large issue).

d2l-my-courses.html
apalaniuk: Yeah, exactly. I think we'll also have to explicitly clone those properties on every load, since `d2l-all-courses` splices on every load also.

dbatiste:
Regarding async: if the timelines are compared, you may be able to determine why the performance is better.  Also, is it better in all browsers?  Do we expect showing the overlay to be common practice, and if so, what's the typical cost (i.e. not D2L admin)?  I think we always need to consider progressively loading and rendering parts of the application.

I am concerned that async could still be too early - competing to render other components required to display the page.  If the typical cost is significant (for a crappy phone on a poor network), my first choice would be to defer rendering the overlay entirely - avoiding the rendering cost and network requests.

Another possible approach: perhaps a user event (such as focus, hover, touch start) provides enough hint to load a little in advance of them actually clicking.  Or perhaps the overlay can be initially rendered, but rendering each tile is done progressively when displaying the overlay.  These approaches perhaps still do not provide the best experience when showing the overlay, but the cost is deferred from the initial page load entirely, and we can probably deliver a good experience.

If loading and rendering are indeed needed at page load time: a couple of other options that may help are: [window.requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) or `Polymer.RenderStatus.afterNextRender`.  However, these may be too early also, and they don't avoid the potentially unnecessary network requests.  Note: `requestIdleCallback` is currently only supported by Chrome (coming in FF).

ctwomey1:
I think this [observer](https://github.com/Brightspace/d2l-my-courses-ui/blob/rmahler/intelligently-aggressively-lazy/d2l-all-courses.html#L251) is the problem @CodeBaboon  might be seeing with double loads.  Polymer setting an enrollment for the first time will cause them to render, and then the observer does it. plus we are setting things during load.  Too many times!

The observer can be removed, if we change the event on clear to do an all courses search

CodeBaboon:
@ctwomey1 It isn't about double loads, it's all about the img srcset and sizes values. I'm stuck in the data generator muck right now but when I get back to this stuff I'll get a fix in. Should be totally unrelated to these changes I suspect.

d2l-my-courses.html
ryantmer: Yup haha, I hadn't realized that either, hence my conflicting comments :P

ryantmer:
Yeah, I suspect the overlay is not actually needed for the majority of page loads, hence my initial approach. The real downside to it was the sub-par experience, but your suggesting of creating the element on focus/hover instead might give juuust the right amount of lead time that we need. It wasn't necessarily _bad_ the way I was doing it initially, it just took longer than previously (as would be expected).

Another possibility would be to rearrange things a bit so that the overlay opens, but it `createElement`s its own content. This would at least give us the option of click button -> overlay opens immediately (but with loading spinner or something) -> content renders on-demand. The issue I was seeing with delaying everything was that in IE11 in particular, you would click the button, then it seems like nothing is happening for a good couple of seconds, which is... not particularly great. But, I'll explore the `on-focus` idea, as well as looking into ways I can make that "creation time" less noticeable to the user, while still only creating things if they are, in fact needed.

ryantmer:
@dbatiste Ah yep, good call. Using `async`/100ms in IE11 doesn't help at all, which I guess makes sense. In Chrome it's noticeable since I'd imagine the `createElement` and associated registrations aren't fighting with other stuff since they're probably mostly already done. But, doesn't look to be the case in IE11. Hokay, back to some variant of the initial approach!

Dammit IE...

ryantmer:
Hokay, tried the only-`createElement`-on-hover/focus that @dbatiste suggested and that looks like a pretty good option. As expected, it reduces `page.display`, down by 100-200ms in Chrome. In IE11, `page.display` is pretty variable it seems, but it's definitely lower (seemingly about 500ms less) with this change. More noticeable, however, is that the My Courses widget renders quite quickly after the page appears, whereas on master right now it takes a solid 1-2 seconds to show up.

Creating the element on hover/focus is pretty unnoticeable; because I'm aware it's there, opening the overlay the first time does take a fraction of a second longer, I think it's worth it considering the rest of the impact.

dbatiste:
If the rendering is slow for some reason, do we need to display a spinner, or do we need the overlay immediately, and progressively load the contents?  What's the mobile experience look like?  (I suspect the hover/focus strategy isn't going to help that case...?)

ryantmer:
In the mobile case we're just going to end up deferring it until the tap, yeah. I'm currently looking at the option of opening the overlay immediately, then loading the contents. It's possible that the visual fade is actually long enough to hide any other delays anyway, which would be great.

ryantmer:
Yep, looks like opening first and then loading contents is even better. I think the visual fade is a bit of a saviour here, as it gives 100ms or so to do stuff in the background without it looking like any lag is occurring - and that seems to be enough, on Chrome/FF/IE11 at least, to load things up and get the first set of results populated (or at least mostly populated) by the time the overlay has finished its fade in.

CodeBaboon:
On Mac Chrome I'm losing the animation on the first display of all-courses with this change. Everything just appears suddenly. I'll see if I can get a video.

CodeBaboon:
Video of the loss of animation frames: https://drive.google.com/open?id=0B8klYrVFXiq3T1hrMjNrRTd3ZGc

CodeBaboon:
Of note: Safari looks fine. Firefox is better than Chrome in that you generally can get a sense of the slide, but it is still skipping most of the frames.

CodeBaboon:
Final note: The 'desktop' width animation, where it sort of zooms down and in as opposed to the slide, seems to work mostly fine on all browsers.

ryantmer:
Hmm, I see what you're saying. Didn't figure this would affect the mobile one all that much... hokay, I'll look back into it, thanks!

ryantmer:
After much poking (and some minor house repair), switch to what should have been obvious in the first place - let the animation finish, then load the things. This means there's a loading spinner the first time you open it, but it's quick to go away and feels quite smooth in fact. Like, to the point where I'm slapping myself for not just doing this in the first place...

Anywho, ready for actual review/test now, since I think this a good approach?

jstefaniuk-d2l:
Code LGTM

d2l-my-courses.html
```diff
@@ -300,15 +304,26 @@
...
 
-				setTimeout(function() {
-					// Slight delay to allow for overlay to get a DOM width before triggering the recalculation
-					this.fire('recalculate-columns');
-				}.bind(this), 50);
+				this.async(function() {
+					// The opening of the overlay lags slightly if we try to do this while it
+					// is opening - do it after the animation is finished. Animation takes
+					// 500ms, but some lag if we wait exactly that long, so +100ms for safety)
+					allCourses.pinnedEnrollments = this.pinnedEnrollments;
```
apalaniuk: Are we not concerned about aliasing the arrays?

ryantmer: The observer for `pinnedEnrollments` in all courses does the splice, so we're good! (Assuming that is what you mean)

apalaniuk: @ryantmer Yeah, and I remember bringing this up before, still just feels a bit odd as the original expectation was one-way data binding. That said, there isn't a huge concern right now.

apalaniuk:
Tests:

- [ ] Opening All Courses view behavior matches current behavior (minus performance improvements/spinning loader); ie. Courses are the same in both instances
- [ ] (Un)pin course behavior matches current behavior; ie. Courses can be pinned/unpinned correctly
- [ ] Close/reopen All Courses view behavior matches current behavior; ie. My Courses is not adversely affected, and the state of All Courses is the same in both cases

d2l-my-courses.html
ryantmer: Fair - I also think that the issue Steve mentioned in Flowdock yesterday with pinned courses not showing up on search might require some rejiggering of our approach to that data anyway =/

PR #258, opened 2017-01-03
Show loading spinner in filter menu

ryantmer:
Until the first two requests have come back, the menu is empty, so show a loading spinner until they come back.

### DIFF
--- a/d2l-filter-menu-content/d2l-filter-menu-content.html
+++ b/d2l-filter-menu-content/d2l-filter-menu-content.html
@@ -1,6 +1,7 @@
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="../../d2l-colors/d2l-colors.html">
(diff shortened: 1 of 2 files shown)


d2l-filter-menu-content/d2l-filter-menu-content.html
```diff
@@ -174,6 +182,10 @@
 						return;
 					}
 
+					this.toggleClass('d2l-filter-menu-content-hidden', true, this.$$('d2l-loading-spinner'));
+					this.toggleClass('d2l-filter-menu-content-hidden', false, this.$.contentView);
+					this.toggleClass('d2l-filter-menu-content-hidden', false, this.$.filterHeader);
+
```
ctwomey1: This is great!  One thing to note, this spinner will just keep on spinning if there's an error. Do we have a story that deals with that?

ryantmer: Yeah, fair point. I'll change it to that behaviour for now, thanks!

PR #261, opened 2017-01-05
US77128 - visual cue if the course hasn't started yet

jstefaniuk-d2l:
https://rally1.rallydev.com/#/51191528503d/detail/userstory/63307413980

- [x] Langterms
- [x] Make it accessible

Needs: https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/4221/overview

image


ryantmer:
Your estimation is correct! Yeah, course tile is a link right now, so that would have to change.

trevorjonesd2l:
I was looking at this yesterday and I don't think that in normal conditions courses will show up before the start date - the `CanAccess` of the `MyOrgUnitInfo` is set to false under these conditions:
```c#
hasDateRestrictions
&& !role.AccessFutureOrgUnits
&& orgUnitStartDate.HasValue
&& orgUnitStartDate.Value > UtcDateTime.Now
```

jstefaniuk-d2l:
For teachers they should right now, otherwise I'm going to change it in the next PR to work for students

jstefaniuk-d2l:
Or d2lsupport at least, in any case

jstefaniuk-d2l:
So those things that I was saying before about showing it for users turned out to not be right, although I'm going to look to see if you can get a course without a href (Like I remember from the University of Manitoba), since our widget would break in that case I think

PR #262, opened 2017-01-06
Fix filter dropdown behavior

apalaniuk:
Fixes DE23756, where clicking on the dropdown arrow for the search filter menu won't size the dropdown correctly or initialize the dropdown contents, while clicking on the "Filter" text would.

Eg, when there are enough departments and/or semesters to show the tabbed view, it should look like this:

filter-dropdown-tabbed-correct

But, if you click on the the dropdown arrow instead of the "Filter" text, it looks like this:

fitler-dropdown-tabbed-broken

Similarly, the untabbed view should be similar to this:

filter-dropdown-untabbed-correct

But if the user clicks the dropdown arrow instead of the "Filter" text, it looks like this:

filter-dropdown-untabbed-broken

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -512,6 +512,7 @@
 			},
 			attached: function() {
 				this.listen(this.$.sortDropdown, 'd2l-menu-item-change', '_updateSortBy');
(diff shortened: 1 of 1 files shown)


jstefaniuk-d2l:
Code LGTM

PR #264, opened 2017-01-06
Us77127, Us81086, US77128 - add a visual cue if a course is starting or ending

jstefaniuk-d2l:
I did all the stories, and then the design for the first one changed, so now it's all one PR

https://rally1.rallydev.com/#/51191528503d/detail/userstory/63307413793?fdp=true
https://rally1.rallydev.com/#/51191528503d/detail/userstory/63307413980
https://rally1.rallydev.com/#/51191528503d/detail/userstory/85215683864

Acceptance criteria:
- When a course is not yet started, add an overlay to the course tile saying so
- When a course has ended, add an overlay to the course tile saying so
- When a course is started, and has not yet ended, the course tile should be normal
- When the user selects an image, show the image for a couple seconds before showing the overlay again
- The course-tile overlay should also work in the "unpinned" and "pinned" sections of the all courses overlay

- When the user has no access to the course due to the course being out of their date range, and the "d2l.Security.BypassDateCheck" config is set:
    - The course tile should not be selectable
    - The course tile should not be clickable
    - The course image should not expand on mouse hover
    - The user should still be able to access the ellipsis menu


jstefaniuk-d2l:
I think this is ready for code review

It still needs some design review and css tweaking, but that'll be mostly just font sizes/font weight

d2l-course-tile.html
```diff
@@ -64,13 +64,22 @@
 								<d2l-icon class$="[[_iconDetails.className]]" icon="[[_iconDetails.iconName]]"></d2l-icon>
 							</div>
 						</div>
+						<div class="notification-overlay" aria-hidden="true">
```
ryantmer: I _think_ just using the `hidden` attribute with the same binding here might be a better idea. The cost of using the `template` here is probably more than the cost of just writing the divs, and since it's only ever set once anyway, we don't really benefit much from the `dom-if`.

This is based on [this](https://www.polymer-project.org/1.0/docs/devguide/templates#dom-if), specifically the part "_there may be performance benefits to not burdening most of the users with the cost of stamping the elements [that aren't applicable to them], especially if it is relatively heavyweight_", whereas in this case I think the opposite applies - there may be performance benefits to not _not_ stamping the elements, especially if it relatively lightweight.

d2l-course-tile.html
```diff
@@ -176,12 +187,11 @@
...
 				},
-				/*
-				* The JSON string of the telemetry event that will be sent
-				* on the opening of the change image UI
-				* @type {String}
-				*/
-				_telemetryEvent: String
+				// The JSON string of the telemetry event that will be sent on the opening of the change image UI */
+				_telemetryEvent: String,
+				_dateFotmatter: Object,
```
ryantmer: [Typoooooo](https://www.youtube.com/watch?v=H5ud1oB4ei4&feature=youtu.be&t=197)

d2l-course-tile.html
jstefaniuk-d2l: I assumed that the article was referring to the cost of restamping multiple times (although I may have been wrong with that), but I can remove the restamp as it really has no purpose here

d2l-course-tile.html
```diff
@@ -404,13 +419,53 @@
...
+						var tileContainer = this.$$('.tile-container');
+						this.toggleClass('d2l-no-access', true, tileContainer);
+					}
 					this._canChangeCourseImage = this._getCanChangeCourseImage(organization);
 					this._pinnedChanged();
 					this._courseSettingsLabel = this.localize('courseSettings', 'course', this._organization.properties.name);
 				}
 			},
+			_checkDateBounds: function(organization, response) {
+				var serverDate = Date.parse(response.detail.xhr.getResponseHeader('Date'));
```
jstefaniuk-d2l: Date.now is bad because if the user has the wrong date on their computer then we don't want them thinking their course is ended, the server date is more reliable this way

d2l-course-tile-styles.html
```diff
@@ -349,6 +374,47 @@
...
+			.change-image-success .notification-overlay,
+			.change-image-failure .notification-overlay {
+				opacity: 0;
+			}
+		}
+
+		.change-image-loading .notification-overlay,
+		.change-image-success .notification-overlay,
+		.change-image-failure .notification-overlay {
+			filter: opacity(0);
```
jstefaniuk-d2l: Chrome doesn't animate opacity on this element unless you remove the boarder radius on the course tile (But it _will_ animate `filter: opacity`), and ie doesn't support filter

d2l-course-tile.html
```diff
@@ -404,13 +419,53 @@
...
+
+				if (organization.properties.endDate) {
```
jstefaniuk-d2l: cool! :+1:

d2l-course-tile-styles.html
ryantmer: Yaaaaay. Okay fair haha. Gee I sure wish there was, like, a standard for all of this stuff or something...

d2l-course-tile.html
ryantmer: As I understand it, there's two considerations - you are right in that `restamp` only really becomes an issue if `_showOverlay` is going to change its truthiness multiple times, which it won't here (so I guess whether it's true or false doesn't matter, at best removing it is just removing unnecessary code). The other consideration, that I'm referring to above, is the cost of `template` itself, which I suspect would be more than the divs. The advantage of `template` being that we can avoid costly rendering, but in this case I think the rendering cost of the divs is _so cheap_ that the cost of the `template` is actually more than the content that we're deferring rendering of.

This is... somewhat speculative, like I don't know where the line of "things that are worth just rendering" and "things that should be `dom-if`'d" cross, but since this is super simple content, I'd put a fair wager on it being better to just render it, rather than the `template`. (And now, I'm curious to see where those lines *would* cross... but have no idea how one would reliably test that.)

d2l-course-tile.html
ryantmer: Ah, yeah that's a fair point. Server/client time differences I usually only think of in the context of like... millisecond differences, which matters for concurrency but not user interaction. Didn't think of a user just being like "nah, I'm pretty sure it's 1938 right now".

d2l-course-tile.html
jstefaniuk-d2l: Also, just for the sake of context, when I initially put the restamp in, there was like 12 divs in the template: https://github.com/Brightspace/d2l-my-courses-ui/pull/264/commits/d6ab69156211158ef24cb0e8651c895281e65977

ryantmer: Haha, fair enough. Like I said, I don't _really_ know where those lines would cross, but I think if we're just rendering native stuff (i.e. no WCs) then we might be okay to just use `hidden` in general.

(We're probably talking fairly small improvements in general here, but the course tile is our most-used element, so in this limited context it might make a difference?)

localize-behavior.html
```diff
@@ -63,6 +63,10 @@
 								'filtering.departmentMultiple': 'Department ({num})',
 								'filtering.searchDepartments': 'Search Departments',
 								'filtering.searchSemesters': 'Search Semesters',
+								'overlay.courseStarting': 'Course starting',
```
ryantmer: Have these terms been submitted for translations as well?

ryantmer:
Seems legit - going to check out and test

localize-behavior.html
jstefaniuk-d2l: Not yet, do they have to be?

We should add all the terms we need and then submit one final set

ryantmer: Yeah, they should be submitted. I don't know that there's a "minimum size" for submitting a translation request or anything, but I'm not actually sure. Either way, since it's an external blocker it's probably reasonable to do it in another PR (even maybe another US).

ryantmer:
This PR covers three stories, so - testing plan, by story:

## US81086 (d2l.Security.BypassDateCheck)

When `d2l.Security.BypassDateCheck` is true,

[x] A student cannot click the tile to enter the course if it has not started or has ended
[x] An admin can click the tile to enter the course if it has not started or has ended
[x] Image does not expand on hover for a student
[x] Image does expand on hover for admins
[x] Ellipses menu still shows up and works for both admins and students

When `d2l.Security.BypassDateCheck` is false,

[x] Normal behaviour - student cannot see the course tile at all if it has not started or has ended

## US77127 (courses that have ended)

For a course that has ended,

[x] Display a message that the course has ended
[x] Display the date the course ended

## US77128 (courses that haven't yet started)

For a course that has not yet started,

[x] Display a message that the course hasn't started
[x] Display the date the course starts
[x] Changing the course image image hides the overlay, shows the spinner, checkmark, then fades the overlay back in (omfg that's satisfyingly nice)

Testing interactions in:

[x] Chrome
[x] FF
[x] IE11
[ ] Edge
[ ] Safari - need to find somewhere with recent Safari...

Problems:

[ ] There seems to be an issue where unpinning a future/past course will result in the next course in the list erroneously having the overlay in the all courses view
[ ] Overlays don't show up in Edge - or, more accurately they did initially, but then pinning/unpinning a course caused them to disappear and now they won't come back ;_;

PR #265, opened 2017-01-09
Added advanced search url

JoshuaKlassen:
US80239

Renders an advanced search url link under the search bar in the all courses view, that redirects to the legacy advanced search page.

To test:
checkout branch: add-advanced-search-url in repo: le/mangeCourses

Has not been tested in Edge

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -210,7 +210,7 @@
 							</d2l-dropdown-menu>
 						</d2l-dropdown>
 			</div>
-			<div id="advancedSearchLink" class="advanced-search-link">
+			<div class="advanced-search-link">
 				<a is="d2l-link" href="[[advancedSearchUrl]]">{{localize('advancedSearch')}}</a>
 					</div>
 				</div>
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -377,7 +377,10 @@
 					}
 				},
 				// URL that directs to the advanced search page
-				advancedSearchUrl: String
+				advancedSearchUrl: {
+					type: String,
+					observer: '_advancedSearchUrlChanged'
+				}
 			},
 			behaviors: [
 				window.D2L.MyCourses.CourseTileResponsiveGridBehavior,
@@ -537,8 +540,6 @@
 				this._updateTileSizes();
 
 				this._filterText = this.localize('filtering.filter');
-
-				this.toggleClass('d2l-all-courses-hidden', !this.advancedSearchUrl, this.$.advancedSearchLink);
 			},
 			attached: function() {
 				this.listen(this.$.sortDropdown, 'd2l-menu-item-change', '_updateSortBy');
@@ -567,6 +568,9 @@
 			},
 			_setFilteredUnpinnedEnrollments: function() {
 				this.filteredUnpinnedEnrollments = this.unpinnedEnrollments.slice();
+			},
+			_advancedSearchUrlChanged: function() {
+				this.toggleClass('d2l-all-courses-hidden', !this.advancedSearchUrl, this.$.advancedSearchLink);
 			},
 			load: function() {
 				// Load course tile and filter contents. Called from d2l-my-courses.
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -152,67 +152,67 @@
 					on-lower-threshold="_loadNextEnrollmentsPage">
 				</iron-scroll-threshold>
 
-		<div id="search-and-filter">
-			<div class="search-and-filter-row">
-				<d2l-search-widget-custom
-					id="search-widget"
-					my-enrollments-entity="[[myEnrollmentsEntity]]"
-					parent-organizations="[[_parentOrganizations]]"
-					sort-field="[[_sortField]]">
-				</d2l-search-widget-custom>
+				<div id="search-and-filter">
+					<div class="search-and-filter-row">
+						<d2l-search-widget-custom
+							id="search-widget"
+							my-enrollments-entity="[[myEnrollmentsEntity]]"
+							parent-organizations="[[_parentOrganizations]]"
+							sort-field="[[_sortField]]">
+						</d2l-search-widget-custom>
 
-				<div id="filterAndSort" class="d2l-all-courses-hidden">
-					<div id="filterSection" class="d2l-all-courses-hidden">
-						<span id="filterText" class="dropdown-opener-text" on-tap="_toggleFilterDropdown">[[_filterText]]</span>
-						<d2l-dropdown id="filterDropdown">
-							<button
-								id="filterDropdownOpener"
-								class="d2l-dropdown-opener dropdown-button"
-								on-focus="_focusFilterText"
-								on-blur="_blurFilterText"
-								on-mouseenter="_focusFilterText"
-								on-mouseleave="_blurFilterText"
-								aria-labelledby="filterText">
-								<d2l-icon icon="d2l-tier1:chevron-down" aria-hidden="true"></d2l-icon>
-							</button>
-							<d2l-dropdown-content id="filterDropdownContent" no-padding min-width="350" render-content>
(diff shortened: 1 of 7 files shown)


localize-behavior.html
```diff
@@ -29,6 +29,7 @@
 					value: function() {
 						var langs = {
 							'en': {
+								'advancedSearch' : 'Advanced Search',
```
apalaniuk: One thing we were talking about was for getting the existing translations for this term from the LMS and putting them in as part of this PR, but not having gone through the translations process, I'm not sure if that's "correct" - seems like something we'd want to do while we're here though.

d2l-search-widget-custom.html
```diff
@@ -187,6 +201,12 @@
 				</div>
 			</d2l-search-dropdown-content>
 		</d2l-dropdown>
+
+		<template is="dom-if" if="[[advancedSearchUrl]]">
```
ryantmer: Hm, can I actually advocate for this to be outside of `d2l-search-widget-custom`? My rationale is that while it's related to search, it is not functionally connected to it, so it could just as easily be within `d2l-all-courses` itself.

In a supporting-but-not-overly-relevant note, the search widget is an offshoot of [d2l-search-widget](https://github.com/Brightspace/d2l-search-widget-ui) (okay technically it's the other way around but shhhh), which is currently just used in the Filter menu. I have been toying with removing the `-custom` one altogether in some way, so it would likely have to move if that change went in anyway. (Again, "but I might do a thing in the future" isn't really good rationale for moving, I'm just adding this bit for context, the first paragraph is the important one 😜 )

localize-behavior.html
ryantmer: [This](http://search.dev.d2l/source/search?q=txtAdvancedSearch&defs=&refs=&path=%2Flp%2F_lang%2FLangTerms%2Ftranslations&hist=&type=&project=LSOne) gets you all the existing translations for "Advanced Search", so I guess we could just use those, yeah? (Okay well it gets you the line before each translation)

d2l-search-widget-custom.html
JoshuaKlassen: Yup, can do!!

localize-behavior.html
JoshuaKlassen: ah, forgot to comment about that! Unless there is any objections, I will use those existing translations!

ryantmer: In hindsight, we could probably have done that for several other langterms, heh. Especially things like "Department" and "Semester", which I suspect _may_ be used in _about 12,000 other places_ in the LMS.

d2l-search-widget-custom.html
JoshuaKlassen: Just to clarify, you mean to move everything up a level out of the `-custom` widget correct?

ryantmer: Oh, yep :) So just sticking it somewhere like right [before this](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-all-courses.html#L153) or wherever around there it should show up.

JoshuaKlassen: Awesome, thanks! :)

d2l-search-widget-custom.html
```diff
@@ -105,6 +105,20 @@
...
 				border-radius: 0.3rem;
 			}
+
+			.advancedSearch-link {
+				font-size: 0.8rem;
+				padding-top: 3px;
+			}
+
+			@media only screen and (max-width: 767px){
+				.advancedSearch-link {
```
jstefaniuk-d2l: In general we try to format css selectors without caps
```
advanced-search-link
```

JoshuaKlassen: Ok, thanks!!

localize-behavior.html
ryantmer: Fun fact! Just ran into @jeff-shepstone who is a knower of many language things, and it turns out it's actually preferred that we _do_ submit them for translation anyway. We won't pay for it if it's something that's already been done, but since context can change the translation, it's worth doing anyway.

I suspect in this case we might be okay, since it's just an independent bit of text and not part of a larger sentence or anything, but at the very least good to know for future reference! (And also means we did the right thing in submitting all the previous ones for translation 😄 )

d2l-all-courses.html
```diff
@@ -47,22 +47,32 @@
...
+			}
+			.advanced-search-link {
+				font-size: 0.8rem;
+				min-width: 100%;
 			}
 			.search-and-filter > d2l-search-widget-custom {
 				flex: calc(5 / 12);
 			}
+			.search-and-filter > #filterAndSort {
+				flex: calc(7 / 12);
```
apalaniuk: One thing I noticed is that Firefox doesn't support `calc` in `flex` properties, so I think we'll need to change them to literal values. This might be a good time to change that to make sure it works properly. If it causes major issues, can defer so long as the current layout looks unchanged from the previous on in Firefox.

apalaniuk:
Code LGTM

apalaniuk:
Code LGTM

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -37,12 +38,24 @@ describe('d2l-all-courses', function() {
...
 		sinon.stub(widget.$['search-widget'], '_search');
+    
+    widgetAdvancedSearch = fixture('d2l-all-courses-without-advanced-search-fixture');
 	});
 
 	afterEach(function() {
 		clock.restore();
 	});
 
+	it('should not render the advanced search link', function() {
```
CodeBaboon: What are the conditions for when this should not render vs when it should render the advanced search link? Something like:

```
describe('when some-sort-of-condition-exists then', function() {
    it('should not render the advanced search link', function() {
    }
});

describe('when some-other-condition-exists-in-fact-quite-possibly-the-exact-opposite-condition then', function() {
    it('should render the advanced search link', function() {
    }
});
```

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -37,12 +38,28 @@ describe('d2l-all-courses', function() {
 
 		widget = fixture('d2l-all-courses-fixture');
 		sinon.stub(widget.$['search-widget'], '_search');
+
+		widgetAdvancedSearch = fixture('d2l-all-courses-without-advanced-search-fixture');
```
CodeBaboon: this name almost means the opposite (might be better as `widgetNoAttributesSet` or something).

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -37,12 +38,28 @@ describe('d2l-all-courses', function() {
...
 	});
 
 	afterEach(function() {
 		clock.restore();
 	});
 
+	describe('when the advancedSearchUrl property has not been set then', function() {
+		it('should not render the advanced search link', function() {
+			expect(widgetAdvancedSearch.$.advancedSearchLink.getAttribute('class')).to.contain('d2l-all-courses-hidden');
```
CodeBaboon: Probably want to also confirm the link isn't there similar to the test below:

```
			var link = widgetAdvancedSearch.querySelector('.advanced-search-link > a');
			expect(link.getAttribute('href')).to.equal(null);
```

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -37,12 +38,28 @@ describe('d2l-all-courses', function() {
...
+		});
+	});
+
+	describe('when the advancedSearchUrl property has been set then', function() {
+		it('should render the advanced search link', function() {
+			var link = widget.querySelector('.advanced-search-link > a');
+			expect(link.getAttribute('href')).equals('/test/url/');
```
CodeBaboon: Sort of a nit, but since the stated expectation (and the logic) is that the link should be there if an href value exists, the expectation around the href could be more accurately stated as something like `expect(link.getAttribute('href')).length.to.be.above(0);`.

PR #267, opened 2017-01-10
Search on filter change

apalaniuk:


d2l-search-widget-custom.html
```diff
@@ -448,6 +452,12 @@
 				}
 			},
 
+			_onSearchDependencyChanged: function() {
+				if (this.isAttached) {
+					this._search();
+				}
```
ctwomey1: We should probably debounce this particular call.  If you check off 3 things quickly that would create 3 searches.  and if they come back out of order you won't get the expected results in the screen

d2l-all-courses.html
```diff
@@ -592,6 +590,7 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				this.toggleClass('focus', true, this.$.filterText);
 			},
 			_onFilterChanged: function(e) {
+				this._parentOrganizations = [];
 				this.set('_parentOrganizations', e.detail.filters);
 				this.set('_filterText', e.detail.text);
```
ctwomey1: Are you setting _parentOrganizations to an empty array to get around the reference issue with arrays?  we need to clone the filter array to get a separate reference?

apalaniuk: Yeah, this is to get around dirty checking and enforce an observable change, since `this.set` isn't observable with arrays, and using the array mutation methods didn't seem to make sense here.

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -36,6 +36,7 @@ describe('d2l-all-courses', function() {
 		clock = sinon.useFakeTimers();
 
 		widget = fixture('d2l-all-courses-fixture');
+		sinon.stub(widget.$['search-widget'], '_search');
```
CodeBaboon: Should there be a couple unit tests that the search is being fired when the sort or filter change?

CodeBaboon: Maybe test that we are debouncing, too.

apalaniuk:
Issues in testing:

- [ ] Search/filter not being applied on clear (pre-exising issue)
- [ ] Slighly unresponsive due to debounce - find reasonable value/talk with Bob
- [ ] Console errors relating to search dropdown (pre-existing issue)

cgalvind2l:
Testing Notes
Acceptance
- [ ] Searching with few results ("french" as d2lsupport)
   - [ ] Applying a filter that removes all the results does so
   - [ ] Applying a filter that removes some of the results does so
   - [ ] Applying a filter that removes no results does so
   - [ ] Applying two filters does an "OR" on the filter conditions
   - [ ] Removing the filters resets the list to all the search results
   - [ ] Sorting by the different fields works as expected
- [ ] Searching with lots of results ("intro" as d2lsupport)
   - [ ] Applying a filter that removes all the results does so
   - [ ] Applying a filter that removes some of the results does so
   - [ ] Applying a filter that removes no results does so
   - [ ] Applying two filters does an "OR" on the filter conditions
   - [ ] Removing the filters resets the list to all the search results
   - [ ] Sorting by the different fields works as expected
- [ ] Searching with no results
   - [ ] Applying a filter doesn't change anything
   - [ ] Applying sort fields doesn't change anything
   - [ ] Removing filters doesn't change anything
- [ ] Clearing the search term still uses the specified filters and sort fields on the course list

Usability
- [ ] The update of the search results list takes the "just right" amount of time (not too fast or too slow)

Accessibility
- [ ] The transition as the results change makes sense when using a screen reader
- [ ] Able to tab through the updated list of courses when the filter/sort is changed

PR #266, opened 2017-01-09
US77132 - View updates UI portion

ctwomey1:
UI work for US77132.  

can be tested by manually using setCourseUpdates method in course tile.

PR #270, opened 2017-01-12
US81171 - Ctwomey/view updates data

ctwomey1:
behavior for getting updates aggregate number.  Branched off #266 

US81171.

Acceptance criteria:
- Pulls updates data per course on initial load
- search provides correct updates when tile renders
- pin/unpin actions shouldn't impact the update number shown
- configuration from lms influences update count.

An update is defined by the configuration found in the lms:

screen shot 2017-01-20 at 4 00 00 pm

Each can be achieved by impersonating a student enrolled in a course and doing the relevant action (ie do a quiz, submit an assignment), or by impersonating a teacher and doing the relevant action (grade/reply to post, etc).


d2l-course-tile.html
```diff
@@ -294,6 +309,34 @@
 						break;
 				}
 			},
+			setCourseUpdates: function(updates) {
+				if (updates > 99) {
+					this._courseUpdates = '99+';
+					this._hasCourseUpdates = true;
+					this.toggleClass('d2l-updates-hidden', false, this.$.courseUpdates);
+				} else if (updates > 0 && updates <= 99) {
```
apalaniuk: Could drop the `<= 99` check (but probably not worth updating right now)

d2l-utility-behavior.html
```diff
@@ -66,6 +66,12 @@
 			var selfLink = entity.getLinkByRel('self');
 			return selfLink.href;
 		},
+		getOrgUnitId: function(enrollmentId) {
+			if (enrollmentId && typeof String) {
```
ctwomey1: Ha def a finger slip on that one, good catch

ctwomey1:
Putting back into WIP based on some feedback about configuration of the updates number.

d2l-my-courses.html
```diff
@@ -111,7 +113,8 @@
 			my-enrollments-entity="[[_lastEnrollmentsSearchResponse]]"
 			lastEnrollmentsSearchResponse="[[_lastEnrollmentsSearchResponse]]"
 			locale="[[locale]]"
-      		show-course-code="[[showCourseCode]]">
+      		show-course-code="[[showCourseCode]]"
+			update-count-map="[[updateCountMap]]">
```
apalaniuk: This should probably be updated with @ryantmer's change to create `d2l-all-courses` dynamically - since `d2l-all-courses` uses `UpdatesBehavior` directly, I don't think there are any issues due to losing binding or anything though.

d2l-updates-behavior.html
```diff
@@ -0,0 +1,57 @@
...
+			updateCountMap: {
+				type: Object,
+				value: function() {
+					return {};
+				}
+			},
+			_updatesURI: {
+				type: String,
+				value: function() {
+					return '/d2l/api/le/1.18/updates/myUpdates/?orgUnitIdsCSV=';
```
apalaniuk: Does this need to be unique per element?

d2l-updates-behavior.html
```diff
@@ -0,0 +1,57 @@
...
+				}
+			},
+			_ajaxElement: {
+				type: Object,
+				value: function() {
+					return document.createElement('d2l-ajax');
```
ryantmer: This should be created on `getUpdates` I think; it's a pretty heavy thing to have here (also not really needed until the function is called anyway).

d2l-all-courses.html
```diff
@@ -254,11 +257,6 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 					size="100">
 				</d2l-loading-spinner>
 			</div>
-
-			<d2l-loading-spinner
```
ryantmer: Why was this removed? :o We're still using it further down (or at least, we're show/hiding it)

d2l-course-tile.html
```diff
@@ -317,13 +332,41 @@
...
+			_onUpdatesChange: function(updateChange) {
+				if (this.enrollmentId) {
+					var orgUnitId = this.getOrgUnitId(this.enrollmentId);
+					var update = updateChange.base[orgUnitId];
+					if (update) {
+						this.setCourseUpdates(update);
+					}
+				}
+			},
+			_shortCourseText: function() {
```
ryantmer: This appears to be unused - not sure if it's usage is still WIP. If not, then this and `_hasCourseUpdates` can both go since it's not used anywhere else.

d2l-updates-behavior.html
```diff
@@ -0,0 +1,57 @@
...
+			var updates = this._updatesURI + csvOrgId;
+			this._ajaxElement.url = updates;
+			this._ajaxElement.onResponse = this._onUpdatesResponse.bind(this);
+			this._ajaxElement.method = 'GET';
+			this._ajaxElement.generateRequest();
+		},
+		_onUpdatesResponse: function(response) {
+			if (response.detail.status === 200) {
+				response.detail.xhr.response['Objects'].forEach(function(object) {
+					this.set('updateCountMap.' + object['OrgUnitId'], object['UnreadDiscussions'] + object['UnapprovedDiscussions'] +
```
ryantmer: nit: long/awkward line is long/awkward

d2l-all-courses.html
ctwomey1: probably just a bad merge

d2l-course-tile.html
ctwomey1: you're right it can go, it was a holdover from when it was a dom-if

d2l-updates-behavior.html
ctwomey1: no, should it be defined inline?

apalaniuk: Yeah, or directly in a `createUpdatesUrl(orgUnitIds)` or similar function so it's still simple to test

ryantmer:
Recent changes look good - I'll do the formal "Approve" once the UI PR has been merged and this one is rebased/merged (just as a last once-over).

d2l-course-tile.html
```diff
@@ -246,7 +251,14 @@
 
 				var organizationLink = this.enrollment.getLinkByRel(/\/organization$/);
 				this._organizationUrl = organizationLink.href + '?embedDepth=1';
-
+				//because the tile might not be init before the first search we need to check the map on ready
```
ryantmer: Can this be delayed? Just to keep `ready` fairly lightweight (also if it's deferred until later I don't know that you'd need the `if (this.updateCountMap)` check anymore... maybe?). I'm thinking maybe move it into `_onOrganizationResponse` - I realize it's not actually related to the organization response coming back, but that's the point where the tile is "ready" from a human-usable perspective, so it seems reasonable to do it in there.

d2l-course-tile.html
```diff
@@ -330,25 +342,33 @@
...
-					this._hasCourseUpdates = true;
 					this.toggleClass('d2l-updates-hidden', false, this.$.courseUpdates);
 				} else {
 					this._courseUpdates = 0;
-					this._hasCourseUpdates = false;
 					this.toggleClass('d2l-updates-hidden', true, this.$.courseUpdates);
 				}
 			},
+			_onUpdatesChange: function(updateChange) {
+				if (this.enrollmentId) {
```
ryantmer: Strictly speaking this check is redundant - you do the same one in `getOrgUnitId` itself.

d2l-course-tile.html
```diff
@@ -330,25 +342,33 @@
...
+			},
+
 			_enrollmentChanged: function() {
 				if (!this.delayLoad && this.enrollment) {
 					var organizationLink = this.enrollment.getLinkByRel(/\/organization$/);
 					if (!this._organizationUrl || this._organizationUrl.indexOf(organizationLink.href) !== 0) {
 						this._organizationUrl = organizationLink.href + '?embedDepth=1';
 						this.$.organizationRequest.generateRequest();
 					}
+
```
ryantmer: nit: extra space ("while you're here" :P)

d2l-updates-behavior.html
```diff
@@ -0,0 +1,79 @@
...
+				type: Object,
+				value: function() {
+					return {};
+				}
+			},
+			//config from outside the widget
+			courseUpdatesConfig: {
+				type: Object
+			},
+			_ajaxElement: {
```
ryantmer: This doesn't actually need to be a Polymer property (not like... a huge deal, but it's not being observed and not in the DOM or anything, so no need).

d2l-updates-behavior.html
```diff
@@ -0,0 +1,79 @@
...
+		getUpdates: function(csvOrgId) {
+			if (this.courseUpdatesConfig) {
+				if (typeof this.courseUpdatesConfig === 'string') {
+					this.courseUpdatesConfig = JSON.parse(this.courseUpdatesConfig);
+				}
+				this._ajaxElement = this._ajaxElement || document.createElement('d2l-ajax');
+				var updates = this._generateUpdatesUri(csvOrgId);
+				if (updates) {
+					this._ajaxElement.url = updates;
+					this._ajaxElement.onResponse = this._onUpdatesResponse.bind(this);
```
ryantmer: A perhaps more Polymer-pragmatic way to do this is to use `this.listen`, i.e. `this.listen(this._ajaxElement, 'iron-ajax-response', '_onUpdatesResponse')` (and then `this.unlisten` in `_onUpdatesResponse`).

d2l-updates-behavior.html
```diff
@@ -0,0 +1,79 @@
...
+				}
+			}
+		},
+		_generateUpdatesUri: function(csvOrgId) {
+			if (csvOrgId && typeof csvOrgId  === 'string') {
+				return '/d2l/api/le/1.18/updates/myUpdates/?orgUnitIdsCSV=' + csvOrgId;
+			}
+		},
+		_onUpdatesResponse: function(response) {
+			if (response.detail.status === 200) {
```
ryantmer: Does this API ever 304 Not Modified? (I'm assuming not, but might be worth checking)

test/index.html
```diff
@@ -25,7 +25,8 @@
 				'./d2l-touch-menu-item/d2l-touch-menu-item.html',
 				'./d2l-search-widget-custom/d2l-search-widget-custom.html',
 				'./d2l-utility-behavior/d2l-utility-behavior.html',
-				'./localize-behavior/localize-behavior.html'
+				'./localize-behavior/localize-behavior.html',
+				'./d2l-updates-behavior/d2l-updates-behavior.html'
```
ryantmer: nit: BUT ALL THE OTHER ONES ARE ALPHABETICAL 😿 😿 😿 😿 😿 😿 y chester y

d2l-updates-behavior.html
ctwomey1: http://docs.valence.desire2learn.com/res/updates.html, doesn't appear to

d2l-updates-behavior.html
ctwomey1: Care to elaborate on why?  Seems like more overhead for the same result.  We do use this method a few places in this repo currently (doesn't mean it's right, just saying)

d2l-updates-behavior.html
ctwomey1: It doesn't need to, but assigning it to polymer gives me scoping for free i think because of the mixin jazz?  Otherwise i'd be attaching it to window.d2l.mycourses.updatesajax, which kind of opens it up to weird abuse i think.

d2l-course-tile.html
ctwomey1: This is a pretty lightweight operation, more or less just setting a property. However i don't believe it can be delayed.

This had to be added because of the way all courses gets init now.  The initial observable event triggered in mycourses happens pre course tiles in d2l-all-courses.

test/index.html
ctwomey1: some people just like to watch the world burn

d2l-updates-behavior.html
ryantmer: Yeah, I'm not super committed to either, just an option that uses more "Polymer-y" things.

d2l-course-tile.html
ryantmer: Truth, it is lightweight - it's not really a performance thing in this case, it's more a "thing doesn't belong here" in a flow sense, as in this isn't required for the element to operate properly. Considering we don't even fetch the organization here (and that is super required for the tile to operate correctly), the updates don't really need to be set this early anyway.

Yarr, the observable event will happen before All Courses is around, but that wouldn't matter in this case - if the event has already happened, that just means All Courses (and therefore this tile) is being created when `updateCountMap` is already populated. If that's the case, `if (this.updateCountMap)` will be true, and we're golden to do it in `_onOrganizationsResponse` - if it still hasn't been set (e.g. if All Courses was opened really quickly, and the Valence updates route is super slow), then then `_onUpdatesChange` will take care of it.

So like:

- Updates call happens, sets `updateCountMap` in d2l-my-courses
- User opens All Courses, `updateCountMap` is propagated to it, then to its child course tiles
- Course tile fetches organization
- Course tile receives organization, sets text/link of the tile, fetches image, and grabs its update count from `updateCountMap`

d2l-updates-behavior.html
ryantmer: Oh derp, the fact that this was in a behaviour completely whooshed over my head, thought it was an element. nvm, carry on! :P

test/index.html
ryantmer: :fire: 🌎

d2l-course-tile.html
ctwomey1: ok, you've convinced me 👍.

My thought process was mostly around I only want to do this once because of the new way we are doing things.  However that probably didn't cover the case of a tile's enrollment changing.  Moving it to the organization response is probably the right thing to do to ensure that the update number is always correct.

ryantmer: Yeah, I mean it is unlikely that the enrollments won't have been set right away (since we're `dom-repeat`ing these to create them), but it is possible. But yeah, guaranteed we'll be fetching the right update number at that point.

d2l-course-tile.html
```diff
@@ -330,18 +334,22 @@
...
 					this._courseUpdates = updates;
-					this._hasCourseUpdates = true;
 					this.toggleClass('d2l-updates-hidden', false, this.$.courseUpdates);
 				} else {
 					this._courseUpdates = 0;
-					this._hasCourseUpdates = false;
 					this.toggleClass('d2l-updates-hidden', true, this.$.courseUpdates);
 				}
 			},
+			_onUpdatesChange: function(updateChange) {
```
ryantmer: Huh, only just noticed this - this and what's being done in `_onOrganizationResponse` is the same. Could just use `this.updateCountMap[orgUnitId]` here instead of the passed-in value, and call this from the response handler.

ctwomey1: what's passed in is guaranteed to be pushed down fully from the event, the map on the object is subject to polymer propagating things fast enough

ryantmer: Yeah fair. I haven't personally come across problems with that, but that was also generally in simpler cases than a potentially-complex object. Sounds good!

JoshuaKlassen:
Testing notes:

---
Acceptance Criteria: 
- [X] Pulls updates data per course on initial load 

- [ ] Search provides correct updates when tile renders 
 - If canonical namespaces is on, then no number will show up in the All Courses view
- [X] Pin/unpin actions shouldn't impact the number shown 

- [X] Configuration from lms influences update count

---
Notes:
- A new Assignment Submission counts as 4 updates?
- Un-attempted Quizzes is really just how many quizzes have not been completed by everybody, not an issue or anything related to this PR, just confusing

ryantmer:
[1942f66 irl](http://static.fjcdn.com/pictures/Csi_3c4720_195814.jpg)

JoshuaKlassen:
Notification count appears in All Courses view now with and without Canonical Namespaces!

---
Acceptance Criteria:

- [X] Pulls updates data per course on initial load

- [X] Search provides correct updates when tile renders

- [X] Pin/unpin actions shouldn't impact the number shown

- [X] Configuration from lms influences update count

PR #271, opened 2017-01-12
DE23814 - Removed display:none from the loading spinner, centered it as well

jstefaniuk-d2l:
https://rally1.rallydev.com/#/51191528503d/detail/defect/86395668444?fdp=true

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -48,6 +48,11 @@
 			.d2l-my-courses-hidden {
 				display: none;
 			}
+			.spinner-container {
+				display: flex;
+				justify-content: center;
+				align-items: center;
+			}
 			d2l-alert {
 				display: block;
 				margin-bottom: 20px;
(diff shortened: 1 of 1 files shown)


apalaniuk:
Code LGTM

apalaniuk:
Actually, one thing I noticed is the DE states that the spinner should hide later if the user has enrollments, which makes sense, but it's still done in `_showContent`. Has that changed?

ryantmer:
I think still do it in `_showContent`, just don't call `_showContent` until after the second (/enrollments/users/169 or whatever) call.

It would be great to even wait until the _n_ /organizations calls had finished as well, but that's trickier since we would have to keep track of them from outside of the course-tile, which we don't.

CodeBaboon:
I'm not sure about all the "wait for this before doing that" stuff, but from a testing perspective this looks pretty reasonably good to my eyes. 

Tested on:
- Chrome (MacOS) (simulating 3g connection speed)
- Firefox (MacOS)
- Safari (MacOS)
- Safari (iPad Air)
- Safari (iPhone 6)

ryantmer:
My concern would mostly be for slower connections - the /enrollments call is quite fast, since it's a small response (also doesn't hit a DB, not that that matters for connection speed), but /enrollments/users/169 can be significantly larger. That said, if it's a small enough time delta between the first coming back and the second coming back, then meh!

CodeBaboon:
@ryantmer I did view it on a (simulated) 3g connection and it seemed quite good. I think you wrote the defect though so 'twill be up to you if it gets past Acceptance.

ryantmer:
Oh, okay! Yeah they defect description was based on the theory that the second network call would be longer, but it was visible for long enough for there to not be a delay on a 3G connection then that seems fine by me :fireworks:

Issue #168, opened 2016-10-26
DE24225 - Filter dropdown doesn't size correctly


ryantmer:
The filter dropdown doesn't size correctly when it is first opened, or after the window has been resized.
1. Open the all courses overlay
2. Click to open the filter dropdown - note that the size is wrong (extends past current viewport if there are enough filter items)
3. Close dropdown
4. Open dropdown - note that the size is now correct, and a scrollbar appears if the list is long

Alternatively,
1. Open the all courses overlay
2. Click to open the filter dropdown - note that the size is wrong (extends past current viewport if there are enough filter items)
3. Close dropdown
4. Resize window
5. Open dropdown - note that the size is still wrong

If the window is resized when the dropdown is open, it correctly resizes itself. Not sure where this behaviour is coming from, but the resizing seems to be taken care of in [d2l-hierarchical-view](https://github.com/Brightspace/d2l-hierarchical-view-ui/blob/2d87071ff36ddee2bef8640cd90c97792b97265d/d2l-hierarchical-view-behavior.html#L194)

ryantmer:
Possibly related: with a user that has many departments/semesters (and so gets the tabbed view with search fields), search for something. Clear the search. On all browsers other than Chrome, the dropdown doesn't properly resize itself back to the correct size (it stays small).

PR #274, opened 2017-01-14
US81119 - Restyle and change search bar's internal behaviour

### LINKED ISSUE #245: Cache search results
As of #241, search in the all-courses overlay works, but results are not cached in any real way (we are using etagging, but that doesn't save us much). This should be changed to reduce server load and delay when searching, in particular when searching for an empty search term (i.e. the default view that shows when the all-courses overlay opens).

A very simple approach to this that may work is something like what [`d2l-search-widget`](https://github.com/Brightspace/d2l-search-widget-ui/) is using, which stores all search results within the component itself. This means it is cleared when the page is reloaded, but the main advantage to this is that it results in the empty search/default results loading immediately when the search field is cleared, rather than redoing the search.

### LINKED ISSUE #246: Clear search when field is emptied
As of #241, clearing the search field using the X button will reset the search, but the same doesn't occur when using backspace to clear the field. As backspace is typed, it should live search (debounced), until the search field is completely cleared. When the search field is emptied, the search should be immediate. In order to make this more responsive to the user, it would be good to implement #245 first.

### LINKED ISSUE #247: Hide search dropdown when focus isn't on text field
Right now, focusing any child of the search box (magnifying glass button (can that be removed?), the text field, or the clearing X button) will open the recent/suggested searches dropdown. This is a bit strange, and it should probably only open when the text field itself is focused, especially as of the search changes implemented in #241.

@apalaniuk Do you recall if there was a specific need for it to open when any child of `d2l-search-widget-custom` was opened, rather than just the text field?

### LINKED ISSUE #248: Hide search dropdown when user pressed Enter to search
As of #241, when a user hits Enter to search, the course tiles update, but the recent/suggested searches dropdown doesn't hide until focus leaves the search bar. This means that the first few results are hidden behind this dropdown.

Suggestion would be to hide the dropdown when Enter is pressed, allowing the search results to be more visible, and only re-open it when either:

- focus leaves, and then re-enters the search bar; or
- a new character is added or removed, i.e. a new live search is triggered

ryantmer:
The search bar as it was previously did not match the behaviour of other search bars in the LMS. In particular, the desired behaviour is such that the state switches between "searched" and "not searched". This is effectively reflected through the state of the button icon - if you have searched for your current term, the button (X) will clear the field, and "redo the search with no query", clearing it back to the "not searched" state. If you change the term (or, in the initial state), the button goes back to the magnifying glass, and activating it (or hitting Enter) does the query and enters the "searched" state.

This also involved some rework of the under-the-hood stuff here, to take fuller advantage of some of Polymer's features as well as address some minor issues that previous existed.

Fixes #247 (kind of a side effect)
Fixes #248 (kind of a side effect)

"While I was here" I also added caching of enrollments search responses, which fixes #245.

This will also close #246, since the search as is done in the LMS doesn't follow that same paradigm - when you use backspace to clear a search field in the LMS, it doesn't re-search with the empty query; either you have to hit Enter, or you can clear it by clicking the X button.

See also https://github.com/Brightspace/d2l-search-widget-ui/pull/13 - a related change (that component already had live search, so it's really just a UI/slight UX change there).


d2l-search-widget-custom.html
```diff
@@ -315,84 +288,104 @@
...
 
+			_clearOrSearch: function() {
+				var icon = this.$$('d2l-icon').getAttribute('icon');
+				if (icon.indexOf('close-default') > -1 || this._searchInput.length === 0) {
+					this._clearSearch();
+					this._setSearchIcon();
+				} else {
+					this._fullSearch();
+					this._setClearIcon();
 				}
```
ctwomey1: Depending on the icons file name to determine actions seems a bit odd to me, would it make more sense to toggle what the on tap does in the _clearSearch and _fullSearch functions

d2l-search-widget-custom.html
```diff
@@ -315,84 +288,104 @@
...
-				this._search();
 			},
 
-			_onSearchStringChanged: function(newString) {
+			_setClearIcon: function() {
+				this.$$('d2l-icon').setAttribute('icon', 'd2l-tier1:close-default');
```
ctwomey1: This is a pretty loose querySelector, you might want to be more specific (ie like adding a id/class or doing something like this.$$('div.search-bar d2l-icon') because if someone added  d2l-icon higher up in the template i believe that this would grab that instead of your intended element.

if you tighten up the querySelector you should do a pass through this file and correct all of them

d2l-search-widget-custom.html
ryantmer: Yeah, I'm not totally sold on this approach, but the alternative was keeping an internal state of "is currently searched"/"is currently cleared". I figured that information was redundant to the state of the icon, but I think you're right, it is more clear that way.

ryantmer:
Argh, yeah forgot tests. I got to the end of the day, committed, and moved to something else the next day. I'll switch this back to WIP while I'm working on them. Thanks!

ryantmer:
Hokay, added a couple tests, tightened the querySelector, changed my keyboard backlight colour, and solidified the "searched"/"not searched" state representation. I _think_ the change I made in that regard makes things a bit less convoluted, hence perhaps fewer tests than expected. I tested the three cases for icon-changing that I was playing with the UI while writing this, notably:

- searching for something (button click/Enter) should do the search and change the icon to the "X" icon
- clearing after searching should re-fetch the (cached irl, not-cached in the test) response and change to the "search" icon
- searching for something, then changing the search term should immediately go back to the "search" icon (not redo the search until button click/Enter though)

The first two use the `_updateIcon` call after `_isSearched` has been changed; unfortunately the third case requires calling the change to the icon directly (since `_isSearched === true` in that case, but we also want the search icon, not the clear icon).

JoshuaKlassen:
These are my notes from testing:

Testing notes:
	
[X] Icon/button is removed from left side of field
	-All browsers passed
[X] With an empty search field, the Search icon is right-aligned in the search field
	-All browsers passed
[] When a user types, no changes occurs
	-Internet Explorer and Edge failed
		-On the All Courses search widget, the X icon appears to the left of the Search Icon when typing
[] When a user clicks the search icon (or presses Enter), the search happens and the icon changes to the X icon
	-Failed on all browsers
		-The filter search widgets don't change the Icon when clicking, but work with pressing enter
[] When the X icon is clicked, the search field is cleared, and results are updated back to the results for an empty search/no search
	-Failed on all browsers
		-With the above criteria failing, I could only test with pressing enter first, then clicking the icons to get the X icon
			-Clicking the X icon does clear the search, but does not change the icon back to the Search Icon
	
Mobile:
Chrome dev tools mobile:
	-Same issues as desktop browser

iOS (Safari):
	-Same issues as desktop browser

Accessibility (NVDA Screen Reader):
	-Tabbing works
	
	-Not sure if the screen reader would original read out results here, but it doesn't now
	
Internet Explorer Notes:
	-The default behaviour in Internet Explorer is that clicking in the box udpates the Icon to the X Icon
	
	-The two search widgets in Filter (Semesters and Departments) appear stacked at first with the X Icon, and are not able to perform searchs, until either the Semester tab or the Department tab is clicked
		-Also when searching for 'Math' in the Departments filter, clearing it does something weird with the original results - the number of rows shrinks in that panel so I can't see all the departments
		
Edge notes:
	-Has the same issues as Internet Explorer with the two filter searches stacking
	
Other notes:
	-I noticed that the Magnifying Glass is flipped with respect to the `/d2l/le/content/123456/Home` Icon
		-I don't know how closly these search widgets are supposed to match the other widgets on the LMS
	
	-I noticed a defect(?) on mobile: Clicking the 'Filter' label opens and immediately closes the dropdown, I had to click the arrow to open it.
	
	-We can do a Zoom meeting to discuss some of the issues if they are confusing to understand - I can share my screen and show you

ryantmer:
Okay, I believe I addressed the three outstanding items you mentioned there - I had made some changes here, but forgot to make the same changes in `d2l-search-widget` (but that PR has now been updated). Also fixed the extra X that IE adds.

For the IE/Edge notes, I'll have to take a look tomorrow; not sure what in here would cause that change :S (does it do it on master, or only with this PR checked out?)

For the other notes, the flipped icon I presume is okay, although I hadn't noticed that... if anything, I'd argue the one in the LMS is backwards 😛 As for the filter menu on mobile, I believe we decided not to make it mobile-friendly (time constraints yay), although it's odd that it's showing up at all... Once I've got all my stuff back up and running I'll reach out to you if I need any more clarifications about the outstanding stuff. Thanks!

JoshuaKlassen:
I only tested with this PR checked out, I will see what master does in IE/Edge!

ryantmer:
Ah, okay - this PR alone will do the things for the All Courses search, but you'll need the search-widget PR for the Filter menu search boxes.

JoshuaKlassen:
Testing notes:

---------------------------------------------
Acceptence Criteria:
- [X] Icon/button is removed from left side of field

- [X] With an empty search field, Search icon is right-aligned in the search field 

- [X] When a user types, no changes occur 

- [X] When a user clicks the search icon (or presses Enter), the search happens and the icon changes to the X Icon 

- [X] When the X icon is clicked, the search field is cleared, and results are updated back to the results for an empty search/no search 

---------------------------------------------
Tested on:
- [X] Chrome 

- [X] Firefox 

- [X] Internet Explorer 

- [X] Microsoft Edge

- [X] Safari

- [X] Android (Chrome dev tools mobile view)

- [X] iOS (iPad)

---------------------------------------------
Other notes:
-The issue with the extra X icon appearing when typing in All Courses in IE/Edge no longer happens! (fixed)

-Edge no longer stacks the filter searches first time opening the dropdown (IE still does) 

-The issue with the list of departments shrinking happens in Firefox, IE, Edge, and Safari (I missed Firefox and Safari last time)

-There is an issue in the All Courses search bar, where after a search, if I delete a character (not the whole search string), the icon changes back to the 'Search Icon' (expected), but clicking it acts as the 'X' icon, cleaing the search bar and updating the results back to the results of an empty search.

-When typing in a filter search bar(Semester or Department), clicking inside the All Courses search bar gives mouse focus, but not keyboard focus, have to click inside again to fix. This exists in all browsers.

ryantmer:
Sweet, thanks! w.r.t. the other notes:

- The department/semester lists shrinking is happening on master, so wasn't introduced here looks like. That said, I suspect it's related to #168 - I've added a note there to look out for fixing that if we ever figure out wtf is causing that
- The weird focus thing also wasn't introduced here, but good catch! I've added a GH issue for that - it's pretty minor, but is the sort of thing that shouldn't be forgotten. See #279.
- [ ] Search button acting as a clear is definitely something I introduced here, I'll fix that up right meow

ryantmer:
Hokay, fixed the one outstanding issue, should be good now :)

JoshuaKlassen:
It looks like filter search bars now don't switch to the search icon, it stays the X icon (in Chrome)

ryantmer:
Oh dammit... that would be because I made the change here, but not in the dependency repo. Note to self: after this PR, move to a common behavior as @ctwomey1 very intelligently suggested...

ryantmer:
Updated the other repo and bumped the version here, should be good now. Sorry for all the back and forth! D:

JoshuaKlassen:
Everything looks good!

PR #275, opened 2017-01-14
US81354 - d2l-course-tile perf

ryantmer:
Spiritual successor to US80997.

There are still a few places where changes can be made to improve the overall performance and intelligence of our usage of WCs within the widget. In particular, the d2l-course-tile WC is a candidate for some changes, as it is the most commonly-used by far within the widget. Initial testing with polyperf showed an average of 2000ms (range 1800-2250ms) to create 250 d2l-course-tile components, which is very high. In comparison, doing the same test with d2l-ajax takes around 200ms, an order of magnitude less. Obviously the course tile is much more complex than d2l-ajax, but the goal of these performance improvements is to potentially defer/avoid some of that complexity, as it will have an effect on page load times.

See US for details, but four things will be done in this PR:

- Remove creation of `d2l-ajax` elements until they are needed
- Use `afterNextRender`
- Don't create the `d2l-dropdown-menu` until the course tile is hovered
- Don't re-create d2l-siren-parser elements, instead create it once and reuse it

## Acceptance criteria:

- Unfortunately, outside of polyperf, it's difficult to quantify the performance improvements of web components within a page (and even the results within polyperf can be inconsistent). It is possible that some improvement may be seen in D2L.Performance.timing on a page that loads the My Courses widget, though - ultimately we would want page.display to be reduced some.
- No change to UI/UX should be noticeable

## Polyperf Results:

- Element created 250 times per run
- 25 runs
- using one standard deviation for average value
- average reduced:
  - IE: ~7100ms ➡️ ~3800ms
  - Chrome: ~1300ms ➡️ ~700ms
  - FF: ~3100ms ➡️ ~1300ms

### Before (Chrome)

image

### After (Chrome)

image

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -25,32 +25,6 @@ the user has in that organization - student, teacher, TA, etc.
 	<template>
 		<style include="d2l-course-tile-styles"></style>
 
-		<!-- needs no-cache so that images refresh if the user gets here using the back button-->
-		<d2l-ajax
-			id="organizationRequest"
-			url="[[_organizationUrl]]"
-			headers='{ "Accept": "application/vnd.siren+json", "Cache-control": "no-cache", "Pragma": "no-cache" }'
-			on-iron-ajax-response="_onOrganizationResponse">
-		</d2l-ajax>
-
-		<d2l-ajax
-			id="enrollmentPinRequest"
-			url="[[_enrollmentPinUrl]]"
-			method="[[_enrollmentPinMethod]]"
-			body="[[_enrollmentPinBody]]"
-			headers='{ "Accept": "application/vnd.siren+json", "Content-Type": "application/x-www-form-urlencoded" }'
-			on-iron-ajax-response="_onEnrollmentPinResponse"
-			on-iron-ajax-error="_onEnrollmentPinError">
-		</d2l-ajax>
-
-		<d2l-ajax
-			id="telemetryRequest"
-			url="[[telemetryEndpoint]]"
-			body="[[_telemetryEvent]]"
-			headers='{ "Content-Type": "application/json" }'
-			method="POST">
-		</d2l-ajax>
-
 		<div class="tile-container"
 			on-mouseover="_hoverCourseTile"
 			on-mouseout="_onTileMouseOut">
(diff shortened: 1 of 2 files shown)


ryantmer:
Yay, finally got around to finishing this one off. Since this is entirely performance-focused (i.e. no behaviour should change), I didn't see a need for unit tests (aside from trivial things like "doesn't initially render with d2l-ajax elements", but that's of questionable value) - but it's entirely possible that some would be good, just in a way I didn't think of!

ctwomey1:
I think this is a great first cut, but i think we can go deeper.  Based on how google seems to be using iron ajax i think we can cut the elements to just 1 for the entire application, instead of 2 for every ajax call per course tile

[Example](https://github.com/PolymerElements/app-localize-behavior/blob/master/app-localize-behavior.html#L183)

If i'm understanding this correctly they set up a static version of iron ajax on the prototype which is then shared between all of the mixins.  to utilize this properly we'd need to have a more programatic approach to building d2l-ajax allowing us to look at the promise and hook into it cleaner.

ctwomey1:
Code LGTM, my suggestion doesn't block the current code from going in as it's still a large improvement

trevorjonesd2l:
Test cases... verifying things still work.

- [ ] Pinned Courses load
- [ ] Overflow menus load
- [ ] Pin/Unpin action performs pinning/unpinning
- [ ] Change Image button opens change image overlay
- [ ] Course Offer link navigates to the course offering page
- [ ] Unpinned Courses page properly in all courses overlay
- [ ] Fills drop-down menu with auto-search results
- [ ] Live updates tile on search
- [ ] Filtering filters search results
- [ ] All Semesters load
- [ ] Can search for semesters
- [ ] All Departments load
- [ ] Can search for departments
- [ ] Sorting sorts the search results
- [ ] Telemetry requests are sent successfully

ryantmer:
Updated the initial description with some polyperf details, showing what kind of effect this had on performance, so they can be reproduced if desired.

ryantmer:
Just so @ctwomey1's earlier comment isn't lost, I've opened it as https://github.com/Brightspace/d2l-my-courses-ui/issues/285.

PR #278, opened 2017-01-17
US80271- Match filter names

JoshuaKlassen:
US80271: My Courses Semester and Department filters should match the lang terms the user has set.

Basically, update the display names in the filter dropdown in the All Courses view to the Standard OU types for Semester and Department that the client has set.

Have only tested Chrome so far, will test other browsers.

To test: https://git.dev.d2l/projects/CORE/repos/le/pull-requests/6530/overview


d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -329,6 +339,12 @@
 			},
 			_updateSemesterFilter: function(e) {
 				this.__updateFilter(e, '_numSemesterFilters');
+			},
+			_filterStandardSemesterNameChanged: function() {
```
apalaniuk: We might be able to simplify this by [binding the property to textContent directly](https://www.polymer-project.org/1.0/docs/devguide/data-binding#bind-to-text-content) where we're currently localizing the department/filter terms, and instead set the default values of these properties to the localized values.

d2l-my-courses.html
```diff
@@ -333,6 +337,15 @@
 					allCourses.locale = this.locale;
 					allCourses.showCourseCode = this.showCourseCode;
 					allCourses.advancedSearchUrl = this.advancedSearchUrl;
+
+					if (this.standardSemesterName) {
```
apalaniuk: This makes me think we should be using [custom constructors](https://www.polymer-project.org/1.0/docs/devguide/registering-elements#custom-constructor) and do the validation there, so that we don't have to make the consumer worry about those checks.

apalaniuk:
Some tasks from offline talk:

- [ ] Ensure default/custom filter names persist when adding/removing filters
- [ ] Ensure screen reader reads custom filter names correctly (or, in the same manner as the default names)

JoshuaKlassen:
I am currently waiting on feedback from a designer about a few accessibility concerns and affecting the search bar in the filter dropdown.

localize-behavior.html
```diff
@@ -228,8 +228,8 @@
 								'filtering.clear': 'Effacer',
 								'filtering.semester': 'Semestre',
 								'filtering.department': 'Département',
-								'filtering.semesterMultiple': '({num}) semestre',
```
apalaniuk: Ah, so there is at least one instance where the order of this can change things like capitalization, so using the localized `filtering.semester`/`filtering.department` directly as an argument instead of using the literal makes a difference. In that case, we might need the original way, and another that accepts it as a parameter. I don't know if there's a more general way to do this..

ryantmer: Looking back at this now, it actually seems a bit weird anyway to me. _Semester (3)_ kinda makes sense in English, but _(3) semestre_ seems... odd, in French. Feels like they should be pluralized, with the number always after*

*at least, in French, which is not how you would say it (you'd say "trois semestres"), but as a visual representation, "Semestres (3)" seems more reasonable. But idk, this is probably a brain-sidetrack.

apalaniuk: I was just talking about this offline as well, but my French is rusty enough that I more or less keep those thoughts to myself ;) I'll leave this consideration to the more multilingual among us.

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -250,13 +260,35 @@
...
+				if (this.filterStandardSemesterName !== this.localize('filtering.semester')) {
+					this.$.semesterListButton.textContent = numSemesterFilters > 0
+						? this.localize('filtering.semesterMultipleCustom', 'semesterName', semesterFilterName, 'num', numSemesterFilters)
+						: semesterFilterName;
+				} else {
+					this.$.semesterListButton.textContent = numSemesterFilters > 0
+						? this.localize('filtering.semesterMultiple', 'num', numSemesterFilters)
+						: this.filterStandardSemesterName;
+				}
+
```
ctwomey1: You might be able to purge a lot of your logic if you use http://formatjs.io/guides/message-syntax/#plural-format for your localization.  The app-localize-behavior is a wrapper for this library

trevorjonesd2l:
The `{plural} Format` would be a nice addition. Overall the code looks good to me 👍

trevorjonesd2l:
Whoops didn't mean to close it

localize-behavior.html
```diff
@@ -52,20 +52,33 @@
...
-								'filtering.department': 'Department',
-								'filtering.semesterMultiple': 'Semester ({num})',
-								'filtering.departmentMultiple': 'Department ({num})',
+								'filtering.semesterDefault': 'Semester',
+								'filtering.departmentDefault': 'Department',
 								'filtering.searchDepartments': 'Search Departments',
 								'filtering.searchSemesters': 'Search Semesters',
+								'filtering.searchBy': 'Search by {filter}',
+								'filtering.semesterCustom':'{semesterName, select,' +
+															'default {Semester}' +
```
trevorjonesd2l: I'm not sure if this is possible in this context but it'd be nice if the `filtering.semesterDefault` text didn't have to be repeated

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -250,13 +260,21 @@
...
 			},
-			_numFiltersChanged: function() {
-				this.$.semesterListButton.textContent = this._numSemesterFilters > 0
-					? this.localize('filtering.semesterMultiple', 'num', this._numSemesterFilters)
-					: this.localize('filtering.semester');
-				this.$.departmentListButton.textContent = this._numDepartmentFilters > 0
-					? this.localize('filtering.departmentMultiple', 'num', this._numDepartmentFilters)
-					: this.localize('filtering.department');
+			_updateDepartmentFilterLabel: function(numDepartmentFilters, departmentFilterName) {
+				var selectArg = this.filterStandardDepartmentName === this.localize('filtering.departmentDefault') ? 'default' : departmentFilterName;
```
trevorjonesd2l: `selectArg` seems a little vague can we use something more specific here like `useDefaultName`

localize-behavior.html
JoshuaKlassen: I agree, I am just not sure if it's possible. I will look into it

JoshuaKlassen: One way would be to pass the value in as another parameter

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -261,8 +261,9 @@
 				}
 			},
 			_updateDepartmentFilterLabel: function(numDepartmentFilters, departmentFilterName) {
-				var selectArg = this.filterStandardDepartmentName === this.localize('filtering.departmentDefault') ? 'default' : departmentFilterName;
-				this.$.departmentListButton.textContent = this.localize('filtering.departmentCustom', 'departmentName', selectArg, 'num', numDepartmentFilters);
+				var defaultName = this.localize('filtering.departmentDefault');
+				var useDefaultName = this.filterStandardDepartmentName === defaultName ? 'default' : departmentFilterName;
+				this.$.departmentListButton.textContent = this.localize('filtering.departmentCustom', 'departmentName', useDefaultName, 'defaultDepartmentName', defaultName, 'num', numDepartmentFilters);
```
trevorjonesd2l: If you're going with this approach I feel like you might as well just pass in the name to use here rather than do the `useDefaultName` select

JoshuaKlassen:
According to the new translations, French won't be the odd case (with the Capital and lowercase issue, and with the number being in front) - meaning that the terms will look identical 

I am going to look into creating a custom formatter for this

trevorjonesd2l:
LGTM 👍

ctwomey1:
Acceptance Criteria:
If the standard type for Department or Semester is changed, the names of the tabs in the filter dropdown should change to reflect this change, e.g. if an OU type called "Term" is created, and is used as the Standard Semester OU type, then the filter menu should have "Term" as one of the tabs (the entries listed will already be the correct, Term, entries)

Testing plan:

- [ ] Toggling of Filters and departments should not change the custom term back

- [ ]  Opening/closing all courses should display correct term

- [ ] searching should not impact filter/department terms

- [ ] custom term should be used across all supported locales

- [ ] Term should be used in place of default term for accessibility

Works in:

- [ ] firefox

- [ ] chrome (of course it will)

- [ ] edge

- [ ] ie 11

- [ ] safari

ctwomey1:
Working as intended now, LGTM

PR #280, opened 2017-01-20
DE23907 Locale fix updates

ctwomey1:
Temporary solution to stop breakages until we have full translations.


apalaniuk:
Can we use [useKeyIfMissing](https://elements.polymer-project.org/elements/app-localize-behavior#property-useKeyIfMissing)? One thing I like about that is it will still display/announce something if it's missing, and is still easy to audit the resource file to see if any terms are missing.

test/localize-behavior/localize-behavior.js
```diff
@@ -25,4 +25,30 @@ describe('localize behavior', function() {
 		expect(component.localize('allCourses')).to.equal('All Courses');
 	});
 
+	describe('localize mappings', function() {
+		it('should have translation for every english term', function() {
```
apalaniuk: One thing about this is it will pass even if the translations are empty strings, ie. there is no translation for the term. So, we might think that we have translations, for all languages if this passes, when we really just have a value for each key in each language, even if that value isn't a translation.

ryantmer:
@apalaniuk We looked into it, but ultimately decided against - there's no reason to add terms without translations (barring delays in getting said translations). Plus, we're currently working off of Brightspace/app-localize-behavior, which doesn't have that in it :( (getting rid of our fork is on my to-do list; the things we added were for IE10, so no longer a concern for us).

test/localize-behavior/localize-behavior.js
ctwomey1: I'm roughly on the same page as you, I did add a separate test that checks that the value of what each translation is (currently commented out till we have proper translations).  My goal is I want it hard for someone to skip something that is required for our application to work in all langs. 

The reason I don't like useKeyIfMissing, is there's really no alert telling us "hey you didn't translate this thing", in general my preference is that if it's required a unit test should fail on it missing.

localize-behavior.html
```diff
@@ -620,7 +718,8 @@
 								'images.noResults': '找不到 {search} 的搜尋結果。',
 								'images.pleaseModify': '請檢查拼字或修改搜尋。',
 								'error.displayingImages': '糟糕！顯示影像時發生問題。請稍後再回來查看。',
-								'error.settingImage': '抱歉，我們現在無法變更您的影像。請稍後再試。'
+								'error.settingImage': '抱歉，我們現在無法變更您的影像。請稍後再試。',
+								'courseTile.updates': 'updates:'
 							}
 						};
```
jstefaniuk-d2l: I would maybe default to english terms at runtime rather than putting them in here:

```
langs.forEach( function(lang) {
    for (text in langs.en) {
        lang[text] = lang[text] || langs.en.text;
    }
}.bind(this))

return langs
```

ctwomey1: I think this function would get called for every instance of an element that implements this mixin  ([reference](https://www.polymer-project.org/1.0/docs/devguide/properties#configure-values)).  unless behavior properties behave slightly different?

If I were to use this method i'd need to cache it in a common place. Looking deep at the app-localize-behavior we're extending they have a deeper cache, but it looks like only for when they are resolving a translated term which doesn't help much.

JoshuaKlassen:
Testing notes:

---
Acceptance Criteria:

- [X] all supported langs have lang terms mapped 

- [X] rendering works in all locations of app in all locales
 - Checked My Courses and All Courses (that's where it would break before)

PR #281, opened 2017-01-21
Use PolymerElements/app-localize-behavior

ryantmer:
Our fork was used because it had changes for IE10, which is no longer required.


ryantmer:
(Failure here is expected, as `d2l-search-widget` change will need to be merged/released first)

PR #282, opened 2017-01-25
US81437: remove d2l-alert and make it a bower component

cgalvind2l:
Acceptance Criteria:

    d2l-alert lives independently from d2l-my-courses-ui as a Polymer web component
    d2l-my-courses-ui is refactored to use d2l-alert as a dependency
    nothing breaks

d2l-alert now lives here: https://github.com/Brightspace/alert

Intention is to someday move it to BrightspaceUI


bower.json
```diff
@@ -48,6 +48,7 @@
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
     "iron-pages": "^1.0.8",
     "neon-animation": "^1.2.4",
-    "polymer": "^1.7.0"
+    "polymer": "^1.7.0",
+    "d2l-alert": "git://github.com/Brightspace/alert.git#master"
```
CodeBaboon: nit, but it would be nice if this were up on line 32

CodeBaboon:
Outside of this repo, but d2l-alert repo could use a readme and a demo

cgalvind2l:
The README and demo is next on my list :)

bower.json
CodeBaboon: Oh and it would behoove us to do a release of d2l-alert and then link to that instead of master before we put this to BSI, I suspect.

cgalvind2l: OK; I'll add the README/Demo & do a release.

CodeBaboon:
Tested this by pulling this branch, clearing `bower_components`, `bower install`, then doing the same for brightspace-integration. My courses still showed the unpinned alert correctly (no noticeable change in behavior). Seems legit.

omsmith:
Don't want to actually register it as _d2l-alert_ just yet?

cgalvind2l:
Honestly I assumed that putting something in BrightspaceUI would require "more" (work? documentation? signed-off-edness? something) than leaving it in Brightspace, but clearly that assumption was incorrect and I'll be putting it into BrightspaceUI etc today.

PR #284, opened 2017-01-25
DE23942 - All Courses state not being cleared when closed and re-opened

JoshuaKlassen:
DE23942: All Courses state not being cleared when closed and re-opened

Acceptance Criteria:
 - Any text in the Search Bar should be cleared when closing and re-opening 
 - Any filters that have been applied should be cleared out (should display no filters)
 - Sort should be set to default Sort (Date Pinned)


d2l-all-courses.html
```diff
@@ -689,6 +693,30 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
+
+				for (var index = 0; index < sortOptions.length; index ++) {
+					var option = sortOptions[index];
+					option.selected = false;
+
+					if (option.value === this.sortDefaultValue) {
+						option.selected = true;
+					}
+				}
+
```
ctwomey1: Is there a reason you are manipulating the radio button list directly like this?  By doing this, you're potentially breaking functionality if the menu-item-radio changes in the future (ie accessibility changes, events that should fire on select, etc).  You could replace the function with :

this.$.sortDropdownMenu.querySelector('d2l-menu-item-radio[value=pinDate]').click();

d2l-all-courses.html
```diff
@@ -199,13 +199,13 @@
...
 								<d2l-dropdown-menu no-padding min-width="350">
-									<d2l-menu label="{{localize('sorting.sortBy')}}">
+									<d2l-menu id="sortDropdownMenu" label="{{localize('sorting.sortBy')}}">
 										<div class="dropdown-content-header">
 											<span class="dropdown-content-header-text">{{localize('sorting.sortBy')}}</span>
 										</div>
 										<d2l-menu-item-radio class="dropdown-content-gradient" value="courseCode" text="{{localize('sorting.courseCode')}}"></d2l-menu-item-radio>
 										<d2l-menu-item-radio value="courseName" text="{{localize('sorting.courseName')}}"></d2l-menu-item-radio>
-										<d2l-menu-item-radio value="pinDate" text="{{localize('sorting.datePinned')}}" selected></d2l-menu-item-radio>
+										<d2l-menu-item-radio value="pinDate" text="{{localize('sorting.datePinned')}}"></d2l-menu-item-radio>
```
ctwomey1: This is causing the selected checkmark to not be present when the all courses overlay loads for the first time, Should probably have the selected set. or set it on load

d2l-all-courses.html
```diff
@@ -689,6 +693,30 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 			_updateTileSizes: function() {
 				this._rescaleCourseTileRegions();
 				this._tileSizes = Math.floor(100 / this._numColsOverlay) + 'vw';
+			},
+			_onSimpleOverlayClosed: function() {
+				this._clearSearchWidget();
```
ctwomey1: Performance wise, if the person is closing the overlay we don't need to do processing until they open the overlay again, doing it on close causes a bunch of things to happen making the page seem less responsive and do glitchy things.

Could these things be done on load instead?

d2l-all-courses.html
JoshuaKlassen: That's how I saw it being done in other places (well in clearFilters)

you suggestion looks better, I will replace it

d2l-all-courses.html
JoshuaKlassen: huh, I could have sworn I set it on load, I guess not
I will set it on load
(I removed it from there so that the default value property would actually control that, changing the default value would change it everywhere, except I forgot the to set it on load)

d2l-all-courses.html
JoshuaKlassen: the overlay opens before the on load seems to complete, so the text would be visible for a bit before being wiped, so I had moved it to on close

d2l-all-courses.html
ctwomey1: Ok that makes sense.  The difference is these are radio button items instead of checkboxes, so they are suppose to be able to handle their own state.

JoshuaKlassen:
@ctwomey1 I made the changes as requested

d2l-all-courses.html
```diff
@@ -545,6 +554,8 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 				this._updateTileSizes();
 
 				this._filterText = this.localize('filtering.filter');
+				this.$.sortDropdownMenu.querySelector('d2l-menu-item-radio[value=' + this.defaultSortValue + ']').selected = true;
+				this._sortField = this.defaultSortValue;
```
ctwomey1: Does this selection need to happen if you're doing this https://github.com/Brightspace/d2l-my-courses-ui/pull/284/files#diff-98931a3c4fd43464bdaaf7d5c18ce360R683 on open?

JoshuaKlassen: Yes - I am not sure why, but without it then it won't be 'check marked'
I do believe I can safely remove the `this._sortField = this.defaultSortValue` though

ctwomey1: So instead of having a selector that can break, maybe we should revert back to the attribute on the html here https://github.com/Brightspace/d2l-my-courses-ui/pull/284/files#diff-98931a3c4fd43464bdaaf7d5c18ce360R208

JoshuaKlassen: I can re-add the selected attribute on 'pinDate' defeating the purpose of having a 'sortDefaultValue'
However, I am not a fan of then hard coding 'pinDate' everywhere

JoshuaKlassen: What I could do is not execute the reset on overlayLoad if there isn't anything to reset, i.e. the first time it's called since initialization of the selector hasn't happened yet
which gives the 'reset' meaning, separated from initialization

ctwomey1: why would you have to hard code pinDate?  I;m not saying you can never use this selector, but why can't you leave the default render value of selected?

ctwomey1: Talked it out offline, Makes sense.

trevorjonesd2l:
Testing


- [ ] Chrome
- [ ] Firefox
- [ ] IE
- [ ] Safari
- [ ] iOS


- [ ] Course search should be cleared
- [ ] Semester filter search should be cleared
- [ ] Department filter search should be cleared
- [ ] Semester filters should be cleared
- [ ] Department filters should be cleared
- [ ] Sort setting should reset to Date Pinned
- [ ] Course tiles should be updated (no live search)
- [ ] Clears on pressing overlay `X`
- [ ] Clears on `Esc`

trevorjonesd2l:
LGTM

PR #286, opened 2017-01-27
Fix sort fields to actually match MyOrgUnitSortField

ryantmer:
Tangentially related to DE23798, these don't actually match the names as defined in [MyOrgUnitSortField](https://git.dev.d2l/projects/CORE/repos/lp/browse/framework/core/D2L.LP/LP/Enrollments/MyOrgUnits/MyOrgUnitSortField.codegen). The disadvantage of this is that the values sent back have to be .ToLowerCase()'d then compared to hardcoded strings - if we use the correct values here, we can instead use the proper MyOrgUnitSortField.TryParse (generated by the codegen).

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -203,10 +203,10 @@
 										<div class="dropdown-content-header">
 											<span class="dropdown-content-header-text">{{localize('sorting.sortBy')}}</span>
 										</div>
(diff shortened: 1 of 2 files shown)


ryantmer:
This requires https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/4525/overview to be in first, hence the WIP.

ryantmer:
This is going to be done by the PR for DE23798, which should be up soon-ish.

PR #288, opened 2017-01-30
US79972: Update to Icons 3.

njostonehouse:


ryantmer:
(Note that the major version bump in both icons and d2l-dropdown is just for the new icons.)

PR #289, opened 2017-01-31
US81659 - Update lang terms

JoshuaKlassen:
Replace English default terms in non English locales with the proper translated term

Add NL locale


ctwomey1:
Tested and looks fine:

Test cases:
Cycle through all lms languages in chrome and ensure no exceptions are thrown in the console
Ensure new lang terms are reflected in the app by setting updates on, and setting courses into inactive, inactive started, inactive ended, starting later, already ended.

PR #291, opened 2017-02-01
specify border-box on elements with a width and padding/border

omsmith:
Found some issues when trying to enable native shadow dom in the LMS.

_homepages-ui_ has a `* { box-sizing: border-box; }` which my courses was inadvertently relying on. It was noticeable in the All Courses filter dropdown header, but I grepped for a few other cases as well. Not necessarily comprehensive. Could consider doing a similar `*` dealio within my-courses, by `*` isn't great.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -97,6 +97,7 @@
 				color: var(--d2l-color-ferrite);
 			}
 			.dropdown-content-header {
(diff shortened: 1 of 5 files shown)


ryantmer:
Dang, these all had the same problem? Dang.

omsmith:
Assumedly, based off of the CSS.

PR #292, opened 2017-02-02
explicit bump of loading-spinner

omsmith:


omsmith:
Fixes loading spinner behaviour under native shadow dom

PR #294, opened 2017-02-02
fix available width calc to account for shadow roots

omsmith:
@dlockhart this seems pretty garbage. thoughts?


d2l-course-tile-responsive-grid-behavior.html
```diff
@@ -69,17 +69,21 @@
 			* fine for the purpose of creating a decent layout for a hidden widget.
 			*/
 			_getAvailableWidth: function(node) {
+				if (node.nodeType === 11) {
+					return this._getAvailableWidth(node.host);
+				}
+
 				var width = node.getBoundingClientRect().width;
```
dlockhart: Not related to your change, but `getBoundingClientReact` can be very expensive. If a node being hidden is a common scenario and is the reason this is handling a width of `0` and walking up the DOM tree, then checking if `node.offsetParent` is `null` is a much cheaper way to know if it's hidden or not.

omsmith:
Adding @apalaniuk because he was maybe the author?

d2l-course-tile-responsive-grid-behavior.html
dlockhart: An alternate attempt at this entire method:

```
_getAvailableWidth: function(node) {

    if (node === null ) {
        return 0;
    }

    // not an "element" type or it's hidden
    if ( node.nodeType !== 1 || node.offsetParent === null ) {
        return _getAvailableWidth(node.parentNode);
    }

    var width = node.getBoundingClientRect().width;
    return width;

}
```

omsmith: Anyone against going with this?

ryantmer: As a total non-expert, seems totally reasonable to me!

omsmith:
Went with @dlockhart's change, it makes sense. Original author was assumedly Joel, so he can't comment.

PR #295, opened 2017-02-02
add explicit styles for hidden attribute

omsmith:
Seems you opted for a lil something different in https://github.com/Brightspace/d2l-my-courses-ui/pull/93 This was my fix though... feel free to do otherwise.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -20,6 +20,11 @@
 			:host {
 				display: block;
 			}
+
+			:host [hidden] {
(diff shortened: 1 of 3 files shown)


d2l-my-courses.html
```diff
@@ -20,6 +20,11 @@
 			:host {
 				display: block;
 			}
+
+			:host [hidden] {
```
omsmith: `hidden` was being used on an `a` tag, so added this as well. Dunno if needed

PR #298, opened 2017-02-04
DE23798 - Switch from sortField/sortDescending to sort

ryantmer:
See LP PR here - https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/4485/overview

Using this new `sort` query parameter allows us to specify multiple sort fields with different sort orders, which is what we want to do in the all courses overlay. When the user sorts by course code, what they're actually getting is enrollments sorted by descending pin date, then sorted by ascending course code. This ensures that pinned courses are always at the top, for any given sort order. This fixes the problem that was found when pinning an alphabetically-low-named course, sorting by course name, then searching for something with a lot of results wouldn't show the pinned course, as it wasn't coming back in the first page.

The exception to this is LastAccessed - due to technical limitations, it is not currently feasible to have multiple sorting. Instead, the results will be sorted only by LastAccessed date.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -161,7 +161,7 @@
 							id="search-widget"
 							my-enrollments-entity="[[myEnrollmentsEntity]]"
 							parent-organizations="[[_parentOrganizations]]"
-							sort-field="[[_sortField]]">
+							sort="[[_sortParameter]]">
 						</d2l-search-widget-custom>
 
 						<div id="filterAndSort" class="d2l-all-courses-hidden">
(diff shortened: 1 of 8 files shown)


trevorjonesd2l:
Code  👀 👍

JoshuaKlassen:
Testing notes:

Pinned `Semidefinite Optimization`
Sorted by `Course Name`

- [X] `Semidefinite Optimization` shows up under `Pinned Courses`

JoshuaKlassen:
LGTM

PR #299, opened 2017-02-08
jklassen/de23942-followup

JoshuaKlassen:
fix for the search-widget icon wasn't getting set back to the search icon when closing the overlay


trevorjonesd2l:
👍

PR #300, opened 2017-02-08
US81632 - Typography auditing

ryantmer:
Turns out a bunch of styles were either being double-applied, not applied, applied in ways that weren't the correct way, or not applied by worked by some random happenstance. These changes remove as much of this uncertainty as possible, by primarily relying on d2l-typography as much as possible, and not modifying those styles unless explicitly required. A lot of small changes here and there, many of which have no visual effect, a few of which have minor visual effects (but these align us to be more consistent with the rest of the LMS, which is using d2l-typography).

See also https://github.com/Brightspace/d2l-search-widget-ui/pull/17 and https://github.com/Brightspace/simple-overlay/pull/2.

PR #309, opened 2017-02-09
Fix button font (was falling back to Arial)

ryantmer:
Spiritual followup to https://github.com/Brightspace/d2l-my-courses-ui/pull/300, just missed a font-family.


ryantmer:
Imma be #2edgy5me and just merge this, since it's just a font change

PR #312, opened 2017-02-10
break out image and utility behavior

jstefaniuk-d2l:
Also see:
https://github.com/Brightspace/utility-behavior/pull/1
https://github.com/Brightspace/course-image/pull/1


bower.json
```diff
@@ -31,6 +31,7 @@
     "d2l-ajax": "^3.2.5",
     "d2l-alert": "git://github.com/Brightspace/alert.git#^0.0.1",
     "d2l-colors": "^2.2.3",
+    "d2l-course-image": "git://github.com/Brightspace/course-image.git#^0.0.1",
```
ryantmer: I'd suggest using `~0.0.1` here instead - it's not likely going to make a huge difference, but the advantage is that you can easily make breaking changes with the minor version that way (until you hit 1.0.0). (Same comment for the `d2l-utility-behavior` below)

d2l-course-tile-grid.html
```diff
@@ -4,7 +4,7 @@
 <link rel="import" href="localize-behavior.html">
 <link rel="import" href="d2l-interaction-detection-behavior.html">
 <link rel="import" href="d2l-course-tile-grid-styles.html">
-<link rel="import" href="d2l-utility-behavior.html">
+<link rel="import" href="../d2l-utility-behavior/d2l-utility-behavior.html">
```
ryantmer: It's nitty, but external deps are listed first - just in the off-chance we're overriding something, we want the "most local" version to be the one we get. (same comment applies elsewhere)

Also it just makes it look nice and I'm annoying like that 💯

cgalvind2l:
`npm test` fails locally with `ERROR finding /home/carolyn/git/d2l-my-courses-ui/components/*.html
Error: ENOENT: no such file or directory, open '/home/carolyn/git/d2l-my-courses-ui/components/*.html'
    at Error (native)`

jstefaniuk-d2l:
This one's going to fail for a bit until I update all of the stuff to use the new repo name

PR #313, opened 2017-02-11
US82482 - Use d2l-search-widget-behavior

ryantmer:
The intent here is to simplify `d2l-search-widget-custom` by relying on the existing behaviour. A few things are overriden from the behaviour (as we have more complex behaviour with this version of it), but it did significantly reduce overlap, and means future updates and changes will generally only have to happen in one place.

Note that a lot of this diff's nastiness is just moving functions around, so that things were easier to find...

TODO:

- [x] Merge https://github.com/Brightspace/d2l-search-widget-ui/pull/20
- [x] Use `clearButtonLabel` and `searchButtonLabel` from above
- [x] Merge https://github.com/Brightspace/d2l-search-widget-ui/pull/21
- [x] Update to use changes from above
- [x] Merge https://github.com/Brightspace/d2l-search-widget-ui/pull/23
- [x] Update to use changes from above... again.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -149,7 +149,9 @@
 					<div class="search-and-filter-row">
 						<d2l-search-widget-custom
 							id="search-widget"
-							my-enrollments-entity="[[myEnrollmentsEntity]]"
+							search-action="[[_enrollmentsSearchAction]]"
+							search-field-name="search"
+							cache-responses
 							parent-organizations="[[_parentOrganizations]]"
 							sort="[[_sortParameter]]">
 						</d2l-search-widget-custom>
@@ -361,7 +363,8 @@
 					type: Object,
 					value: function() {
 						return {};
-					}
+					},
+					observer: '_myEnrollmentsEntityChanged'
 				},
 				// Object containing the last response from an enrollments search request
 				lastEnrollmentsSearchResponse: Object,
@@ -399,7 +402,8 @@
 				// Determines if the filter menu can be shown (The number of enrollments >= enrollments required)
 				_hasManyEnrollments: false,
 				_noPinnedCoursesInSearch: false,
-				_noUnpinnedCoursesInSearch: false
+				_noUnpinnedCoursesInSearch: false,
+				_enrollmentsSearchAction: Object
 			},
 			behaviors: [
 				window.D2L.MyCourses.CourseTileResponsiveGridBehavior,
@@ -413,8 +417,7 @@
 				'd2l-simple-overlay-opening': '_onSimpleOverlayOpening',
 				'tile-remove-complete': '_onTileRemoveComplete',
 				'enrollment-pin-complete': '_onEnrollmentPinComplete',
-				'enrollment-unpin-complete': '_onEnrollmentUnpinComplete',
-				'd2l-search-enrollment-response': '_updateFilteredEnrollmentsEvent'
+				'enrollment-unpin-complete': '_onEnrollmentUnpinComplete'
 			},
 			observers: [
 				'_updateEnrollmentAlerts(_hasFilteredPinnedEnrollments, _hasFilteredUnpinnedEnrollments)',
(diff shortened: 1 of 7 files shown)


d2l-all-courses.html
```diff
@@ -502,50 +505,50 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
...
-						} else {
-							if (!this._unpinnedCoursesMap.hasOwnProperty(enrollmentId)) {
-								newUnpinnedEnrollments.push(enrollment);
-								this._unpinnedCoursesMap[enrollmentId] = true;
-							}
+					this._updateFilteredEnrollments(enrollments, true);
+				}
+			},
+			_updateFilteredEnrollments: function(enrollments, append) {
+				var enrollmentEntities = enrollments.getSubEntitiesByClass('enrollment');
```
ryantmer: Note that the majority of the change in this method was whitespace; I moved the 200 check up one level, since we only have to do it in the one case now.

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -101,6 +101,8 @@
 			<d2l-search-widget
 				id="semesterSearchWidget"
 				placeholder-text="{{localize('filtering.searchSemesters')}}"
+				search-button-label="{{localize('search')}}"
+				clear-button-label="{{localize('search.clearSearch')}}"
```
JoshuaKlassen: Can I request that instead of using `search` as the localized term, you use `filtering.searchBy` ?
(`filtering.searchBy` in EN is `Search by {filter}`, which handles pluralizing of Standard OrgUnitType names Semester and Department, and is used at line `286` and `297`)

ryantmer: Ah, yes! Totally forgot that that change went in a while ago, will do!

JoshuaKlassen: Thanks! 😄

ryantmer: Added in 94998a2 - changed the name of the standard department to "Department123" and the button got the correct "Search by Department123" `aria-label`.

JoshuaKlassen:
Testing notes:

---
Test plan:
- Open filter menu
- Select filters
- Search for `Semesters` and `Departments`
- Switch locales
- Use custom Standard OrgUnitType display names

---
Tested on:

- [ ] Chrome

- [ ] Firefox

- [ ] Internet Explorer

- [ ] Edge

- [ ] Safari

---
Notes:
- Not testing on mobile since the filter & sort dropdowns aren't visible on mobile

JoshuaKlassen:
Finished testing. The only issue I found is that the dropdown menu shrinking bug is back

test/d2l-search-widget-custom/d2l-search-widget-custom.js
```diff
@@ -63,92 +58,22 @@ describe('<d2l-search-widget-custom>', function() {
...
-	it('should perform a search when filter changes', function(done) {
-		widget.$.fullSearchRequest.addEventListener('iron-ajax-response', function() {
-			done();
-		});
-
+	it('should perform a search when filter changes', function() {
+		var spy = sandbox.spy(widget, 'search');
 		widget.set('parentOrganizations', [1, 2, 3]);
-
 		clock.tick(501);
```
JoshuaKlassen: should this test `expect(spy.called).to.be.true;` ?

ryantmer: You mean like one line down? 😝

JoshuaKlassen: ohhhhhh, yeah, ignore that 😛

ryantmer:
Mostly fixed the dropdown shrinking thing, but it looks like there's another possibly-better solution for that (which will also be a better fix than was implemented for #168 in #321). Added as #324.

JoshuaKlassen:
Thanks for creating #324, with that being tracked I am comfortable with signing off on this PR

PR #319, opened 2017-02-14
DE24238 - Use two columns for skinny screen + 2 or 3 pinned courses

ryantmer:
Without this, having 3 pinned courses would give a really weird display, where there was only one column on skinny screens, which resulted in huge course tiles. Only case with one column now is if you're on a skinny screen, and also only have one pinned course; otherwise it's at least two.

### DIFF
--- a/d2l-course-tile-responsive-grid-behavior.html
+++ b/d2l-course-tile-responsive-grid-behavior.html
@@ -40,7 +40,7 @@ s		*/
 				var numCols = 1;
 
 				if (width < 768) {
-					if (itemCount >= 4) {
+					if (itemCount >= 2) {
 						numCols = 2;
 					}
 				} else if (width < 992) {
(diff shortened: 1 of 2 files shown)


JoshuaKlassen:
Testing notes:

---
Test plan:

- pin 2-3 courses
- unpin courses till only 1 course is pinned

---
Tested on:

- [X] Chrome 

- [X] FireFox

- [X] Internet Explorer

- [X] Edge

- [X] Android: Chrome

- [X] iOS: Safari

- [X] iOS: Chrome

---
Other notes:

It looked like when unpinning courses till there is only 1 left, that last course tile doesn't resize on browser, and on mobile takes a little while before resizing.

JoshuaKlassen:
Tested, other than that last course tile resizing thing, which, as we discussed offline, needs further investigation, it looks good to me

PR #320, opened 2017-02-15
US82454 Empty Search States

trevorjonesd2l:
Waiting for lang term stuffs


d2l-all-courses.html
```diff
@@ -223,6 +223,10 @@
 
 				<h2 class="d2l-heading-3">{{localize('pinnedCourses')}}</h2>
 
+				<span id="no-pinned-in-search" hidden$="{{!_noPinnedCoursesInSearch}}">
```
ryantmer: I believe these can be `[[]]` rather than `{{}}`

PR #321, opened 2017-02-15
DE24225 - Force filter menu to resize after opening

### LINKED ISSUE #168: DE24225 - Filter dropdown doesn't size correctly

ryantmer:
If the list of semesters or departments is long, for some reason it will just extend past the bottom of the viewport rather than resizing correctly. To force it to size correctly, we can just fake-trigger a resize event, which causes magic to happen deep within d2l-menu (or maybe d2l-dropdown (or maybe both)) and it resizes to the right size. This isn't a hack, it's more just a... suboptimal solution, but I've spent hours trying to figure out another way, and not been able to really identify a "cleaner" way to do it.

Fixes #168.

### DIFF
--- a/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
+++ b/d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
@@ -341,12 +341,22 @@
 				this.__toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, false);
 				this.$.departmentSearchWidget.clear();
 				this.$.departmentList.querySelector('d2l-menu').resize();
+
+				setTimeout(function() {
+					// DE24225 - force dropdown to resize after opening
+					window.dispatchEvent(new Event('resize'));
+				}.bind(this), 200);
 			},
 			_selectSemesterList: function() {
 				this.__toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, true);
 				this.__toggleSelectedList(this.$.departmentList, this.$.departmentListButton, this.$.departmentListHighlight, false);
 				this.$.semesterSearchWidget.clear();
 				this.$.semesterList.querySelector('d2l-menu').resize();
+
+				setTimeout(function() {
+					// DE24225 - force dropdown to resize after opening
+					window.dispatchEvent(new Event('resize'));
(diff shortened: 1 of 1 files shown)


d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -341,12 +341,22 @@
 				this.__toggleSelectedList(this.$.semesterList, this.$.semesterListButton, this.$.semesterListHighlight, false);
 				this.$.departmentSearchWidget.clear();
 				this.$.departmentList.querySelector('d2l-menu').resize();
```
JoshuaKlassen: Is this the resize that isn't happening?

ryantmer: Nah, it does - I can't entirely explain why this doesn't work; I think the issue is that this happens before the menu is opened (on first creation), but moving it makes the menu work even less 😕

trevorjonesd2l: Do we need this `resize`?

ryantmer: Seemingly yes - if it's removed, the embedded `d2l-menu`s don't get a size at all (they're just the search widgets, no list of selectable items)

trevorjonesd2l:
Tested in IE, Chrome, FF, Safari

✔️

PR #322, opened 2017-02-15
Us81439 refactor image selector

cgalvind2l:
remove image selector and its various friends.

currently released as v0.0.1 here: https://github.com/Brightspace/image-selector


jstefaniuk-d2l:
It would be also nice if we could beak out all the simple overlay portions into a "image-selector-container" so that it'll be easier to just drop the image selector in somewhere

https://github.com/Brightspace/d2l-my-courses-ui/blob/138b1ec49f460a59b09ac740234719f328a82e83/d2l-my-courses.html#L123

Otherwise LGTM

CodeBaboon:
I'mma be, I'mma be, I'mma I'mma I'mma be
I'mma be, I'mma be, I'mma I'mma I'mma be
I'mma be, I'mma be, I'mma I'mma I'mma be
I'mma be be be be I'mma I'mma be
I'mma be be be be I'mma I'mma be
I'mma be be be be I'mma I'mma be

testing this.

CodeBaboon:
So when we launch the change image overlay we get an incorrect `POST`:

`POST http://vv81-dharris:44444/d2l/home 500 (Internal Server Error)`

Seems like we aren't correctly flowing the telemetry url down anymore because the Request Payload is `{"name":"LaunchChangeImage","ts":1487346664}`.

cgalvind2l:
Thanks Dave, will investigate.

cgalvind2l:
@CodeBaboon I'm seeing this problem on master as well, I do not think it is related to this particular PR. I will continue to investigate, but it shouldn't be a blocker here.

PR #323, opened 2017-02-16
Native transitions

omsmith:
Replace web-animations-js with direct mutations of `transform`/`transition` and additional refactorings. Also happens to fix some rendering issues with the current transitions in Chrome 57, and some that @trevorjonesd2l is seeing locally.

Will squash/rebase later today.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -52,7 +52,6 @@
     "iron-overlay-behavior": "^1.10.1",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
     "iron-pages": "^1.0.8",
-    "polymer": "^1.7.0",
-    "web-animations-js": "^2.2.2"
+    "polymer": "^1.7.0"
   }
 }
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -223,7 +223,7 @@
 				<d2l-course-tile-grid
 					id="all-courses-pinned"
 					enrollments="[[filteredPinnedEnrollments]]"
-					enrollments-queue="[[filteredPinnedEnrollmentsQueue]]"
+					enrollments-to-animate="[[_pinnedEnrollmentsToAnimate]]"
 					delay-load="[[delayLoad]]"
 					tile-sizes="[[_tileSizes]]"
 					locale="[[locale]]"
@@ -243,7 +243,7 @@
 				<d2l-course-tile-grid
 					id="all-courses-unpinned"
 					enrollments="[[filteredUnpinnedEnrollments]]"
-					enrollments-queue="[[filteredUnpinnedEnrollmentsQueue]]"
+					enrollments-to-animate="[[_unpinnedEnrollmentsToAnimate]]"
 					delay-load="[[delayLoad]]"
 					tile-sizes="[[_tileSizes]]"
 					locale="[[locale]]"
@@ -299,32 +299,6 @@
 					observer: '_unpinnedEnrollmentsChanged'
 				},
 				filteredUnpinnedEnrollments: {
-					type: Array,
-					value: function() {
-						return [];
-					}
-				},
-				// Set of enrollments Entities that are being transferred from unpinned to pinned state
-				pinnedEnrollmentsQueue: {
-					type: Array,
-					value: function() {
-						return [];
-					}
-				},
-				filteredPinnedEnrollmentsQueue: {
-					type: Array,
-					value: function() {
-						return [];
-					}
-				},
-				// Set of enrollments that are being transferred from pinned to unpinned state
-				unpinnedEnrollmentsQueue: {
-					type: Array,
-					value: function() {
-						return [];
-					}
-				},
-				filteredUnpinnedEnrollmentsQueue: {
 					type: Array,
 					value: function() {
 						return [];
@@ -415,7 +389,19 @@
 				// the endpoint for the telemetry service
 				telemetryEndpoint: String,
 				userId: String,
-				tenantId: String
+				tenantId: String,
+				_pinnedEnrollmentsToAnimate: {
+					type: Array,
+					value: function() {
+						return [];
+					}
+				},
+				_unpinnedEnrollmentsToAnimate: {
+					type: Array,
+					value: function() {
+						return [];
+					}
+				}
 			},
 			behaviors: [
 				window.D2L.MyCourses.CourseTileResponsiveGridBehavior,
@@ -433,11 +419,8 @@
 			},
 			observers: [
 				'_updateEnrollmentAlerts(_hasFilteredPinnedEnrollments, _hasFilteredUnpinnedEnrollments)',
-				'_enrollmentsChanged(filteredPinnedEnrollments.length, filteredPinnedEnrollmentsQueue.length, filteredUnpinnedEnrollments.length, filteredUnpinnedEnrollmentsQueue.length)'
+				'_enrollmentsChanged(filteredPinnedEnrollments.length, filteredUnpinnedEnrollments.length)'
 			],
-			created: function() {
-				this._tilesInPinStateTransition = [];
-			},
 			_loadNextEnrollmentsPage: function() {
 				if (this.$['all-courses'].opened && this.lastEnrollmentsSearchResponse) {
 					var parser = document.createElement('d2l-siren-parser');
@@ -452,14 +435,8 @@
 				}
 			},
 			_enrollmentsChanged: function() {
-				this._hasFilteredPinnedEnrollments =
-					this.filteredPinnedEnrollments.length > 0 ||
-					this.filteredPinnedEnrollmentsQueue.length > 0;
-				this._hasFilteredUnpinnedEnrollments =
-					this.filteredUnpinnedEnrollments.length > 0 ||
-					this.filteredUnpinnedEnrollmentsQueue.length > 0;
-				// Need to factor in queues here - if user has 1 enrollment, it will briefly
-				// disappear from both enrollments arrays when pinned/unpinned
+				this._hasFilteredPinnedEnrollments = this.filteredPinnedEnrollments.length > 0;
+				this._hasFilteredUnpinnedEnrollments = this.filteredUnpinnedEnrollments.length > 0;
 				this._hasFilteredEnrollments = this._hasFilteredPinnedEnrollments || this._hasFilteredUnpinnedEnrollments;
 				this.$['all-courses-scroll-threshold'].clearTriggers();
 			},
@@ -472,7 +449,8 @@
 					if (pinnedEnrollmentId === enrollmentId) {
 						var foundEnrollment = this.filteredUnpinnedEnrollments[index];
 						this._setEnrollmentPinData(foundEnrollment, true);
-						this.push('filteredPinnedEnrollmentsQueue', foundEnrollment);
+						this._pinnedEnrollmentsToAnimate.push(enrollmentId);
+						this.unshift('filteredPinnedEnrollments', foundEnrollment);
 						this.splice('filteredUnpinnedEnrollments', index, 1);
 						break;
 					}
@@ -487,7 +465,8 @@
 					if (unpinnedEnrollmentId === enrollmentId) {
 						var foundEnrollment = this.filteredPinnedEnrollments[index];
 						this._setEnrollmentPinData(foundEnrollment, false);
-						this.push('filteredUnpinnedEnrollmentsQueue', foundEnrollment);
+						this._unpinnedEnrollmentsToAnimate.push(enrollmentId);
+						this.unshift('filteredUnpinnedEnrollments', foundEnrollment);
 						this.splice('filteredPinnedEnrollments', index, 1);
 						break;
 					}
--- a/d2l-course-management-behavior.html
+++ b/d2l-course-management-behavior.html
@@ -35,30 +35,6 @@
 					},
 					notify: true
 				},
-				// Set of enrollments Entities that are being transferred from unpinned to pinned state
-				// Only used by `d2l-all-courses`
-				pinnedEnrollmentsQueue: {
-					type: Array,
-					value: function() {
-						return [];
-					},
-					notify: true
-				},
-				// Set of enrollments that are being transferred from pinned to unpinned state
-				// Only used by `d2l-all-courses`
-				unpinnedEnrollmentsQueue: {
-					type: Array,
-					value: function() {
-						return [];
-					},
-					notify: true
-				},
-				// If enabled, the component will use the pinned/unpinned enrollment queues
-				// In the case of `d2l-all-courses`, the enrollments are switching lists, so this is required.
-				usePendingLists: {
-					type: Boolean,
-					value: false
-				},
 				// True when there are pinned enrollments (i.e. `pinnedEnrollments.length > 0`)
 				_hasPinnedEnrollments: {
 					type: Boolean,
@@ -75,33 +51,17 @@
 				'tile-remove-complete': '_onTileRemoveComplete'
 			},
 			observers: [
-				'_enrollmentsChanged(pinnedEnrollments.length, pinnedEnrollmentsQueue.length, unpinnedEnrollments.length, unpinnedEnrollmentsQueue.length)'
+				'_enrollmentsChanged(pinnedEnrollments.length, unpinnedEnrollments.length)'
 			],
-			created: function() {
-				this._tilesInPinStateTransition = [];
-			},
 			_enrollmentsChanged: function() {
-				this.set('_hasPinnedEnrollments',
-					this.pinnedEnrollments.length > 0 ||
-					this.pinnedEnrollmentsQueue.length > 0);
-				// Need to factor in queues here - if user has 1 enrollment, it will briefly
-				// disappear from both enrollments arrays when pinned/unpinned
-				this.set('_hasEnrollments',
-					this._hasPinnedEnrollments ||
-					this.unpinnedEnrollments.length > 0 ||
-					this.unpinnedEnrollmentsQueue.length > 0);
+				this.set('_hasPinnedEnrollments', this.pinnedEnrollments.length > 0);
+				this.set('_hasEnrollments', this._hasPinnedEnrollments || this.unpinnedEnrollments.length > 0);
 			},
-			_addEnrollmentToList: function(pinned, enrollmentEntity, enrollmentId) {
+			_addEnrollmentToList: function(pinned, enrollmentEntity) {
 				var listName = (pinned ? 'pinned' : 'unpinned') + 'Enrollments';
 
 				this._setEnrollmentPinData(enrollmentEntity, pinned);
-
-				if (this.usePendingLists) {
-					this.push(listName + 'Queue', enrollmentEntity);
-				} else {
-					this._tilesInPinStateTransition.push(enrollmentId);
-					this.unshift(listName, enrollmentEntity);
-				}
+				this.unshift(listName, enrollmentEntity);
 			},
 			_moveEnrollmentToPinnedList: function(enrollment) {
 				var enrollmentId, enrollmentEntity;
@@ -123,7 +83,7 @@
 				}
 
 				if (enrollmentEntity) {
-					this._addEnrollmentToList(true, enrollmentEntity, enrollmentId);
+					this._addEnrollmentToList(true, enrollmentEntity);
 				}
 			},
 			_moveEnrollmentToUnpinnedList: function(enrollment) {
@@ -146,7 +106,7 @@
 				}
 
 				if (enrollmentEntity) {
-					this._addEnrollmentToList(false, enrollmentEntity, enrollmentId);
+					this._addEnrollmentToList(false, enrollmentEntity);
 				}
 			},
 			_onTileRemoveComplete: function(e) {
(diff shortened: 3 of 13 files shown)


bower.json
```diff
@@ -49,7 +49,6 @@
     "iron-overlay-behavior": "^1.10.1",
     "iron-scroll-threshold": "PolymerElements/iron-scroll-threshold#^1.0.2",
     "iron-pages": "^1.0.8",
-    "polymer": "^1.7.0",
-    "web-animations-js": "^2.2.2"
```
omsmith: This was the overall goal.

d2l-all-courses.html
```diff
@@ -303,32 +303,6 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 						return [];
 					}
 				},
-				// Set of enrollments Entities that are being transferred from unpinned to pinned state
-				pinnedEnrollmentsQueue: {
```
omsmith: Queues were legit and ensured overall correctness. The solved a race condition of pinning a new item before animations completed and potentially messing stuff up. I would also be pretty hard pressed to pin a new item in the ~500ms things are animating though.

d2l-all-courses.html
```diff
@@ -400,7 +374,19 @@ <h2 class="d2l-heading-3">{{localize('unpinnedCourses')}}</h2>
 					value: 20
 				},
 				// Determines if the filter menu can be shown (The number of enrollments >= enrollments required)
-				_hasManyEnrollments: false
+				_hasManyEnrollments: false,
+				_pinnedEnrollmentsToAnimate: {
```
omsmith: This could really probably be a single list (which I believe it was before as "pins in transition"). Up to whoever if we want to change it.

d2l-course-management-behavior.html
```diff
@@ -35,30 +35,6 @@
 					},
 					notify: true
 				},
-				// Set of enrollments Entities that are being transferred from unpinned to pinned state
-				// Only used by `d2l-all-courses`
-				pinnedEnrollmentsQueue: {
```
omsmith: This functionality wasn't actually being used within course-management-behavior, seemingly having been superseded by d2l-all-courses.

d2l-course-tile-grid-styles.html
```diff
@@ -36,6 +36,7 @@
 			flex-wrap: wrap;
 			flex-direction: row;
 			align-items: flex-start;
+			align-content: flex-start;
```
omsmith: This should've should've maybe been seen before? Maybe it only came up in my "only resize the container" case though. The default value is `stretch` so the rows of tiles were spacing themselves out through the height of the container, and thus moving around while it animated.

d2l-course-tile-grid.html
```diff
@@ -19,7 +18,7 @@
 				<d2l-course-tile
 					enrollment="[[item]]"
 					hover-enabled="[[_hoverInteractionEnabled]]"
-					animate-insertion="[[_removeEnrollmentFromTransitionList(item)]]"
+					animate-insertion="[[_grid_shouldAnimateEnrollmentInsertion(item)]]"
```
omsmith: Same strategy, basically

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+		}
+
+		/*
+		* Behavior for sliding grid tiles on insertion/removal
+		*
+		* @polymerBehavior window.D2L.MyCourses.CourseTileSlidingGridBehavior
+		*/
+		window.D2L.MyCourses.CourseTileSlidingGridBehavior = {
+			observers: [
+				'__slide_onEnrollmentsChanged(enrollments.*)'
```
omsmith: I unfortunately couldn't think of a good way to pass in the "collection to watch" as a parameter or whatever, nor did I see a programmatic api for observers.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+
+	(function() {
+		window.D2L = window.D2L || {};
+		window.D2L.MyCourses = window.D2L.MyCourses || {};
+
+		/**
+		 * this method, given an index in the grid, and a count of inserted
+		 * or removed tiles, will describe the visual position change of the tile.
+		 */
+		function calculatePositionChange(columns, i, insert, count) {
```
omsmith: This supports `count` because it made the algorithm more correct (I was sure it worked opposed to just working due to columns and count = 1).

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+			// "current" position
+			var col = i % columns;
+			var row = Math.floor(i / columns);
+
+			// position after change
+			var newCol = insert
+				? (col + count) % columns
+				: (((col - count) % columns) + columns) % columns;
```
omsmith: This insane line is to get around JS's modulus doing negative numbers incorrectly. Good memory @ryantmer!

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+				var newTiles = this.__slide_knownGridTileElements = this._getGridTileElements();
+
+				if (oldTiles.length < newTiles.length) {
+					this.__slide_onTilesAdded(oldTiles, newTiles);
+				} else {
+					this.__slide_onTilesRemoved(oldTiles, newTiles);
+				}
+			},
+
+			__slide_onTilesAdded: function slideOnTilesAdded(oldTiles, newTiles) {
```
omsmith: This and `_onTilesRemoved` should really just be merged, they're exactly the same.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+			detached: function slideBehaviorDetached() {
+				this.__slide_knownGridTileElements = null;
+				this.unlisten(this._getGridTileRepeat(), 'dom-change', '__slide_onDomChange');
+			},
+
+			__slide_onEnrollmentsChanged: function slideOnEnrolmmentsChanged() {
+				// dom is about to change. preemptively set current container height
+				// so we can animate it later
+				var container = this._getGridContainerElement();
+				container.style.height = container.scrollHeight + 'px';
```
omsmith: Kinda seemed like the timing on this wasn't perfect in Safari when Trevor was testing it. Would have to dig further.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					this.__slide_onTilesAdded(oldTiles, newTiles);
+				} else {
+					this.__slide_onTilesRemoved(oldTiles, newTiles);
+				}
+			},
+
+			__slide_onTilesAdded: function slideOnTilesAdded(oldTiles, newTiles) {
+				var diff = findDifferenceInLists(newTiles, oldTiles);
+
+				this.__slide_rowChange = calculatePositionChange(
```
omsmith: This is a little globally, so a lil yuck there, but I didn't want to pass it through the few functions at the time. This is needed for `.row === 0` and `.row < 0`.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					? n - 1
+					: pos + count;
+				var next = insert
+					? function() { --i; }
+					: function() { ++i; };
+				var done = insert
+					? function() { return i === pos - 1; }
+					: function() { return i === n; };
+
+				var delayPerTile = Math.min(50, 200 / (n - pos - count));
```
omsmith: A little different than the previous `delayPerTile`. I'm only dividing by tiles we'll be animating, previous one did the full set of tiles. Only different when pinning/unpinning from middle of the list. Oh boy does it not matter though when you're talking portions of 50ms though.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					delay += delayPerTile;
+
+					next();
+					if (done()) {
+						break;
+					}
+				}
+
+				if (!insert) {
+					this.__slide_resizeContainer(delay + 100);
```
omsmith: 100 is hardcoded as "give tiles moving up a row enough time to move out of the way"

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+				} else {
+					this.customStyle['--insertion-delay'] = (delay + 300) + 'ms';
```
omsmith: 300 is hardcoded as the translate tile time

omsmith: Really though I think a good follow-up to this would be to remove the insertion transitions and handle that here. Right now a lot of this is "quick translate things and set styles before the page actually (noticably) renders with the new dom elements!" which means we're forcing ourselves into doing forced recalculations. Avoiding forced recalcs is waiting for new frames, but at that point your new tiles have rendered :P.

Might be able to do better if we handled more of the stack.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+				}
+			},
+
+			__slide_resizeContainer: function slideResizeContainer(delay) {
+				var container = this._getGridContainerElement();
+
+				if (this.__slide_rowChange === 0
+					&& container.offsetHeight !== 0
+				) {
+					container.style.height = '';
```
omsmith: Maybe if we interpreted the enrollment array change (in the observer) we could avoid setting `height` in the first place, and thus also avoiding having to clear it here. I would leave that as a follow-up.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					return;
+				}
+
+				var targetHeight = 0;
+
+				if (this.__slide_rowChange < 0) {
```
omsmith: Unfortunately we can't just set `scrollHeight` in this case, since we've explicitly set the height to something larger than the new content, and thus `scrollHeight` will be as well.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					var tiles = this.__slide_knownGridTileElements;
+					if (tiles.length > 0) {
+						var lastTile = tiles[tiles.length - 1];
+						targetHeight = lastTile.offsetTop;
```
omsmith: I thought this was going to be `offsetTop + offsetHeight`, but I guess not?

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					}
+				} else {
+					targetHeight = container.scrollHeight;
+				}
+
+				if (targetHeight === container.offsetHeight) {
```
omsmith: Don't 100% recall where this happens, but it does and it's important to check, because otherwise there won't be a transition and we'll be stuck.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					container.style.height = '';
+					return;
+				}
+
+				container.style.transition = '200ms height linear ' + delay + 'ms';
+				container.style.height = targetHeight + 'px';
+
+				container.addEventListener('transitionend', function resized(e) {
+					if (e.target !== container) {
```
omsmith: Check the target, because _transitionend_ bubbles and it might've been one of the tiles or whatever.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+						return;
+					}
+
+					container.removeEventListener('transitionend', resized);
+
+					container.style.transition = '';
```
omsmith: Using `''` because IE interprets `null` as like, a string value.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					container.style.height = '';
+				});
+			},
+
+			__slide_translateTile: function slideTranslateTile(tile, change, delay) {
+				tile.style.transform =
+					'translate3d('
+					+ (-change.col * 101) + '%,'
```
omsmith: `101` accounts for the tile width + 1% margin from the styles. Makes sense?

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					+ (-change.row * 102) + '%,'
```
omsmith: This used to be 100% + 0.75rem (for the bottom margin), but IE bails on calc within the translate. That wasn't perfect either though, since row height can vary based on the course name in other tiles.

Either way, felt it was close enough for 300ms of visual flair.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					+ '0)';
+
+				Polymer.RenderStatus.afterNextRender(tile, function slideTileAfterNextRender() {
+					tile.style.transition = 'transform 300ms cubic-bezier(0.91, 0.03, 0.85, 0.36) ' + delay + 'ms';
+					tile.style.transform = 'translate3d(0, 0, 0)';
```
omsmith: IE didn't like `transform` just getting cleared, so set it to zero.

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+			},
+
+			__slide_translateTile: function slideTranslateTile(tile, change, delay) {
+				tile.style.transform =
+					'translate3d('
+					+ (-change.col * 101) + '%,'
+					+ (-change.row * 102) + '%,'
+					+ '0)';
+
+				Polymer.RenderStatus.afterNextRender(tile, function slideTileAfterNextRender() {
```
omsmith: Doing it `afterNextRender` (alternative syntax would be `requestAnimationFrame(() => requestAnimationFrame(() => {}))`) so the initial `transform` style can take effect before we change it (we also want to make sure it doesn't _transition_ to it's old location).

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,266 @@
...
+					tile.style.transition = 'transform 300ms cubic-bezier(0.91, 0.03, 0.85, 0.36) ' + delay + 'ms';
+					tile.style.transform = 'translate3d(0, 0, 0)';
+
+					tile.addEventListener('transitionend', function translated(e) {
+						if (e.target !== tile) {
```
omsmith: Check for target since it could've been a hover effect or something

d2l-course-tile-styles.html
```diff
@@ -14,7 +14,6 @@
 				margin-bottom: 0px;
 				--scale-fade-max-scale: 1.5;
 				--scale-fade-min-scale: 1.05;
-				--insertion-delay: 0.01s;
```
omsmith: Reading this with a default below let me set `--insertion-delay` on the whole grid instead of directly on the tile.

Directly on the tile might mean less css rewriting in the polyfill, but... *shrug*

d2l-course-tile-styles.html
```diff
@@ -53,19 +52,15 @@
 				transform: scale(1.05);
 				box-shadow: 3px 2px 10px rgba(0, 0, 0, 0.3);
 			}
-			:host(.animate-insertion-pre) {
-				position: absolute;
```
omsmith: Didn't see why we wanted these absolutes

d2l-course-tile-styles.html
```diff
@@ -53,19 +52,15 @@
...
-			}
-			:host(.animate-insertion-pre) > .tile-container {
-				animation: var(--insertion-delay) forwards tile-pre-insertion;
+			:host([animate-insertion]:not(.animate-insertion):not(.animation-complete)) > .tile-container {
```
omsmith: Base it off of the attribute / something that is there immediately. Previously the property guided us setting a class within `ready`, but that actually meant that the tile wasn't set to `opacity:0` very briefly, and just causing additional jank potential. Now it will be opacity 0 as soon as it renders. Added `animation-complete` to compensate / replace `.animation-insertion-pre`

d2l-course-tile-styles.html
```diff
@@ -53,19 +52,15 @@
...
+				animation: var(--insertion-delay, 10ms) forwards tile-pre-insertion;
+				animation-iteration-count: 1;
```
omsmith: Wasn't necessary, but it felt like there would be some possibility that it could start a second iteration before our javascript code ran to remove it? I dunno, made sense to me.

d2l-course-tile.html
```diff
@@ -111,7 +111,8 @@
 
 			properties: {
 				animateInsertion: {
-					type: Boolean
+					type: Boolean,
+					reflectToAttribute: true
```
omsmith: Wanted this to end up as an attribute.

d2l-course-tile.html
```diff
@@ -422,18 +420,17 @@
 				this._hoverCourseTile();
 			},
 			_onAnimationComplete: function(e) {
-				var tileContainer = this.$$('.tile-container');
+				if (e.target !== this.$$('.tile-container')) {
```
omsmith: Not interested in any inner animations.

d2l-course-tile.html
```diff
@@ -422,18 +420,17 @@
...
+					return;
+				}
 
 				if (e.animationName.indexOf('scale-and-fade-out') > -1) {
 					this.fire('tile-remove-complete', {enrollment: this.enrollment, pinned: this._pendingPinAction});
 				} else if (e.animationName.indexOf('scale-and-fade-in') > -1) {
-					this.toggleClass('animate-insertion', false, tileContainer);
```
omsmith: Don't know why these classes were being added on `d2l-course-tile` and `.tile-container`

d2l-course-tile.html
```diff
@@ -422,18 +420,17 @@
...
 					this.toggleClass('animate-insertion', false, this);
-					this.toggleClass('animate-insertion', false, this.$$('.tile-container'));
```
omsmith: Also this was the same as `tileContainer` 2 lines above, lol.

d2l-course-tile.html
```diff
@@ -422,18 +420,17 @@
...
+					this.toggleClass('animation-complete', true, this);
```
omsmith: Also, `this.toggleClass` is frustrating because it doesn't appear to actually be the same API as `ClassList.toggle`, which is, uhg.

d2l-course-tile-grid.html
```diff
@@ -133,232 +117,30 @@
...
-				this._animationState = this._animationState || {};
-
-				this._animationState.isAnimating = false;
-				this._animationState.isInsertion = false;
-				this._animationState.scaleAnimationComplete = false;
-				this._animationState.slideAnimationComplete = false;
-				this._animationState.lastAnimatedCourseTileId = null;
-				this._animationState.newCourseTile = null;
-			},
-			_doTileSlideAnimation: function(isInsertion, tileIndex, container, columns) {
```
omsmith: This method was very likely more accurate, positionally speaking, but grabbing clientRectangles for everything is undesirably.

d2l-course-tile-sliding-grid-behavior.html
apalaniuk: `(col + columns - count) % columns`?

omsmith: I think that's approximately right, as long as there isn't a 2-row change. Which current designs obviously isn't given only one tile ever changes at a tim), but it would start being incorrect if we did.

d2l-course-tile-sliding-grid-behavior.html
ryantmer: I... cannot figure that out either =/ Is the bottom the top in `offset` or something?

d2l-course-tile-sliding-grid-behavior.html
ryantmer: wat.

d2l-course-tile-sliding-grid-behavior.html
ryantmer: Makes sense!

d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -0,0 +1,246 @@
...
+			}
+
+			return {
+				pos: smallerList.length,
+				count: count
+			};
+		}
+
+		/* @this */
+		function verifyFunctionPresent(name) {
```
omsmith: Mm, no hadn't super considered doing it that way (thought I do remember seeing it now). I do believe I might prefer this though, as it's less lines overall, and it'll fail immediately, as opposed to when that function happens to get used (and as such the bug should be easier to catch).

omsmith: Mm, no hadn't super considered doing it that way (thought I do remember seeing it now). I do believe I might prefer this though, as it's less lines overall, and it'll fail immediately, as opposed to when that function happens to get used (and as such the bug should be easier to catch).

omsmith:
I'll add a test for the positioning stuff - sounds like a plan.

PR #325, opened 2017-02-17
Fix Telemetry Fail

cgalvind2l:
Fixes a problem where launching the Change Image UI from the All Courses view would fail the telemetry request.


CodeBaboon:
Code LGTM, I'll give it a quick spin for test porpoises.

ryantmer:
https://www.youtube.com/watch?v=Lj0ehBeCu0c

PR #330, opened 2017-02-22
US82453 - Split lang terms

trevorjonesd2l:
Split the lang terms into seperate files to make updating languages a bit easier


lang/ar.html
```diff
@@ -0,0 +1,80 @@
...
+	'use strict';
+
+	window.D2L = window.D2L || {};
+	window.D2L.MyCourses = window.D2L.MyCourses || {};
+
+	/*
+	* Ar lang terms.
+	* @polymerBehavior window.D2L.MyCourses.LangArBehavior
+	*/
+	window.D2L.MyCourses.LangArBehavior = {
```
ryantmer: Could these instead be in, e.g., `window.D2L.MyCourses.LocalizeBehavior.LangArBehavior`? [Namespaces are one honking great idea -- let's do more of those!](https://www.python.org/dev/peps/pep-0020/)

trevorjonesd2l: `world.D2L.Ryan.Idea.Good`

PR #332, opened 2017-02-23
Remove typography imports

ryantmer:
Since d2l-typography doesn't contain any mixins or variables, we don't actually need to import them - using the styles contained within is taken care of by BSI, which does the import. (This doesn't apply to d2l-typography-shared-styles, which does contain mixins and therefore does need to be imported. Also, we still need to import it in the demo page(s), since there's no BSI there.)


d2l-all-courses.html
```diff
@@ -99,10 +96,10 @@
...
 											<span>{{localize('sorting.sortBy')}}</span>
 										</div>
-										<d2l-menu-item-radio class="d2l-body-small dropdown-content-gradient" value="OrgUnitName" text="{{localize('sorting.sortCourseName')}}"></d2l-menu-item-radio>
-										<d2l-menu-item-radio class="d2l-body-small" value="OrgUnitCode" text="{{localize('sorting.sortCourseCode')}}"></d2l-menu-item-radio>
-										<d2l-menu-item-radio class="d2l-body-small" value="PinDate" text="{{localize('sorting.sortDatePinned')}}"></d2l-menu-item-radio>
-										<d2l-menu-item-radio class="d2l-body-small" value="LastAccessed" text="{{localize('sorting.sortLastAccessed')}}"></d2l-menu-item-radio>
+										<d2l-menu-item-radio class="dropdown-content-gradient" value="OrgUnitName" text="{{localize('sorting.sortCourseName')}}"></d2l-menu-item-radio>
+										<d2l-menu-item-radio value="OrgUnitCode" text="{{localize('sorting.sortCourseCode')}}"></d2l-menu-item-radio>
+										<d2l-menu-item-radio value="PinDate" text="{{localize('sorting.sortDatePinned')}}"></d2l-menu-item-radio>
+										<d2l-menu-item-radio value="LastAccessed" text="{{localize('sorting.sortLastAccessed')}}"></d2l-menu-item-radio>
```
dlockhart: These we just decided to revert to the default styles, which have been adjusted recently to be 0.8rem

PR #335, opened 2017-02-23
Clean up d2l-all-courses

ryantmer:
Been meaning to do this for a while, so [why not now](https://media.makeameme.org/created/Why-not-Zoidberg-vyotfu.jpg)? (Actual reason for now: tried to implement the two-call method, but this was just dizzying to constantly scroll through and try to find things. Now, it's organized! Yay!)

This is a strictly non-functional change, this component was just becoming rather unwieldly to work with. Broke the styles out into their own file, and rearranged all the functions and properties into more logical groupings (+ sub-grouped alphabetically yaaaay). Also renamed a couple private functions (so if that causes problems... they shouldn't be private...) and removed two mysterious `enrollment-pin/unpin-complete` listeners that don't exist anymore.

#sorrynotsorry

PR #336, opened 2017-02-24
Clean up d2l-my-courses

ryantmer:
Similar to #335, but for `d2l-my-courses.html` instead. Broken into another PR to not cause too much eye bleeding. The changes are are quite a bit less significant than the all-courses one; other files could be similarly organized I guess, but none are as large/complex, so probably not as important.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -53,6 +53,7 @@
 			url="[[enrollmentsUrl]]"
 			headers='{ "Accept": "application/vnd.siren+json" }'
 			on-iron-ajax-response="_onEnrollmentsRootResponse">
+
 		</d2l-ajax>
 
 		<d2l-ajax
@@ -116,7 +117,7 @@
 			with-backdrop>
 			<iron-scroll-threshold
 				id="image-selector-threshold"
-				on-lower-threshold="_changeImageLoadMore">
+				on-lower-threshold="_onChangeImageLowerThreshold">
 			</iron-scroll-threshold>
 			<d2l-basic-image-selector
 				image-catalog-location="[[imageCatalogLocation]]"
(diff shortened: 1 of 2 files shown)


d2l-my-courses.html
```diff
@@ -349,81 +441,20 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
+			},
+			//encapsulates root enrollment request, the debounce method loses context when directly using this
+			_rootRequest: function() {
+				this.$.enrollmentsRootRequest.generateRequest();
+			},
 			_showContent: function() {
 				this.toggleClass('d2l-my-courses-hidden', true, this.$.initialLoadingSpinner);
 				this.toggleClass('d2l-my-courses-hidden', false, this.$$('.my-courses-content'));
-			},
-			_setCourseImageEvent: function(e) {
```
jstefaniuk-d2l: So I personally don't see any value in alphabetically re-ordering all the functions, generally I'd rather have functions which may interact with each-other grouped together regardless or name (I see you've done other refactoring here based on public/ private, but this is just an in-general comment)

CodeBaboon: I agree, alphabetical deps makes sense to me but with functions I don't necessarily see the value. Public/Private holds a lot of value so that would definitely be a sort I would keep, but otherwise I agree that grouping based on some sort of other logical idea is more useful (things that interact, or things that do similar things, or whatever).

ryantmer: Yeah, makes sense. Alphabetical was mostly just laziness here, since it's easy. I'll re-move (not remove) them into a bit more logical order this afternoon.

ryantmer: Er, to clarify - are you saying the grouping into observers/listeners/etc. isn't worth it, or the alphabetical grouping within those categories isn't worth it?

CodeBaboon: I think observers, listeners, publics, privates, eventses, handlers... those sorts of things, that makes sense. Inside those I think it makes more sense to sort them in other ways if they need to be sorted at all.

jstefaniuk-d2l: I'd say play it by ear, if the file has 4 functions, then sorting by anything has no value, if it has a lot of stuff (like here) then can be some value

ryantmer: Yeah, makes sense - I've moved a few around in the new commit, although in this particular file most functions are fairly disconnected.

Side note: anybody know of an editor that supports ctrl+click for JS functions? 😢 (in C# I would never care about anything like this 😛)

ryantmer: Fair - "care less" in that case 😜 More that I wouldn't ever try to alphabetize a C# file, just "group by topic"-ize as Jeff initially suggested.

ryantmer:
Ah blerg - US84675 could do with this being in first - any objections to merging?

ryantmer:
Totally missed @jstefaniuk-d2l's :+1: on the comment above. Just checked this out, rebased it, and tested it, all seems good so going to merge.

PR #338, opened 2017-02-24
Fix resize on final course unpin

trevorjonesd2l:
@omsmith wrote this code


d2l-course-tile-sliding-grid-behavior.html
```diff
@@ -65,9 +65,10 @@
 				diff.count
 			).row;
 
-			if (diff.pos === oldTiles.length || diff.count > 1) {
-				// tiles were added/removed at the end, so only container needs to be resized
-				// also, don't animate anything other than pin/unpin
+			if (diff.count > 1 // only animate user's pin/unpin actions for now
+				|| diff.pos === oldTiles.length // tiles were added on to the end
```
omsmith: thinking about it, this could probably be `(insert && diff.pos === oldTiles.length)`, and similarly `!insert`, but seems like it should probably be okay without

PR #339, opened 2017-02-24
US83010 - Add error message when images aren't available

ryantmer:
Couple things here

- if no pinned courses, don't display the message on the home screen (only in the overlay)
- if > 0 pinned courses, display the messsage on the home screen and in the overlay
- _removeAlert wasn't actually working properly, so fixed it
- also slightly changed alert behaviour so that only one alert of a particular name can be shown at a time
- update tests
- remove _clearAlerts, as it's no longer really appropriate to use in any cases (should remove your own alerts)

TODO:

- [ ] Add unit test(s)

### DIFF
--- a/d2l-alert-behavior.html
+++ b/d2l-alert-behavior.html
@@ -24,6 +24,7 @@
 			},
 			_addAlert: function(type, name, message) {
 				if (this._alerts) {
+					this._removeAlert(name);
 					this.push('_alerts', {
 						alertType: type,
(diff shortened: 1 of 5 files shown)


ryantmer:
Yeah so I implemented this to check for the `organization-image`-rel Entity on the enrollments response, which is in the entirely-incorrect place... This US is being shelved anyway, so going to close this PR.

PR #348, opened 2017-03-07
DE24521 - Navigation of course tiles and course settings on screen readers

trevorjonesd2l:
Moved the template dom-if to the actual dropdown-menu so that the settings menu can be focused on safari


trevorjonesd2l:
Tested this myself on _Safari + VoiceOver_, _Chrome + NVDA_, _Chrome + JAWs_ and _IE + JAWs_.

PR #349, opened 2017-03-16
DE24833: Change the All Courses button to a link.

njostonehouse:


d2l-my-courses.html
```diff
@@ -30,25 +30,18 @@
...
-				cursor: pointer;
-				position: relative;
-			}
 			.my-courses-content {
 				padding-top: 10px;
 			}
 			button:hover, button:focus {
 				text-decoration: underline;
 			}
+			#all-courses-button {
```
ryantmer: Hm, should possibly be renamed to `allCourses` or something? Since it's no longer a button, although calling it `allCoursesLink` also seems somewhat confusing :/

d2l-my-courses.html
```diff
@@ -30,25 +30,18 @@
...
-				background: none;
-				border: none;
-				padding: 0;
-				cursor: pointer;
-				position: relative;
-			}
 			.my-courses-content {
 				padding-top: 10px;
 			}
 			button:hover, button:focus {
```
ryantmer: Ah, found the focus issue I was talking about - this rule can be removed since there are no longer any buttons anyway, but a similar one should be added for `<a>`s.

d2l-my-courses.html
njostonehouse: `all-courses-shower`?

d2l-my-courses.html
njostonehouse: [https://github.com/BrightspaceUI/link/blob/master/d2l-link.html#L25](This) doesn't carry over?

d2l-my-courses.html
ryantmer: Haha, definitely read that as like... `all-courses-`🚿 though... maybe `viewAllCourses` or something, since that's the text anyway :)

d2l-my-courses.html
ryantmer: Oh hm, good point... it seems like it partially does, hovering with my house does show the underline, just that keyboard focusing doesn't 😕

njostonehouse: Woah. That's no good. I will look into that.

dbatiste:
@njostonehouse : awesome!

One thing that's irking me is the fact that this has been different size/weight than the similar link for the announcements widget.  I think they should be consistent... not sure if this something you're interested in making consistent now...

I am guessing they should both be 14 or 16px, and normal font-weight, but I'm not sure.

d2l-my-courses.html
```diff
@@ -30,24 +30,14 @@
...
-				cursor: pointer;
-				position: relative;
-			}
 			.my-courses-content {
 				padding-top: 10px;
 			}
-			button:hover, button:focus {
-				text-decoration: underline;
+			#viewAllCourses {
+				float: right;
```
dbatiste: One thing to be careful of is clearing the float.  Usually there should be a sibling element that follows with `clear: both;` to clear the float.  Otherwise elements that follow may not be rendered below the floating elements.

ryantmer:
@dbatiste Tis done! :) Yeah the thought had come up earlier - but since this was a button, the different styling made a bit of sense. But now that it's a link, it looks the same as the "View All Announcements" link :D

dbatiste:
The float too was inconsistent I think?  Looks like you're making it consistent now.  :)

ryantmer:
Ah fair, hadn't noticed that one!

PR #354, opened 2017-03-21
DE24835: Only announce the course name when accessing the course link…

njostonehouse:
…s by giving the anchor an aria-label.


njostonehouse:
This introduces a host of regressions, unfortunately.

PR #356, opened 2017-04-03
US84675 - Remove d2l-siren-parser-ui

ryantmer:
Guess who's back with the brand new ~rap~ PR. It's My Courses!

Added the browserified node-siren-parser script to d2l-utility-behavior, as well as a _parseEntity function there. This pretty much replaces the functionality of the d2l-siren-parser WC, but without the performance overhead of the document.createElement calls.


d2l-search-widget-custom.html
```diff
@@ -365,21 +364,20 @@
...
 						// Fetch each search result's organization's information
-						(function(enrollment, index, numEntities, self, parser) {
+						(function(enrollment, index, numEntities, self) {
 							var ajax = document.createElement('d2l-ajax');
 							ajax.url = enrollment.getLinkByRel(/\/organization$/).href;
 							ajax.method = 'GET';
 							ajax.onResponse = function(response) {
 								if (response.detail.status === 200 || response.detail.status === 304) {
-									var organization = parser.parse(response.detail.response);
+									var organization = self.parseEntity(response.detail.response);
```
jstefaniuk-d2l: Whoa this one's not `response.detail.xhr.response`

ryantmer: Woah. Yeah what. Maybe the `.xhr` is superfluous? O_o

jstefaniuk-d2l:
Code LGTM :+1:

ryantmer:
Thanks! I'm self-testing this atm to make sure nothing broke, if somebody else wants to give a quick check, cool, if not then meh - any problems would likely be quite evident anyway (i.e. data that should show up wouldn't show up).

package.json
```diff
@@ -7,9 +7,8 @@
     "build": "cp d2l-*.html ./dist/",
     "postinstall": "bower install",
     "publish:cdn": "npm run build -s && frau-publisher",
-    "test": "npm run test:lint:js && npm run test:lint:wc && wct",
```
cgalvind2l: Why did you remove this linting?

jstefaniuk-d2l: I remember @ryantmer leaving a comment about this on another repo, forget which one though

ryantmer: Oh right, forgot to make that comment here as well - https://github.com/PolymerLabs/polylint/issues/141 is an open issue with Polylint where you can't have absolute URLs in any script, anywhere. Which is... pretty broken. It's also not overly useful, hence the seemingly-nonchalant removal of it.

I tried a few alternate methods, but ultimately you always end up with a `<script src="https://some-absolute-url.com"></script>` somewhere, and that breaks polylint :( We still have eslint though, and it's doing most of the work anyway. Polylint just adds a few things like calling out when you have undefined Polymer properties and such.

ryantmer:
Testing notes:

- [x] Basic fetch of enrollments works (enrollments + organization info)
- [x] All courses displays more enrollments
- [x] Lazy load works in all-courses
- [x] Can pin/unpin courses
- [x] Pinning a course in the Course Selector adds it to pinned courses in My Courses
- [x] Can search for courses
- [x] Live search results populate
- [x] Can sort courses
- [x] Can filter courses
- [x] Both simple and tabbed filter menus load entry items correctly
- [x] Lazy load works in tabbed filter menu
- [x] No console errors or anything
- [x] Check all these things in Chrome, FF, Edge, IE, Safari

So, barring any major issues that come up, I'll merge and release tomorrow morning!

PR #360, opened 2017-04-24
Remove forEach

ryantmer:
forEach is not supported in IE11, so any of these uses were erroring out. Surprised we didn't notice this earlier tbh.

TODO:

- [x] rebase once #359 is merged


ryantmer:
IE11 totally supports `forEach`, the actual problem here is that IE11 doesn't bother parsing `response.detail.xhr.response` (just leaves it as a string), whereas other browsers do. So, this isn't needed; I'll open a new PR with that fix.

PR #362, opened 2017-05-03
making course name wrapping consistent across all browsers

dlockhart:
This fixes a subtle issue (`DE25712`) with course name wrapping in Firefox, IE and Edge. This screenshot demonstrates it:

screen shot 2017-05-03 at 11 08 54 am

There are 2 important test cases here:
- "Coursewithalongnamethatwilltest the overflowing things": a long word without spaces that we'd ideally like it to break part-way through if necessary
- "Course3 - The Science of Nature": normal-length words that have lots of good break points

Notice how in the "Science of Nature" course, it's breaking after the "N" in "Nature", even though there was a perfectly good break point right before that.

After doing lots of research and trial & error, there are many ways to control wrapping/breaking in CSS, and each browser has varying levels of support for each of them:
- [`overflow-wrap`](https://developer.mozilla.org/en/docs/Web/CSS/overflow-wrap?v=example): this is the newest and preferred method -- it has "normal" and "break-word". Supported in Chrome, Firefox and Safari.
- [`word-wrap`](https://www.w3schools.com/cssref/css3_pr_word-wrap.asp): was replaced by `overflow-wrap` -- also has "normal" and "break-word". Supported in all browsers.
- [`word-break`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break): older, it has "normal", "break-word", "break-all". Firefox and IE/Edge only support "break-all". Chrome and Safari support both.

Basically, previously we had `word-break: break-all; word-break: break-word;`, which is no good because Firefox, IE and Edge don't understand "break-word" so will fall back to "break-all" -- which has the effect of breaking words partway through, hence the bug.

Instead, using a combination of `word-wrap: break-word` for IE/Edge and `overflow-wrap: break-word` for Chrome, Firefox and Safari does the trick!

Here's the result:

screen shot 2017-05-03 at 11 22 03 am

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -125,13 +125,12 @@
 				text-transform: uppercase;
 			}
 			.course-text {
-				position: relative;
 				margin-top: 0.6rem;
 				float: left;
 				flex: 1;
-				white-space: pre-wrap;
-				word-break: break-all;
-				word-break: break-word;
+				overflow: hidden;
+				word-wrap: break-word; /* IE/Edge */
+				overflow-wrap: break-word; /* replaces 'word-wrap' in Firefox, Chrome, Safari */
 			}
 			.course-info {
 				display: flex;


d2l-course-tile-styles.html
```diff
@@ -125,13 +125,12 @@
 				text-transform: uppercase;
 			}
 			.course-text {
-				position: relative;
```
dlockhart: I removed this because the `.course-text` element has no children that are positioned absolutely, so it was unnecessary.

d2l-course-tile-styles.html
```diff
@@ -125,13 +125,12 @@
...
 				margin-top: 0.6rem;
 				float: left;
 				flex: 1;
-				white-space: pre-wrap;
```
dlockhart: Also removed `white-space: pre-wrap`. Here's what that rule [is described as doing](https://developer.mozilla.org/en/docs/Web/CSS/white-space?v=example):

> Sequences of whitespace are preserved. Lines are broken at newline characters, at <br>, and as necessary to fill line boxes.

Pretty sure we don't want things breaking at newlines. Removing it had no noticeable effect.

d2l-course-tile-styles.html
```diff
@@ -125,13 +125,12 @@
...
-				word-break: break-all;
-				word-break: break-word;
+				overflow: hidden;
+				word-wrap: break-word; /* IE/Edge */
+				overflow-wrap: break-word; /* replaces 'word-wrap' in Firefox, Chrome, Safari */
```
dlockhart: To get the long-worded course name to break, I needed to add `overflow: hidden` -- otherwise it would overflow in a visible manor like this:

screen shot 2017-05-03 at 11 29 26 am

njostonehouse: Would `overflow: auto` cause problems (scroll bars)?

dlockhart: Yeah, you get hidden scrollbars on OSX and visible ones in Windows:

screen shot 2017-05-03 at 1 26 50 pm

PR #363, opened 2017-05-15
upgrading to app-localize-behavior to fix Safari 9 JS error

dlockhart:
This fixes [this issue](https://github.com/PolymerElements/app-localize-behavior/issues/74).


dlockhart:
Breaking because `course-image` and `image-selector` need to be bumped first. I'll get back to this shortly.

PR #365, opened 2017-06-29
switching to use the new homepage subentity, which has a relative path

dlockhart:
Corresponds with this PR in the monolith:
https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/5967/overview

Main motivation here is to use a relative path instead of one that includes the full hostname, since the hostname (which comes from the WebServerName config) may not be the same as the host the user is accessing the site from. Some clients have multiple vanity domains.

### DIFF
--- a/d2l-search-widget-custom.html
+++ b/d2l-search-widget-custom.html
@@ -93,7 +93,7 @@
 						<div class="search-results-listbox-page" data-page-name="search-results-page">
 							<d2l-search-listbox>
 								<template id="courseSearchResultsTemplate" is="dom-repeat" items="[[_liveSearchResults]]">
-									<div class="d2l-search-widget-custom-item" selectable data-text$="[[item.name]]" data-href$="[[item.href]]" role="option">
+									<div class="d2l-search-widget-custom-item" selectable data-text$="[[item.name]]" role="option">
 										[[item.name]]
 									</div>
 								</template>
@@ -381,8 +381,7 @@
 									// Polymer's splice method causes empty results to show up for an
 									// unknown reason, so use a normal Array.prototype.splice...
 									self._liveSearchResults.splice(index, 0, {
-										name: organization.properties.name,
-										href: (organization.getLinkByRel(/\/organization-homepage$/) || {}).href
+										name: organization.properties.name
 									});
 									// ...then, if this is the last result to get spliced in, manually
 									// call notifySplices with the changes.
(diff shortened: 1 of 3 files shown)


d2l-search-widget-custom.html
```diff
@@ -93,9 +93,7 @@
 						<div class="search-results-listbox-page" data-page-name="search-results-page">
 							<d2l-search-listbox>
 								<template id="courseSearchResultsTemplate" is="dom-repeat" items="[[_liveSearchResults]]">
-									<div class="d2l-search-widget-custom-item" selectable data-text$="[[item.name]]" data-href$="[[item.href]]" role="option">
-										[[item.name]]
-									</div>
+									<div class="d2l-search-widget-custom-item" selectable role="option">[[item.name]]</div>
 								</template>
```
dlockhart: @ryantmer I couldn't find any code that was actually using `data-text` or `data-href`, so I removed them entirely, as well as the code below that was populating them. I tested this out and things seemed to behave properly. Do you recall why it was originally needed?

ryantmer: They're HTML5 `data` attributes, so they aren't really what they appear to be (i.e. they're not anywhere as `dataText` or `dataHref`) - in this case they're being used [here](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-search-widget-custom.html#L418) to add the selected item to the search history by name [when it's selected](https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-search-listbox.html#L141). ([This page](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) explains the magic of `dataset` in particular - really it's just so we can arbitrarily tack on additional data to the event from the DOM element.)

d2l-search-widget-custom.html
```diff
@@ -381,8 +379,7 @@
 									// Polymer's splice method causes empty results to show up for an
 									// unknown reason, so use a normal Array.prototype.splice...
 									self._liveSearchResults.splice(index, 0, {
-										name: organization.properties.name,
-										href: (organization.getLinkByRel(/\/organization-homepage$/) || {}).href
```
ryantmer: If I'm following correctly, this means clicking a search result won't go directly to the course's homepage anymore correct?

Which... actually makes more sense, I think!

dlockhart: As we discussed offline, I couldn't find any reference to using this `data-href` attribute. I also couldn't get clicking the item in the search results to take me to the course -- so that functionality must have changed at some point in the past, making the `href` here unnecessary.

PR #370, opened 2017-08-01
US89271 - Updates (notifications) cleanup

ryantmer:
Removes the updates behavior and the `getOrgUnitId` utility function, then moves the fetching of the updates down into the course tile.

Also
- lint everything, not just the root
- remove non-working handler for `d2l-course-pinned-change`

TODO
- [x] Merge #371 
- [x] Merge #372 
- [x] Merge LMS-side changes


d2l-course-tile.html
```diff
@@ -470,25 +462,47 @@
...
+			},
+			_setCourseUpdates: function(updates) {
+				this.toggleClass('d2l-updates-hidden', updates <= 0, this.$.courseUpdates);
+
+				if (updates > 99) {
+					this._courseUpdates = '99+';
+				} else if (updates > 0 && updates <= 99) {
+					this._courseUpdates = updates;
+				} else {
+					this._courseUpdates = 0;
```
jstefaniuk-d2l: Is this `else` needed any more?

Maybe:
```
+				if (updates > 99) {
 +					this._courseUpdates = '99+';
 +				} else {
 +					this._courseUpdates = updates;
 +				}
```

Or ternary, even!

ryantmer: Truth! Didn't see the forest from the trees, apparently :P

ryantmer: Ah, seems like some tests actually rely on the more-explicit version of this. Could probably remove the tests as `updates` really shouldn't ever have a negative or null value, but eh I'll just leave it.

ryantmer:
YAY LMS-side changes have now been merged. This is ready for review/test, sorry for the delay!

d2l-course-tile.html
```diff
@@ -470,26 +462,40 @@
...
-						if (orgUnitId) {
-							var update = this.updateCountMap[orgUnitId];
-							if (update) {
-								this.setCourseUpdates(update);
-							}
-						}
-					}
+
+					var notificationsLink = organization && organization.getLinkByRel(/\/organization-notifications$/);
+					this._notificationsUrl = notificationsLink ? notificationsLink.href : null;
```
ctwomey1: Any chance a null notificationsurl will mess up _generateNotificationsRequest?

ryantmer: Wasn't actually sure, but the answer turns out to be [nope](https://github.com/PolymerElements/iron-ajax/blob/master/iron-ajax.html#L569)! :)

ryantmer: WAIT that's for when you're using `auto`... doesn't look like the same check is done when you call `generateRequest` directly, so I guess it's up to XMLHttpRequest?

ctwomey1: could maybe just wrap that small block in an if and not call a function we know won't do anything anyways

ryantmer: Just verified that it does loudly throw some shade, so will indeed so. Good catch!

ctwomey1: yeah so this line will set url to '' https://github.com/PolymerElements/iron-ajax/blob/master/iron-ajax.html#L388 I think.

Testing an xhr request in a browser:

`function reqListener () {
  console.log(this.responseText);
}

var oReq = new XMLHttpRequest();
oReq.addEventListener("load", reqListener);
oReq.open("GET", '');
oReq.send();`

Seems to give you the default new tab page in chrome, so you do get a response.  If you do the above request with null it cries about invalid url

ryantmer: Hehe, I [recognize that code](https://developer.mozilla.org/en/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest) 😉 Yeah I'll just make it a "truthy" check sorta thing, since requesting an empty URL is kinda pointless too.

ryantmer: Added the check for both the notifications and organization URL in 5ab5f86

PR #371, opened 2017-08-02
Refactor d2l-course-management-behavior

ryantmer:
The vast majority of the functionality contained within this behavior was either only used in one of d2l-my-courses/d2l-all-courses, or it was overridden in one or both anyway. So, moved things up a level to make it clearer (aside from the few things that actually _are_ common to both).

Found this while doing #370, but twas unrelated so gets its own PR. Tested it and all seems to work - pinning/unpinning courses moves the tiles as expected. If this didn't work, that wouldn't work, so seems like adequate testing to me 😛

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -12,6 +12,7 @@
 <link rel="import" href="../d2l-simple-overlay/d2l-simple-overlay.html">
 <link rel="import" href="d2l-alert-behavior.html">
 <link rel="import" href="d2l-all-courses-styles.html">
+<link rel="import" href="d2l-course-management-behavior.html">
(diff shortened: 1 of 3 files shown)


ryantmer:
Good point! I quickly looked at a couple, but I'll look deeper and see if there are more duplicates. I think a lot of this stuff got pulled out into behaviors weeks or months after it was first written, so very likely some things got missed!

ryantmer:
Fun fact, turns out the function I had initially de-duplicated is like... the only one that actually makes sense to have in the behavior. Everything else is either only used in one of d2l-my-courses/d2l-all-courses, or it's defined in the behavior and overridden in one of both of the above, so I moved things around so that things are exactly as specific as they need to be!

PR #372, opened 2017-08-02
Remove unused event stuff

ryantmer:
Requires #371.

Like #371, I came across this while doing #370, but it was unrelated. Soooo, different PR! Yay small safe changes. This doesn't work anymore because of the enrolledUser changes - we can't create/directly link to an enrollment, so removing this. No immediately-obvious way to fix this unfortunately...

TODO
- [x] Merge #371 and rebase this

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -221,19 +221,14 @@
 			],
 			ready: function() {
 				this._updateEnrollmentAlerts(this._hasEnrollments, this._hasPinnedEnrollments);
-				this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
 				this.toggleClass('d2l-my-courses-hidden', true, this.$.courseTiles);
 				this.toggleClass('d2l-my-courses-hidden', true, this.$.courseList);
 			},
 			attached: function() {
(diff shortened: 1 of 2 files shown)


jstefaniuk-d2l:
image

Hmm, only one codeowner? XD

ryantmer:
Yeah, not sure what's up with that...

jstefaniuk-d2l:
I guess github likes it highlander-style

ryantmer:
Spoke to Owen, should be a fairly easy way to re-introduce this event, but it'll be fairly different, so I think it still makes sense to remove the old stuff, make the change (in the LMS), then add new (simpler) stuff to handle it here.

ryantmer:
Oh, no the scope is less than that - yeah sorry, I really didn't describe what's actually going down here.

When you pin/unpin a course from the course selector in the minibar, it generates this event, which we listen for and update our list accordingly, same for vice versa. So all this change really does is means that when you pin/unpin something in the widget, the change won't appear in the course selector until you refresh (again, or vice versa). It's a pretty small use case either way.

ryantmer:
Also it's not so much "break in master" in this case as it is "remove the code for the thing that's broken anyway" haha 😝

PR #374, opened 2017-08-03
Use Polymer CLI including polymer lint and fix the things it complains about

ryantmer:
Currently includes the changes for #372 and #370, so the diff is nonsense atm don't waste your time here.

TODO

- [x] Merge #372 and rebase
- [x] Merge #370 and rebase


d2l-interaction-detection-behavior.html
```diff
@@ -11,7 +11,6 @@
 		* Behavior that detects whether an interaction occurs via touch or hover
 		* @polymerBehavior window.D2L.MyCourses.InteractionDetectionBehavior
 		*/
-
```
ryantmer: This isn't just picky whitespace - it actually causes a lint error claiming it can't find the `@polymerBehavior`.

localize-behavior.html
```diff
@@ -16,14 +16,8 @@
...
 
 	(function() {
-		window.D2L = window.D2L || {};
-		window.D2L.MyCourses = window.D2L.MyCourses || {};
-
-		/*
-		* Behavior for elements that require localization, based on `app-localize-behavior`
-		* @polymerBehavior window.D2L.MyCourses.LocalizeBehavior
-		*/
-		window.D2L.MyCourses.LocalizeBehaviorImpl = {
```
ryantmer: Functionally equivalent, but using the `var` pollutes the global namespace a bit less, I guess, whatever, don't judge me.

polymer.json
```diff
@@ -0,0 +1,8 @@
+{
+  "lint": {
+    "rules": ["polymer-2-hybrid"]
+  },
+  "sources": [
+    "**/*.html"
```
ctwomey1: maybe instead of including everything ever in this repo we move things to src so it follows polymer defaults?

ryantmer: Hmm, fair point. Would be a much larger change though, okay "for now"*?

*where for now is who knows how long (although adding an .html file in the root that shouldn't get linted sounds fairly unlikely anyway)

ctwomey1: doesn't polymer.json also affect how polymer builds things?  using sources would include everything ish?

ryantmer: It does, although we're not actually using polymer build here - we just [copy things] (My Couses all came about before we really started using the polymer build stuff).

ctwomey1: Maybe too much scope to add, but should definitely be an issue?

ryantmer: I may or may not have a secret after-hours branch with some delicious candy you'll enjoy 😄 

But yeah, good call :)

ryantmer: #375 is now a thing!

PR #377, opened 2017-08-30
Add DE and NB to localize-behavior

ryantmer:
I added these a while ago, but they weren't working since I didn't add them here...


Surekh:
@ryantmer - Is there a need to add them here - https://github.com/Brightspace/d2l-my-courses-ui/blob/2df3204ff8a1ece7c1b2b8dc96efc98aa36f05ce/localize-behavior.html#L36?

ryantmer:
Arg, there sure is! Good catch.

Surekh:
LGTM! But my review doesn't matter :( . I will let the in-office code owner @apalaniuk approve. :)

PR #380, opened 2017-09-08
DE27525 Fix unpinning enrollment in main screen

ryantmer:
This issue was actually two problems - first, the console error was happening as we changed to the enrolledUser HM API a while back, but that broke the extraction of the org unit ID here. Switched to use the organization instead.

Second, the animation was failing to complete. This I cannot explain, as there was never a handler for this event here as far as I can recall, but adding the handler does fix it. Not sure how it was working before...

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -214,7 +214,8 @@
 				'clear-image-scroll-threshold': '_onClearImageScrollThreshold',
 				'd2l-simple-overlay-closed': '_onSimpleOverlayClosed',
 				'enrollment-pinned': '_onEnrollmentPinAction',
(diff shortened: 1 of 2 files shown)


omsmith:
https://github.com/Brightspace/d2l-my-courses-ui/commit/a739ea7f3233c3f901846b6076d2cd55a5263b8f#diff-1b7e12ef66e494afa4279da5b61b708aL111

ryantmer:
Ah! Good sleuthing, all makes sense now.

dlockhart:
@apalaniuk, @CodeBaboon, @jstefaniuk-d2l @trevorjonesd2l could we get another approve on this? Sorry trying to get this in this afternoon.

PR #382, opened 2017-09-25
US90475 - Use d2l-fetch in favour of d2l-ajax

ryantmer:
This seems to improve our loading time locally by about 1 second. There is about a 1-second gap between the initial loading the page and when we fetch the token using `d2l-ajax`, but with `d2lfetch` this gap is all but gone.

Before - note the gap between the `Framework.UserProfileBadge` and `token` requests, resulting in the enrollments request happening around the 2.2s mark:

image

After - note the lack of gap between the `Framework.UserProfileBadge` and `token` requests, resulting in the enrollments request happening around the 1.2s mark:

image

Actual results may vary, but using d2lfetch also improves our code maintainability/cleanliness as it allows for more promisey structures in fetching data.

TODO:

- [x] Merge #387 and rebase this
- [x] Merge and release https://github.com/Brightspace/d2l-search-widget-ui/pull/32
- [x] Remove d2l-ajax from remaining components
- [x] Add/update/remove tests as required

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -107,6 +107,8 @@ the user has in that organization - student, teacher, TA, etc.
 	<script>
 		'use strict';
 
+		/* global Promise */
+
 		Polymer({
 			is: 'd2l-course-tile',
 
@@ -234,7 +236,7 @@ the user has in that organization - student, teacher, TA, etc.
 					body[field.name] = field.value;
 				});
 
-				window.d2lfetch
+				var promise = window.d2lfetch
 					.fetch(new Request(pinAction.href, {
 						method: pinAction.method,
 						body: new URLSearchParams(body),
@@ -250,8 +252,7 @@ the user has in that organization - student, teacher, TA, etc.
 						this.enrollment = enrollment;
 						this.pinned = this.enrollment.hasClass(this.HypermediaClasses.enrollments.pinned);
 					}.bind(this))
-					.catch(function(e) {
-						console.error(e, body);
+					.catch(function() {
 						// Just revert back to whatever pin state we already had stored
 						this.pinned = this.enrollment.hasClass(this.HypermediaClasses.enrollments.pinned);
 					}.bind(this));
@@ -279,6 +280,8 @@ the user has in that organization - student, teacher, TA, etc.
 				}, { bubbles: true });
 
 				this._pendingPinAction = this.pinned;
+
+				return promise;
 			},
 			// Sets classes for this tile having the touch menu on it when it is opened/closed
 			setTouchMenuOpen: function(isOpen) {
@@ -329,16 +332,14 @@ the user has in that organization - student, teacher, TA, etc.
 							'pragma': 'no-cache'
 						}
 					}))
-					.then(this.responseToSirenEntity.bind(this))
-					.then(this._onOrganizationResponse.bind(this));
+					.then(this.responseToSirenEntity.bind(this));
 			},
 			_fetchNotifications: function() {
 				if (!this._notificationsUrl || !this.courseUpdatesConfig) {
 					return;
 				}
 
-				return this.fetchSirenEntity(this._notificationsUrl)
-					.then(this._onNotificationsResponse.bind(this));
+				return this.fetchSirenEntity(this._notificationsUrl);
 			},
 			_enrollmentChanged: function() {
 				this.pinned = this.enrollment.hasClass(this.HypermediaClasses.enrollments.pinned);
@@ -350,14 +351,18 @@ the user has in that organization - student, teacher, TA, etc.
 
 					if (!this.delayLoad) {
 						return this._fetchOrganization()
-							.then(this._fetchNotifications.bind(this));
+							.then(this._onOrganizationResponse.bind(this))
+							.then(this._fetchNotifications.bind(this))
+							.then(this._onNotificationsResponse.bind(this));
 					}
 				}
 			},
 			_delayLoadChanged: function(delayLoad) {
 				if (this.isAttached && !delayLoad && !this._organization) {
 					return this._fetchOrganization()
-						.then(this._fetchNotifications.bind(this));
+						.then(this._onOrganizationResponse.bind(this))
+						.then(this._fetchNotifications.bind(this))
+						.then(this._onNotificationsResponse.bind(this));
 				}
 			},
 			_displaySetImageResult: function(success) {
@@ -463,6 +468,8 @@ the user has in that organization - student, teacher, TA, etc.
 
 				var notificationsLink = organization && organization.getLinkByRel(this.HypermediaRels.organizationNotifications);
 				this._notificationsUrl = notificationsLink ? notificationsLink.href : null;
+
+				return Promise.resolve();
 			},
 			_onNotificationsResponse: function(notifications) {
 				var total = 0;
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -1,5 +1,4 @@
 <link rel="import" href="../polymer/polymer.html">
-<link rel="import" href="../d2l-ajax/d2l-ajax.html">
 <link rel="import" href="../d2l-course-image/d2l-course-image.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown.html">
 <link rel="import" href="../d2l-dropdown/d2l-dropdown-menu.html">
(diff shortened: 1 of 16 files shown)


d2l-my-courses.html
```diff
@@ -222,12 +211,16 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 				'_updateEnrollmentAlerts(_hasEnrollments, _hasPinnedEnrollments)'
 			],
 			ready: function() {
+				window.d2lfetch.use({name: 'auth', fn: window.d2lfetch.auth});
+				window.d2lfetch.use({name: 'dedupe', fn: window.d2lfetch.dedupe});
+				window.d2lfetch.use({name: 'simpleCache', fn: window.d2lfetch.simpleCache});
```
CodeBaboon: I wonder if this is the right place for registering the middleware, given the grander context of the BSI and the LMS. Not sure of the answer, though :S

ryantmer: Hmmm, good point. I'll give that some further thought; does `d2lfetch` replace middlewares if a duplicate name is defined, or ignore the replacement?

d2l-course-tile.html
```diff
@@ -235,18 +234,27 @@
...
-				this._enrollmentPinRequest.headers = {
-					'accept':'application/vnd.siren+json',
-					'content-type':'application/x-www-form-urlencoded'
-				};
-				this.listen(this._enrollmentPinRequest, 'iron-ajax-response', '_onEnrollmentPinResponse');
-				this.listen(this._enrollmentPinRequest, 'iron-ajax-error', '_onEnrollmentPinError');
-
-				this._enrollmentPinRequest.generateRequest();
+				window.d2lfetch
+					.fetch(new Request(pinAction.href, {
```
jstefaniuk-d2l: is Request polyfilled? (It seems to lack some support) https://developer.mozilla.org/en-US/docs/Web/API/Request

ryantmer: It's not yet, nor is `Promise` afaik. Planning to look into those today, especially since I think we're still supporting IE11 in My Courses 😱

d2l-my-courses.html
ryantmer: Duly noted! This actually seems like a case where a `console.warn` might be warranted (trying to add a middleware that already exists), but I'll think about it a bit more. And probably bug you about it a bit more 😛

CodeBaboon: Yeah, we knew this aspect wasn't fully fleshed out, but basically deferred a solution until we had a specific use case. Looks like this might be it! We can definitely put in betterer handling for you here.

ryantmer: I shall give it some thought (by which I mean ask smarter people than I for their opinions) and see what comes up!

ryantmer:
This isn't passing CI due to https://github.com/Brightspace/d2l-search-widget-ui/pull/32 not yet being merged/released, but this and that are now both ready for review/testing :D

d2l-my-courses.html
ryantmer: More thought was given! Right now, this is the only place we're using d2lfetch in the LMS as far as I know, so I think keeping here for now is okay. However, I do have some further plans to land and expand with d2lfetch which will likely require moving its inclusion to BSI; so, I'll look back into it then if that's cool?

CodeBaboon: I imagine at some point soon we should be making use of it in the banner overlay, so we're just kicking this can down the road a short ways and possibly creating bugs/issues if someone comes in without this prior knowledge and just drops new usages in place somewhere else that clash with this.

ryantmer: That's exactly the further plans, in fact! :) Hoping to start on that today, but if it's okay with you I'd rather solve it as part of that? (Just to prevent blocking this change, which is intended to ship next week.)

CodeBaboon: Sounds reasonable enough. APPROVED!

ryantmer: Super-duper promise I'll address it within the next week! :D

CodeBaboon: Alright, only because you super-**duper** promised :P

d2l-course-tile.html
```diff
@@ -108,6 +107,8 @@
 	<script>
 		'use strict';
 
+		/* global Promise */
```
dlockhart: Nit: could throw this in `.eslintrc.json`

d2l-course-tile.html
```diff
@@ -438,77 +445,58 @@
...
 			},
 			_setCourseUpdates: function(updates) {
 				this._courseUpdates = updates > 99 ? '99+' : Math.max(updates, 0);
 				this.toggleClass('d2l-updates-hidden', updates <= 0, this.$.courseUpdates);
 				this.$$('#courseUpdates').setAttribute( 'aria-hidden', updates === 0 ? 'true' : 'false' );
 			},
-			_checkDateBounds: function(organization, response) {
-				var serverDate = Date.parse(response.detail.xhr.getResponseHeader('Date'));
+			_checkDateBounds: function(organization) {
+				var serverDate = Date.now();
```
dlockhart: Hmm, this isn't always going to result in the same value as you had before. You're now relying on the user's browser's clock to define "now" vs. before it was what the server's clock thought "now" was. Best case, it's likely off by seconds or minutes, but if the user-computer timezone differs from what they've set up in their account settings in the LMS, it could be off by many hours.

dlockhart: Looking at the existing code, using the HTTP response header isn't amazing either, since I believe that's just going to be the server's timezone's "now", not the user's. A proper solution would involve including the server's "now" converted to the user's timezone and comparing that to the organization start/end dates (which should also be in the user's timezone).

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
```diff
@@ -14,7 +13,8 @@
 	<template>
 		<style>
 			:host {
-				display: block;
+				display: flex;
+				flex-direction: column;
```
dlockhart: Are any vendor prefixes needed for `flex` and `flex-direction`? Specifically thinking about IE11....

d2l-course-tile.html
ryantmer: Wow, totally missed that. Good call! I could use `Date.UTC()` instead in that case? The start/end date coming back from the server should be in UTC as well, so as long as we're after start and before end, then we should be good?

d2l-my-courses.html
```diff
@@ -129,9 +114,15 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 
 	</template>
 
+	<script src="https://s.brightspace.com/lib/d2lfetch/1.2.0/d2lfetch.js"></script>
+	<script src="https://s.brightspace.com/lib/d2lfetch-auth/0.6.0/d2lfetch-auth.js"></script>
+	<script src="https://s.brightspace.com/lib/d2lfetch-dedupe/0.10.0/d2lfetch-dedupe.js"></script>
+	<script src="https://s.brightspace.com/lib/d2lfetch-simple-cache/0.3.0/d2lfetch-simple-cache.js"></script>
```
dlockhart: Once this gets used in 2 places, we'll probably want to include `.html` imports similar to what we did for [d2l-performance](https://github.com/Brightspace/d2l-performance-ui). That way we can de-dup and be confident that anything pulled into BSI that uses `d2l-fetch` is using the same version.

polyfill/Array.prototype.includes.js
```diff
@@ -0,0 +1,54 @@
+// Required to use d2l-fetch-simple-cache
```
dlockhart: Again, once `d2l-fetch` is used in 2 places, this likely belongs in BSI.

d2l-my-courses.html
ryantmer: For sure! Working on that right now in fact, as I'm converting `d2l-image-selector` to use d2l-fetch as well :)

d2l-filter-menu-content/d2l-filter-menu-content-tabbed.html
ryantmer: I did test it out on IE11 and it appears to work; caniuse indicates that there are a lot of bugs, granted, but I guess this is a simple enough case? 😕

dlockhart: Cool yeah as long as it works, I'm good with no prefixes. :)

d2l-course-tile.html
dlockhart: I don't think ever using anything on the client can be trusted when comparing with anything from the server. You need to either always compare 2 client-side dates or 2 server-side dates.

If we're displaying them to the user, dates coming back from the server should not be UTC and should be converted to the user's time zone. Not sure if that's actually happening though in the HM responses, and they're typically UTC?

ryantmer: Yeah, in this case it's whether to show a course as ended/started so I wasn't really considering it a high-risk case, but it could be confusing if that got out of line. But yeah, HM responses are UTC - should this be changed so that it's converted to the time zone of the caller?

dlockhart: If we're actually displaying the date to the user (e.g. "This course starts tomorrow at 3pm"), then that needs to be in their time zone, not UTC. If we're just using it to see if it's "in the future" or "in the past", then we can keep it UTC as long as we're comparing it to "now in UTC". But that "now in UTC" value should probably be coming from the server, not the client, since those times are likely off by minutes.

ryantmer: Hm, we're display just the date ("Course Starts September 1st, 2017"), so UTC comparison should be okay? I'll update it to use the server time again, though!

PR #383, opened 2017-09-26
US90690 Remove 7 item logic

ryantmer:
This removes the `d2l-filter-menu-content-simple` component (which was the one used for >=7 semesters/departments), in favour of always using the tabbed view regardless of how many semesters/departments the user has.

Removing that component meant the `d2l-filter-menu-content` component could also be removed, as its purpose was really to just decide which view (simple vs. tabbed) to show in the filter menu, and relay some events between `d2l-all-courses` and whichever content component was being used. Now, we can just use the `d2l-filter-menu-content-tabbed` directly. I also cleaned up some other related stuff, like removing `loadMore` et al., as these are no longer required as of #379 it turns out! :) 

I will create a follow-up PR to rename `d2l-filter-menu-content-tabbed` to just `d2l-filter-menu-content` (since it's now the only one), but doing this in the same PR made the diff suuuuper nasty, whereas a rename would look much nicer.


ryantmer:
Note: moved from a task under the previous US into its own, US90690.

ryantmer:
Closing this in favour of multiple smaller PRs, since this is... kinda three changes in one.

PR #384, opened 2017-09-26
US89544 - Allow image upload callback to be provided

capajon:
Allows a new course image upload callback to be provided that is passed along to the course image tiles.

After the image is uploaded courseImageUploadCompleted should be called.  Currently there is some outstanding work to be done to refresh the image of the course in question.  I will follow up with this shortly.  It should be very similar to how it is done in the course image banner (ie call out to the organization's api, then the image api, and update the image src with a date time stamp appended as a #).

Focus is also a new function that can be called.  It will try to focus the last "modified" course tile, and if not that, then the all courses link.


capajon:
This is failing because an update to image-selector is not updated yet so I'll need that to merge first basically (https://github.com/Brightspace/image-selector).

Related LMS PR (has links to all web component PRs, etc): https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/1781/overview

Happy to discuss in person to give more context.

d2l-course-tile-grid.html
```diff
@@ -121,6 +121,24 @@
 					selectedTile.setCourseImage(detail.detail);
 				}
 			},
+			focus: function(organization) {
```
dbatiste: I think I may be inclined to call this `tryFocus` or `tryFocusCourse` or `tryFocus...`.  It makes the consuming code a little clearer.  I don't feel super strongly about it though, since native `focus` method may also not apply focus in certain cases (example, element cannot be focused).

capajon: Yeah I agree.  But the main reason I kept it as just focus is that the LMS dialog function wants to take in an element to call focus on when it is done.  So .focus just conveniently works with that.  Otherwise I might have had to create a "redirector" div or something for the dialog function to call (I know we discussed that at one point but it ended up being easier to do this).

d2l-my-courses.html
```diff
@@ -237,10 +240,32 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 			* Public API functions
 			*/
 
+			courseImageUploadCompleted: function(success) {
+				if (success) {
+					this.$['basic-image-selector-overlay'].close();
+					// TODO: refresh image for this._setImageOrg
```
ryantmer: Is it necessary to refresh the image within `_setImageOrg`? It's re-set whenever the overlay opens, so wouldn't it already be updated when the change image overlay is re-opened anyway?

d2l-my-courses.html
```diff
@@ -237,10 +240,32 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
+					return;
+				}
+				this.$.viewAllCourses.focus();
+			},
+
 			getCourseTileItemCount: function() {
 				return this.pinnedEnrollments.length;
 			},
 
+			getLastOrgUnitId: function() {
```
ryantmer: Is this being called from the upload image dialog? Looks like it's not used here (but this time, I noticed it was a public method 😜)

d2l-my-courses.html
```diff
@@ -509,6 +533,14 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 			_showContent: function() {
 				this.toggleClass('d2l-my-courses-hidden', true, this.$.initialLoadingSpinner);
 				this.toggleClass('d2l-my-courses-hidden', false, this.$$('.my-courses-content'));
+			},
+			_getOrgUnitId: function(organization) {
```
ryantmer: I realize this was already here, and untested, but mind making a sanity test or two for it? It makes me nervous to be parsing org unit IDs out of entities as it's already caused us problems once 😢

d2l-course-tile.html
```diff
@@ -218,6 +218,9 @@
 			focus: function() {
 				this.$['d2l-course-tile-anchor'].focus();
 			},
+			getOrganization: function() {
```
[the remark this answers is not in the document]
capajon: Yeah I should really either just use _organization or switch everything.  I didn't notice it was extensively used like that. I'll do one or the other!

d2l-my-courses.html
capajon: I this case courseImageUploadCompleted will be called by the LMS after it is done uploading through the dialog it creates (where you can chose from my computer and LMS sources like locker).  So it needs to happen before the image overlay is re-opened.

ryantmer: Gotcha, makes sense!

d2l-my-courses.html
capajon: Haha, yeah it is used by the LMS.  It is a little smelly.

Basically after the user goes through the upload dialog (which is completely 'controlled' by the LMS) it needs to know what orgunitid to upload the image to (creating the rpc route, etc).  So this is how it does it, because otherwise it doesn't know what org the user last interacted with.

You might say, but Jon, why not just make a parameters on the courseImageUploadCb so we can just feed it directly to the upload function and not have this weirdness, and to that I would say, yes, I would prefer to do that.  But I never figured out a way to do that in the LMS, as astoundingly simple as you might expect that to be.  Part of it is that I was reusing the file upload dialog that already exists in the LMS, and it wasn't really built for that.  Part of it might just be that I didn't fully understand our javascript pipeline and other such things.... but this wasn't throw up in my mouth bad.  Just a bit smelly.

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -314,6 +314,94 @@ describe('d2l-my-courses', function() {
 			expect(widget._alerts).not.to.include({ alertName: 'setCourseImageFailure', alertType: 'warning', alertMessage: 'Sorry, we\'re unable to change your image right now. Please try again later.' });
 		});
 
+		describe('coruse image upload', function() {
```
ryantmer: Ah, I missed this - typo!

d2l-my-courses.html
capajon: Sure. :)

test/d2l-my-courses/d2l-my-courses.js
capajon: blah I have done that so many times now!  And funtcion and other various on function that aren't quite function.  At least that gets caught right away.  Will...adjust.

d2l-my-courses.html
ryantmer: "It works" and "not throw up in your mouth bad" is good enough for me, then! 😆

test/d2l-my-courses/d2l-my-courses.js
ryantmer: Oh - also these tests are doubled! This entire `describe` block is here twice with the same content, I do believe. At the very least that's causing the number of failed tests to be doubled 😉

d2l-course-tile-grid.html
dougmoscrop: seems like you could move the org identifier part out of the loop, and it's a `return courseTiles.some(...)` at that point. (can also short circuit evaluating the tiles if no org id at that point too)

d2l-my-courses.html
capajon: Missed this - if it is okay I'll follow up with some tests on this along with my hoped for firefox fix on Travis in separate PR (I still have to do the image refresh too so I'm not done adding tests etc).

PR #390, opened 2017-10-03
US90475 PR follow-up feedback

ryantmer:
Addresses some additional feedback from https://github.com/Brightspace/d2l-my-courses-ui/pull/382

- Move Promise to eslintrc
- Use UTC date for comparison, since that's what comes back from the server

Doesn't include the `d2l-fetch` -> BSI changes, as I'm still working on those, but figured I should get these in as they're small, and independent of that change.

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -107,8 +107,6 @@ the user has in that organization - student, teacher, TA, etc.
 	<script>
 		'use strict';
 
-		/* global Promise */
-
 		Polymer({
 			is: 'd2l-course-tile',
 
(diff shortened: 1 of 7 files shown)


ryantmer:
Yeah, I've got an idea for that one, but it requires a bit more re-jiggering of response/promise structures, so I'll take another crack at that separately :)

d2l-course-tile.html
```diff
@@ -499,14 +497,14 @@
 				this.$$('#courseUpdates').setAttribute( 'aria-hidden', updates === 0 ? 'true' : 'false' );
 			},
 			_checkDateBounds: function(organization) {
-				var serverDate = Date.now();
+				var nowDate = Date.UTC();
```
dlockhart: Not sure if this is quite what you want. `Date.UTC()` is [going to be a timestamp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC), not an actual date object.

dlockhart: Hmm, although `Date.now()` is also a timestamp...

ryantmer: Turns out it failed the tests anyway! Looking into it.

ryantmer: Looks like `Date.now()` is actually UTC anyway, so I'll change this back since pretty sure both would have the same problems. I'll make a note to do it properly with the server timestamp, though.

PR #391, opened 2017-10-03
Use d2l-fetch from bower, to take advantage of version resolution

ryantmer:
Under this new method, `.use`ing the various middlewares is done by the consumer of the element, in this case BSI. This means we can get the same versions for everything, and the same middlewares. In addition, My Courses shouldn't really _care_ how the d2l-fetch sausage is made, only that it gets made somehow.

Unclear whether I really need to have the HTML import at all in My Courses, but hopefully if nothing else it makes it more obvious that it's not just an unused bower reference.

TODO

- [x] Merge and release https://github.com/Brightspace/d2l-fetch/pull/10
- [x] Update this to point at actual versions, not a branch

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -35,7 +35,7 @@
     "d2l-fetch": "Brightspace/d2l-fetch#^1.5.1",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^2.4.1",
     "d2l-icons": "^3.2.0",
-    "d2l-image-selector": "git://github.com/Brightspace/image-selector#^1.1.0",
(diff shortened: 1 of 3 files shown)


ryantmer:
Yay, finally ready to merge. Note that this will be a major version bump, as it changes how My Courses fetches data (it now requires BSI set up the various `d2l-fetch` middlewares, most notably `d2l-fetch-auth`, for it to work).

Verified that this works by linking it, and using https://github.com/Brightspace/brightspace-integration/pull/656. This results in us pulling a single version of `d2l-fetch` and each of its middlewares, and all things continue to work!

PR #395, opened 2017-10-04
documenting some of the existing telemetry the component uses

dlockhart:
Going to expand this out as I add in performance telemetry, but wanted a list of what we already track.


README.md
```diff
@@ -64,6 +64,30 @@ To lint and run unit tests, run:
...
+
+### My Courses Telemetry
+
+| Event | Description | Initiators | Parameters |
+| ----- | ----------- | ---------- | ---------- |
+| LaunchChangeImage | Fired when the "change image" menu item is clicked from a course tile. | `<d2l-course-tile>` | |
+
+### Change Image Overlay Telemetry
+
+| Event | Description | Initiators | Parameters |
```
ryantmer: The info in `initiators` is maybe redundant? Not really a problem, just makes the table more squished.

dlockhart: Agreed, killed it.

PR #396, opened 2017-10-05
Fix empty results errors

ryantmer:
Three small problems fixed here:

- If a search term returned no results in the live search, we were getting an error when trying to get .length of the returned entities
- If the results of a search are empty, don't show the "you have no semesters message" and hide the search widget
- If a search term returns no entries, show "No results" message (this bug has existed forever, figured might as well fix it while I'm in there)


ryantmer:
🙏

PR #397, opened 2017-10-06
Rename filter menu files

ryantmer:
There's no longer a simplified view of this, so renaming d2l-filter-menu-content-tabbed to just d2l-filter-menu. This is also in preparation for more incoming changes to the structure of this menu (to allow for adding the role filter tab).

(Note that this is a straight-up rename, no functional changes or anything.)


ryantmer:
Turns out my current (post-this-PR) rework is going to be more involved than expected, so this isn't needed anymore.

PR #399, opened 2017-10-10
Performance timings

dlockhart:
This adds marks and measures for various performance timings in my-courses. Initially these will only be available in the browser console, but the hope is in the short-term to find somewhere to log these to.

From this we get what I feel like are the important measures:
1. Time to complete "root enrollment" API call (often unfortunately including fetching the token)
2. Time to complete the "enrollment search" API call
3. After the search comes back, time until each course tile has rendered once (probably just a skeleton as the organizations calls haven't completed yet)
4. After the search comes back, time until each organization API call has completed and each tile has re-rendered (so probably displaying the course name now)
5. After the first organization API call happens, time until all the course images have finished downloading
6. Finally, from the initial "attach" call to the last course image loads is a measure that tries to capture the "total my course load time"

Possible future enhancements:
- A way to capture time to fetch the token, probably part of `d2l-fetch-auth`
- A way to factor out the token-fetch time from the root enrollment API call, since it's not really its fault
- Send these measures somewhere where we can track them over time. Currently looking at some of the infrastructure Littlefoot has.


d2l-course-tile.html
```diff
@@ -447,7 +448,12 @@
 				}
 			},
 			_onOrganizationResponse: function(organization) {
+
 				this._organization = organization;
+				Polymer.RenderStatus.afterNextRender(this, function() {
```
ryantmer: Is doing this `afterNextRender` just to avoid causing any further perf impact?

dlockhart: No, it's to more accurately represent the time at which the organization information (course name, etc.) actually _rendered_ to the user, as opposed to when the API call returned.

PR #400, opened 2017-10-11
US90997 Rework filter menu

ryantmer:
The previous architecture of this filter menu was highly tailored to just having the two tabs, semesters and departments. Trying to shoehorn the upcoming role filtering tab into it would have been quite messy, so instead it's been refactored into something a bit more generic - there's still a bit of semester/department knowledge in the new d2l-filter-menu that'll have to be expanded to include roles, but it's much less than there previously was; in addition, a good chunk of the common logic was factored out into the d2l-filter-menu-tab component. Using iron-pages to switch between them also means added a new d2l-filter-menu-tab will be fairly easy (although roles may have to have their own component, since they're filtered a bit differently than semesters/departments, but either way - much simpler this way).

TODO:

 - [x] Fix tests since I forgot to do that before putting this PR up.


d2l-filter-menu/d2l-filter-menu-tab.html
```diff
@@ -0,0 +1,149 @@
...
+			}
+			:host([hidden]) {
+				display: none !important;
+			}
+			d2l-search-widget {
+				--d2l-search-widget-height: 45px;
+				margin: 10px 20px;
+			}
+			.no-items-text {
+				@apply --d2l-body-compact-text;
```
dlockhart: You should probably explicitly import typography shared styles (which defines this mixin) in case whatever else is importing it goes away one day.

d2l-filter-menu/d2l-filter-menu-tab.html
```diff
@@ -0,0 +1,149 @@
...
+					value: false
+				}
+			},
+			behaviors: [
+				window.D2L.MyCourses.LocalizeBehavior,
+				window.D2L.MyCourses.UtilityBehavior
+			],
+			attached: function() {
+				this.listen(this.$$('d2l-search-widget'), 'd2l-search-widget-results-changed', '_onSearchWidgetResultsChanged');
+				this.listen(this.$$('d2l-menu'), 'd2l-menu-item-change', '_onMenuItemChange');
```
dlockhart: Instead of wiring up and unlistening, you could either:
- if you don't care about where the events came from, [use `listeners`](https://www.polymer-project.org/1.0/docs/devguide/events#event-listeners) (automatically listens and unlistens)
- if you're worried about targeting a specific element, use `on-d2l-search-widget-results-changed` attribute on the `d2l-search-widget` element

d2l-filter-menu/d2l-filter-menu-tab.html
```diff
@@ -0,0 +1,149 @@
...
+				var items = this.$$('d2l-menu').querySelectorAll('d2l-list-item-filter');
+				items.forEach(function(item) {
+					item.selected = false;
+				});
+
+				this.$$('d2l-search-widget').clear();
+				this.set('selectedFilters', []);
+			},
+			resize: function() {
+				this.$$('d2l-menu').resize();
```
dlockhart: All the code in here seems kind of unfortunate... curious why it's necessary.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+				display: flex;
+				flex-direction: column;
+			}
+			:host([hidden]) {
+				display: none !important;
+			}
+			button:hover,
+			button:focus {
+				text-decoration: underline;
+				color: var(--d2l-color-celestine);
```
dlockhart: Don't assume colors will be imported for you... they might not be one day.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+			.dropdown-content-header {
+				box-sizing: border-box;
+				display: flex;
+				justify-content: space-between;
+				border-bottom: 1px solid var(--d2l-color-titanius);
+				width: 100%;
+				padding: 20px;
+			}
+			.clear-button {
+				@apply --d2l-body-small-text;
```
dlockhart: Again, you'll want to import typography shared styles if you're using their mixins.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+		<div class="dropdown-content-header">
+			<span>[[localize('filtering.filterBy')]]</span>
+			<template is="dom-if" if="[[_hasFilters(currentFilters)]]">
+				<button class="clear-button" on-tap="clearFilters">[[localize('filtering.clear')]]</button>
+			</template>
+		</div>
+
+		<div id="contentView">
+			<div class="dropdown-content-tabs" role="tablist">
+				<div class="dropdown-content-tab" role="tab" aria-controls="semesterList">
```
dlockhart: I can't find an element that has `id="semesterList"` anymore

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+					value: false
+				}
+			},
+			behaviors: [
+				window.D2L.Hypermedia.HMConstantsBehavior,
+				window.D2L.MyCourses.LocalizeBehavior,
+				window.D2L.MyCourses.UtilityBehavior
+			],
+			attached: function() {
+				this.listen(this.$.semestersTab, 'selected-filters-changed', '_onSelectedFiltersChanged');
```
dlockhart: Again, listeners might be a cleaner way to do this.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+			detached: function() {
+				this.unlisten(this.$.semestersTab, 'selected-filters-changed', '_onSelectedFiltersChanged');
+				this.unlisten(this.$.departmentsTab, 'selected-filters-changed', '_onSelectedFiltersChanged');
+			},
+			observers: [
+				'_updateNames(filterStandardSemesterName, filterStandardDepartmentName)'
+			],
+			ready: function() {
+				this.filterStandardSemesterName = this.localize('filtering.semesterDefault');
+				this.filterStandardDepartmentName = this.localize('filtering.departmentDefault');
```
dlockhart: Not sure about these 2 lines -- what are they needed for? Worried that occasionally "ready" will be called before localization gets resolved.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+				this._semestersTabVisible = tabName === 'semesters';
+				this._departmentsTabVisible = tabName === 'departments';
+
+				this.$.semestersTab.resize();
+				this.$.departmentsTab.resize();
+
+				this.$.semestersTabButton.setAttribute('aria-pressed', this._semestersTabVisible);
+				this.$.departmentsTabButton.setAttribute('aria-pressed', this._departmentsTabVisible);
+			},
+			_updateTabText: function() {
```
dlockhart: Could the semesters/departments tab button text be a computed property instead? Setting `textContent` seems super dirty, and I'd be worried about encoding of accents and such.

d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -0,0 +1,262 @@
...
+				this.$.departmentsTabButton.textContent = this.localize(
+					'filtering.filterLabel',
+					'filterLabel', this.filterStandardDepartmentName,
+					'num', this.departmentFilters.length
+				);
+			},
+			_updateNames: function(semesterName, departmentName) {
+				this._updateTabText();
+				this.$.semestersTab.searchPlaceholderText = this.localize('filtering.searchBy', 'filter', semesterName);
+				this.$.departmentsTab.searchPlaceholderText = this.localize('filtering.searchBy', 'filter', departmentName);
```
dlockhart: Same here. Could these not be computed properties?

d2l-filter-menu/d2l-filter-menu-tab.html
ryantmer: Ah yes, forgot about `listeners`! Works perfectly in this case, I'll do that. Thanks :)

d2l-filter-menu/d2l-filter-menu-tab.html
ryantmer: The DE mentioned I think has more info, but the problem is that the `d2l-menu` doesn't always resize correctly after the list of semesters comes back and is rendered; the result of this is that the menu can be suuuuuper long. Until you resize the window, at which point it'll shrink down to the right size, so this forces that.

There is like, 100% a better solution out there, but

d2l-filter-menu/d2l-filter-menu.html
ryantmer: Ah yep, missed these. I'll update them (should be `semestersTab`/`departmentsTab`)

d2l-filter-menu/d2l-filter-menu.html
ryantmer: Really all this is doing is covering for when there isn't a `filterStandardSemesterName` coming down from the LMS (e.g. if Departments are called "Terms" or something, we should show "Terms" on this tab instead). In practice, I've not seen a problem without this from a UI perspective, so might be safe to ditch it. I'll try it out!

d2l-filter-menu/d2l-filter-menu.html
ryantmer: Weird - I was relying on the automatic [change notification events](https://www.polymer-project.org/2.0/docs/devguide/data-system#change-events) for this one, but it seems like `listeners` doesn't support that? Either way, I like `listeners` better, so I'll just explicitly `this.fire('selected-filters-changed')`, but... weird!

d2l-filter-menu/d2l-filter-menu.html
ryantmer: A much better approach indeed! Turns out there were a few more that could be done nicer, added in the next commit.

ryantmer:
Hokay, fixed the outstanding issues and added tests. Ready for further review!

PR #401, opened 2017-10-13
Build/Travis updates

ryantmer:
- Remove frau-publisher stuff, which hasn't been used since the very, very early days
- Remove unneeded steps from .travis.yml
- Add Sauce, but disabled (some of the tests are failing, no time to fix them right now)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,13 @@
 language: node_js
-node_js:
-- 6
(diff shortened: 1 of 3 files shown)


ryantmer:
Just going to merge this - Travis out is as expected (still lints and runs tests, which is all it was actually doing before).

PR #404, opened 2017-10-16
Show semester sometimes

trevorjonesd2l:
Friendly PRs
✔️  Merged: [Add widget customization for Show Semester (le)](https://git.dev.d2l/projects/CORE/repos/le/pull-requests/8378/overview) 
✔️  Merged: [Add link to course's semester (lp)](https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/6815/overview)

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -121,8 +121,15 @@
 			}
 			.course-code-text {
 				@apply --d2l-body-small-text;
-				display: block;
+				display: inline-block;
 				text-transform: uppercase;
+			}
+			.course-semester-text {
+				@apply --d2l-body-small-text;
+				display: inline-block;
+			}
+			.separator-icon {
+				color: var(--d2l-color-tungsten)
 			}
 			.course-text {
 				margin-top: 0.6rem;
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -119,15 +119,25 @@
 					filter: none;
 				}
 			}
-			.course-code-text {
-				@apply --d2l-body-small-text;
+			.course-code-text,
+			.separator-icon,
+			.course-semester-text {
+				display: none;
+			}
+			:host([show-course-code]) .course-code-text,
+			:host([show-course-code]) .separator-icon,
+			:host([show-semester]) .course-semester-text {
 				display: inline-block;
+			}
+
+			.course-code-text, .course-semester-text  {
+				@apply --d2l-body-small-text;
+			}
+
+			.uppercase {
 				text-transform: uppercase;
 			}
-			.course-semester-text {
-				@apply --d2l-body-small-text;
-				display: inline-block;
-			}
+
 			.separator-icon {
 				color: var(--d2l-color-tungsten)
 			}
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -56,17 +56,10 @@ the user has in that organization - student, teacher, TA, etc.
 				<div class="course-info">
 					<div class="course-text">
 						[[_organization.properties.name]]
-						<div style="display:block">
-							<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]">
-								<span class="course-code-text">[[_organization.properties.code]]</span>
-							</template>
-
-							<template id="courseSemesterTemplate" is="dom-if" if="[[_semesterName]]">
-								<template is="dom-if" if="[[showCourseCode]]">
-									<d2l-icon class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
-								</template>
-								<span class="course-semester-text">[[_semesterName]]</span>
-							</template>
+						<div>
+							<span show-course-code class="course-code-text uppercase">[[_organization.properties.code]]</span>
+							<d2l-icon class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
+							<span class="course-semester-text">[[_semesterName]]</span>
 						</div>
 						<d2l-offscreen id="courseNotificationLabel" aria-hidden="true">[[_courseNotificationLabel]]</d2l-offscreen>
 					</div>
@@ -161,9 +154,15 @@ the user has in that organization - student, teacher, TA, etc.
 				},
 				userId: String,
 				// Configuration value passed in to toggle course code
-				showCourseCode: Boolean,
+				showCourseCode: {
+					type: Boolean,
+					reflectToAttribute: true
+				},
 				// Configuration value passed in to toggle course code
-				showSemester: Boolean,
+				showSemester: {
+					type: Boolean,
+					reflectToAttribute: true
+				},
 				isStartedInactive: {
 					type: Boolean,
 					value: false
@@ -360,7 +359,7 @@ the user has in that organization - student, teacher, TA, etc.
 			},
 			_fetchNotifications: function() {
 				if (!this._notificationsUrl) {
-					return;
+					return Promise.resolve();
 				}
 
 				return this.fetchSirenEntity(this._notificationsUrl);
@@ -372,6 +371,17 @@ the user has in that organization - student, teacher, TA, etc.
 				}
 				return Promise.resolve();
 			},
+			_fetchEnrollmentData: function() {
+				return this._fetchOrganization()
+					.then(this._onOrganizationResponse.bind(this))
+					.then(function() {
+						return Promise.all([
+							this._fetchSemester(),
+							this._fetchNotifications()
+								.then(this._onNotificationsResponse.bind(this))
+						]);
+					}.bind(this))
+			},
 			_enrollmentChanged: function() {
 				this.pinned = this.enrollment.hasClass(this.HypermediaClasses.enrollments.pinned);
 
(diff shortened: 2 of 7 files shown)


d2l-course-tile-styles.html
```diff
@@ -121,9 +121,16 @@
...
-				display: block;
+				display: inline-block;
 				text-transform: uppercase;
 			}
+			.course-semester-text {
+				@apply --d2l-body-small-text;
+				display: inline-block;
+			}
+			.separator-icon {
+				color: var(--d2l-color-tungsten)
```
dlockhart: Is there a special reason it's "tungsten" and not just using the default "ferrite"?
https://github.com/BrightspaceUI/icons/blob/master/d2l-icon.html#L27

d2l-course-tile.html
```diff
@@ -56,9 +56,18 @@
 				<div class="course-info">
 					<div class="course-text">
 						[[_organization.properties.name]]
-						<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]">
-							<span class="course-code-text">[[_organization.properties.code]]</span>
-						</template>
+						<div style="display:block">
```
dlockhart: Don't need `style="display:block"` -- DIVs are block elements by default.

d2l-course-tile-styles.html
trevorjonesd2l: To match the font of the course code and semester

d2l-course-tile.html
```diff
@@ -56,9 +56,18 @@
...
+							<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]">
```
dlockhart: Another option instead of all these `dom-if`s is to use CSS attribute selectors to hide things you don't want to show. It'll result in simpler (and therefore more performant) markup because of the fewer nodes, and CSS will always be faster than waiting for Polymer to resolve these properties:

```
.course-code-text,
.separator-icon,
.course-semester-text {
    display: none;
}
:host([show-course-code]) .course-code-text,
:host([show-course-code]) .separator-icon,
:host([show-semester]) .course-semester-text {
    display: inline-block;
}
```

d2l-course-tile.html
```diff
@@ -364,17 +383,27 @@
...
 						.then(this._onOrganizationResponse.bind(this))
-						.then(this._fetchNotifications.bind(this))
-						.then(this._onNotificationsResponse.bind(this));
+						.then(function() {
+							var fetchNotificationsFunction = this._fetchNotifications.bind(this);
+							fetchNotificationsFunction()
+								.then(this._onNotificationsResponse.bind(this));
+
+							this._fetchSemester();
+						}.bind(this));
```
dlockhart: I know this code was already here, but it's getting extra... duplicate-y.

d2l-course-tile-styles.html
dlockhart: OK cool. In the navbar we use `mica` for our separators that we want to be a bit lighter than the other icons. We should probably be consistent. @thehappypixel any thoughts?

d2l-course-tile.html
```diff
@@ -364,17 +383,27 @@
 					if (!this.delayLoad) {
 						return this._fetchOrganization()
 							.then(this._onOrganizationResponse.bind(this))
-							.then(this._fetchNotifications.bind(this))
-							.then(this._onNotificationsResponse.bind(this));
+							.then(function() {
```
ryantmer: This block can be just

```
.then(Promise.all([
    this._fetchSemester.bind(this),
    this._fetchNotifications.bind(this).then(this._onNotificationsResponse.bind(this))
]);
```

The change you've got here actually slightly changes the flow, since it's not returning the result of the last call (not that we do anything with it anyway, granted).

It's a little wordy, though, so I agree with Dave's comment below that it can probably be moved into a function.

d2l-course-tile.html
```diff
@@ -511,6 +543,10 @@
 				}
 				this._setCourseUpdates(total);
 			},
+			_onSemesterResponse: function(semester) {
+				this._semesterName = semester.properties.name;
```
ryantmer: It's probably overly cautious, but doing this super-safely wouldn't be a bad idea probably. (i.e. `(semester.properties || {}).name`)

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -151,6 +195,68 @@ describe('<d2l-course-tile>', function() {
...
+			var separator = widget.$$('.separator-icon');
+			expect(separator).to.not.be.null;
+		});
+
+		it('should not set the semester name if the show semester config is false', function() {
+			widget.fetchSirenEntity = sinon.stub().returns(Promise.resolve(semesterOrganizationEntity));
+
+			widget.showSemester = false;
+			widget._semesterUrl = '/organizations/2';
+			widget._fetchSemester().then(function() {
```
ryantmer: When using this format with Mocha, you'll have to [`return` the Promise](http://tobyho.com/2015/12/16/mocha-with-promises/), otherwise it's entirely possible the `expect`s aren't actually being verified :) (or if they are, the test will still pass and it'll result in other console errors)

d2l-course-tile.html
```diff
@@ -56,9 +56,11 @@
 				<div class="course-info">
 					<div class="course-text">
 						[[_organization.properties.name]]
-						<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]">
-							<span class="course-code-text">[[_organization.properties.code]]</span>
-						</template>
+						<div>
```
ryantmer: Is this div needed? Can't tell what it's doing, just `inline-block` magic?

d2l-course-tile.html
```diff
@@ -191,6 +201,12 @@
 				_showHoverMenu: {
 					type: Boolean,
 					value: false
+				},
+				_semesterName: String,
+				_hideSeparator: {
+					type: Boolean,
+					value: true,
+					computed: '_computeHideSeperator(_semesterName)'
```
ryantmer: > _computeHideSeperator
> Seperator

🤕

d2l-course-tile.html
```diff
@@ -191,6 +201,12 @@
 				_showHoverMenu: {
 					type: Boolean,
 					value: false
+				},
+				_semesterName: String,
+				_hideSeparator: {
```
ryantmer: Y'know... I keep calling these kind of variables `_showX`, but it is kinda confusing when it's a `hidden$="[[!_showX]]"`... Maybe I should use `_hideX` instead.

d2l-course-tile.html
```diff
@@ -56,9 +56,11 @@
 				<div class="course-info">
 					<div class="course-text">
 						[[_organization.properties.name]]
-						<template id="courseCodeTemplate" is="dom-if" if="[[showCourseCode]]">
-							<span class="course-code-text">[[_organization.properties.code]]</span>
-						</template>
+						<div>
+							<span show-course-code class="course-code-text uppercase">[[_organization.properties.code]]</span>
+							<d2l-icon hidden="{{_hideSeparator}}" class="separator-icon" icon="d2l-tier1:bullet"></d2l-icon>
```
ryantmer: Since `hidden` is a native-HTML attribute, it's recommended* to use the `$=` rather than just `=`. (Also this can be a one-way binding, doesn't need to be two-way.)

*I've never actually seen it make a difference, but... yeah idk.

d2l-course-tile.html
```diff
@@ -489,7 +520,7 @@
 				return Promise.resolve();
 			},
 			_onNotificationsResponse: function(notifications) {
-				if (!this.courseUpdatesConfig) {
+				if (!this.courseUpdatesConfig || !notifications) {
 					return;
```
ryantmer: Eh, not really a big deal, but this should probs return a resolve as well.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -133,22 +177,97 @@ describe('<d2l-course-tile>', function() {
...
+			var separator = widget.$$('.separator-icon');
+			expect(separator.getAttribute('hidden')).to.not.be.null;
+		});
+
+		it('should show the seperator if the course code is on and semester name is a real thing', function() {
+			widget.showCourseCode = true;
+			widget._semesterName = 'Doop';
+			widget.showSemester = true;
+			var separator = widget.$$('.separator-icon');
+			expect(separator.getAttribute('hidden')).to.be.null;
```
ryantmer: Dave L in my other PR just pointed out to me - `expect(separator.hasAttribute('hidden')).to.be.false` is cleaner.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -133,22 +177,97 @@ describe('<d2l-course-tile>', function() {
...
 
 		it('should not show the course code if configured false', function() {
 			widget.showCourseCode = false;
-			widget.$.courseCodeTemplate.render();
 			var courseCode = widget.$$('.course-code-text');
-			expect(courseCode).to.be.null;
+			expect(window.getComputedStyle(courseCode).getPropertyValue('display')).to.equal('none');
+		});
+
+		it('should show the semester if the showSemester is set', function() {
```
ryantmer: These 5-6 tests here could be combined into a `forEach` if you wanted, [something like this](https://github.com/Brightspace/d2l-my-courses-ui/blob/9a2c7a0779b4c4f510b19ab846b5b176e4974bef/test/d2l-course-tile-responsive-grid-behavior/d2l-course-tile-responsive-grid-behavior.js#L112)

d2l-course-tile.html
trevorjonesd2l: pretty much grouping the info below course title

d2l-course-tile.html
trevorjonesd2l: Separator is not spelled how I'd expect

ryantmer: > _hideSeparator

but also 

> _computeHideSeperator

so like, _part_ of you expects it to be spelled correctly, at least 😜

PR #405, opened 2017-10-17
US90278 - Role filter tab

ryantmer:
This builds off of #403 and adds the Role tab to the filter menu. Works mostly the same as the department/semester tabs do, except no search field as that's not supported on the API.

Still need to work out some minor kinks, and add/fix tests, but ready for an initial look at least.


d2l-all-courses.html
```diff
@@ -389,23 +391,17 @@
...
 				var text;
-				if (length === 0) {
+				if (this._filterCount === 0) {
 					text = this.localize('filtering.filter');
-				} else if (length === 1) {
+				} else if (this._filterCount === 1) {
 					text = this.localize('filtering.filterSingle');
 				} else {
-					text = this.localize('filtering.filterMultiple', 'num', length);
+					text = this.localize('filtering.filterMultiple', 'num', this._filterCount);
```
dlockhart: Not something to do now since it would require re-translation, but this would have been a great use case for the [pluralization stuff](https://formatjs.io/guides/message-syntax/#plural-format) in the ICU message format.

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+<link rel="import" href="../../d2l-icons/d2l-icons.html">
+<link rel="import" href="../../d2l-menu/d2l-menu-item-selectable-behavior.html">
+<link rel="import" href="../../d2l-menu/d2l-menu-item-selectable-styles.html">
+<link rel="import" href="../../d2l-typography/d2l-typography-shared-styles.html">
+
+<dom-module id="d2l-filter-list-item-role">
+	<template>
+		<style include="d2l-menu-item-selectable-styles">
+			:host {
+				padding: 0.5rem 1rem;
```
dlockhart: By default this WC is going to be `display:inline`, which means it won't have top/bottom padding. Do you want it to be `inline-block`, `block` or not have top/bottom padding?

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			:host > span {
+				@apply --d2l-body-compact-text;
+			}
+
```
dlockhart: These 2 rules could be simplified to just be `span` and `d2l-icon`, since there's only 1 of each.

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+			:host d2l-icon {
+				--d2l-icon-height: 1rem;
+				--d2l-icon-width: 1rem;
```
dlockhart: The intention with the size is for it to scale up/down with the user's font preference?

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+				visibility: visible;
```
dlockhart: Default for `visibility` is `visible`, so not needed?

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+				margin-right: 0.5rem;
```
dlockhart: Probably needs to be flipped in RTL?

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+			}
+		</style>
+		<d2l-icon icon="d2l-tier2:check-box-unchecked" aria-hidden="true"></d2l-icon>
+		<span>[[text]]</span>
```
dlockhart: Is the span even needed or could the `compact` style just be applied to the host? Less DOM nodes = better performance. :)

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+	</template>
+	<script>
+		'use strict';
```
dlockhart: I've seen this a bunch, but I don't think we need strict mode in our WCs.

d2l-filter-menu/d2l-filter-list-item-role.html
```diff
@@ -0,0 +1,59 @@
...
+				}.bind(this));
+			},
+
+			_onSelect: function(e) {
+				this.set('selected', !this.selected);
+				this.__onSelect(e);
+			},
+			_onSelectedChanged: function(e) {
+				// The selected state can be changed by click/tap - need to update the icon when this happens
+				this.$$('d2l-icon').setAttribute('icon', this.selected ? 'd2l-tier2:check-box' : 'd2l-tier2:check-box-unchecked');
```
dlockhart: Could the icon just be a computed property?

dlockhart: Alternate: have 2 icons and use a CSS selector to switch between them

```
.icon-checked,
:host([selected]) .icon-unchecked {
    display: none;
}
:host([selected]) .icon-checked {
    display: inline-block;
}
```

d2l-filter-menu/d2l-filter-menu-tab-roles.html
```diff
@@ -0,0 +1,127 @@
...
+
+<dom-module id="d2l-filter-menu-tab-roles">
+	<template>
+		<style>
+			:host {
+				display: flex;
+				flex-direction: column;
+			}
+			:host([hidden]) {
+				display: none !important;
```
dlockhart: I've never needed to use `!important` in a WC and I've read it's a code smell. Is there a use case you've run into for it?

d2l-filter-menu/d2l-filter-menu-tab-roles.html
```diff
@@ -0,0 +1,127 @@
...
+			},
+			behaviors: [
+				window.D2L.MyCourses.UtilityBehavior
+			],
+			listeners: {
+				'd2l-menu-item-change': '_onMenuItemChange'
+			},
+
+			clear: function() {
+				var items = this.$$('d2l-menu').querySelectorAll('d2l-filter-list-item-role');
```
dlockhart: Pretty sure `NodeList`s can't be `foreach`ed in Safari, IE and Edge:
https://css-tricks.com/snippets/javascript/loop-queryselectorall-matches/

d2l-all-courses.html
ryantmer: Neat - didn't know that was a thing, thanks!

d2l-filter-menu/d2l-filter-list-item-role.html
ryantmer: Admittedly, most of this was just copy+pasted from `d2l-filter-list-item` - I'll update both of them with these suggestions though.

d2l-filter-menu/d2l-filter-list-item-role.html
ryantmer: That I don't know... I'll look back in the commits on the other file and see if can figure it out/remember, because not sure 😮

d2l-filter-menu/d2l-filter-list-item-role.html
ryantmer: `eslint` complains about it when it's lacking, granted we can always change that!

d2l-filter-menu/d2l-filter-menu-tab-roles.html
ryantmer: I feel like this surfaced somewhere IE, but I'll check to make sure; quite possible that it was a holdover from some specific circumstance where it _was_ required, that turned into a general-case addition that wasn't needed.

d2l-filter-menu/d2l-filter-menu-tab-roles.html
ryantmer: Huh, I thought it was being done elsewhere. I'll verify that! (This needs a few changes anyway, since there's one case where I can consistently break this functionality)

d2l-filter-menu/d2l-filter-list-item-role.html
ryantmer: Huh - the padding is still being applied, although I'll set them to `display: block` anyway.

d2l-filter-menu/d2l-filter-list-item-role.html
ryantmer: Ah, right - it's an override for [this](https://github.com/BrightspaceUI/menu/blob/master/d2l-menu-item-selectable-styles.html#L33), because the default menu-item-selectable-behavior doesn't show the "empty checkbox" state, it just doesn't show a checkbox.

d2l-filter-menu/d2l-filter-menu-tab-roles.html
```diff
@@ -0,0 +1,127 @@
...
+					changeFilterAction = e.detail.value.getActionByName('remove-filter');
+				}
+				// Create the URL to add/remove the filter to overall filter
+				var url = this.createActionUrl(changeFilterAction);
+
+				this.fetchSirenEntity(url)
+					.then(function(roleFiltersEntity) {
+						// Use the apply-role-filters action to create the new searchUrl
+						var applyAction = roleFiltersEntity.getActionByName('apply-role-filters');
+						var roles = applyAction.getFieldByName('roles').value;
```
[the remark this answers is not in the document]
ryantmer: It is, this is just used for the `filterCount` - there's probably other ways to do this granted, but I figured this way we're taking it right from the state Entity

d2l-utility-behavior.html
```diff
@@ -31,7 +31,16 @@
 				return key + '=' + query[key];
 			}).join('&');
 
-			return queryString ? action.href + '?' + queryString : action.href;
+			if (!queryString) {
+				return action.href;
+			}
+
+			if (action.href.indexOf('?') > -1) {
```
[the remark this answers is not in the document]
ryantmer: Yeah, I guess ideally I'd build `query` off the existing URL, but as you say, building a query object from the `href` is kinda... argh.

The downside to this approach is that if the query already has a particular param, we'll add it a second time; however, we take the overwrite path in general here anyway, and multiple query params _usually_ just get squished down into "whatever the last one was"*

(*In general, at least. In Valence, however, I've found an instance where `?sort=A&sort=B&sort=C` will result in `sort == C`, and another case where it'll result in `sort == A,B,C`. Not be confused with `?sort=A,B,C`, which still results in `sort == A,B,C`, but via a different method... yeah it's wut.)

PR #408, opened 2017-10-20
Clean up hiding things

ryantmer:
Yeeeaaah, this is kinda out of nowhere, but didn't want to to inflate another PR. This just leverages the `hidden` attribute more (and computed properties, where possible), and explicitly `toggleClass`-ing less. I personally find this method a lot easier to read, as it includes any hide/unhide stuff in the DOM, which is always nice. Thoughts/opinions?

I've confirmed that the things I touched here (main content, overlay content, notifications, advanced search URL, and infinite-scroll-spinner) all continue to work as intended.

### DIFF
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -353,11 +353,10 @@
 				}, this);
 				this.set('delayLoad', false);
 
-				setTimeout(function() {
-					// Slight delay to allow for overlay to get a DOM width before triggering the recalculation
+				requestAnimationFrame(function() {
 					this.fire('recalculate-columns');
 					this._showContent = true;
-				}.bind(this), 50);
+				}.bind(this));
 			},
 			open: function() {
 				this.$$('#all-courses').open();
--- a/d2l-all-courses.html
+++ b/d2l-all-courses.html
@@ -32,10 +32,7 @@
 			with-backdrop
 			restore-focus-on-close>
 
-			<div
-				id="overlayContent"
-				class="d2l-all-courses-hidden">
-
+			<div hidden$="[[!_showContent]]">
 				<iron-scroll-threshold
 					id="all-courses-scroll-threshold"
 					on-lower-threshold="_onAllCoursesLowerThreshold">
@@ -90,10 +87,8 @@
 							</d2l-dropdown>
 						</div>
 					</div>
-					<div id="advancedSearchLink" class="search-and-filter-row d2l-all-courses-hidden">
-						<div class="advanced-search-link">
-							<a is="d2l-link" href$="[[advancedSearchUrl]]">{{localize('advancedSearch')}}</a>
-						</div>
+					<div class="search-and-filter-row advanced-search-link" hidden$="[[!_showAdvancedSearchLink]]" >
+						<a is="d2l-link" href$="[[advancedSearchUrl]]">{{localize('advancedSearch')}}</a>
 					</div>
 				</div>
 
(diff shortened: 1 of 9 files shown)


d2l-all-courses.html
```diff
@@ -347,9 +353,7 @@
 				setTimeout(function() {
 					// Slight delay to allow for overlay to get a DOM width before triggering the recalculation
 					this.fire('recalculate-columns');
-
-					this.toggleClass('d2l-all-courses-hidden', false, this.$.overlayContent);
-					this.toggleClass('d2l-all-courses-hidden', true, this.$.overlayLoadingSpinner);
+					this._showContent = true;
```
dlockhart: Does this line still need to be in the `setTimeout`? Timeouts are so gross...

ryantmer: Unfortunately, yeah :( There's _some_ delay before everything kinda settles and gets width, so putting this outside the timeout still works, but is kinda janky - you get a full-width course-tile flashing for a fraction of a second before everything arranges properly.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -486,36 +486,36 @@ describe('<d2l-course-tile>', function() {
...
+			expect(widget.$.courseUpdates.getAttribute('hidden')).to.be.null;
 			expect(widget._courseUpdates).to.equal('99+');
 			expect(widget.$$('.update-text-box').innerText).to.equal('99+');
 
 		});
 
 		it('should not show updates number when 0', function() {
 			widget._setCourseUpdates(0);
-			expect(widget.$.courseUpdates.getAttribute('class')).to.contain('d2l-updates-hidden');
+			expect(widget.$.courseUpdates.getAttribute('hidden')).to.not.be.null;
```
dlockhart: Suggestion: `courseUpdates.hasAttribute('hidden').to.be.true`

d2l-all-courses.html
ryantmer: I'd love to know _why_ it takes that extra time, granted... Maybe I'll look into that, since I think this is the only case where we need to do it. I assumed it was just due to how heavy All Courses was, but that's not really great if that is the case...

dbatiste: Yeah, usually `requestAnimationFrame` is what we really want.  Maybe we don't even need that though.

test/d2l-course-tile/d2l-course-tile.js
ryantmer: Woaah that works?! How did I not think of that... much nicer, I'll be updating #405 with that as well since otherwise this is about a quadruple negative to figure out what the values should be.

d2l-all-courses.html
dbatiste: Or even better, fast-dom it.

ryantmer: I'll give that a shot!

dbatiste: `requestAnimationFrame` is easy to switch to.  Switching to fast-dom may be a bigger lift for this.

dlockhart: But [fast-dom](https://github.com/Brightspace/fastdom-import) is likely the better actual play.

test/d2l-course-tile/d2l-course-tile.js
ryantmer: Ah, I had indeed tried this, but

`expect(blah.getAttribute('hidden')).to.be.true` -> _Error: expected '' to be true_
`expect(blah.getAttribute('hidden')).to.be.false` -> _Error: expected null to be false_

So looks like mocha can't do the conversion itself :(

dlockhart: `hasAttribute` though instead of `getAttribute`

ryantmer: ...oh man.

Reading is not my strong suit today, apparently. Thanks!

PR #412, opened 2017-10-31
DE27982 Combine role filters with the same name

ryantmer:
This isn't a super performant approach, although it is probably the most robust this can be approached (that I can think of). The idea is that we combine role filters with the same name (easy enough), and when a multi-filter item is selected, we make subsequent "add-filter" action calls for each of the individual filters. This needs to be sequentially, unfortunately - doing them in parallel would just end up in a state where only one is enabled - but it's probably "fast enough", especially once things are cached. If we want to hide the loading, it might also be a good idea to disable further input in the filter menu until the result set is actually updated.

### DIFF
--- a/d2l-filter-menu/d2l-filter-menu-tab-roles.html
+++ b/d2l-filter-menu/d2l-filter-menu-tab-roles.html
@@ -1,4 +1,5 @@
 <link rel="import" href="../../polymer/polymer.html">
+<link rel="import" href="../../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
 <link rel="import" href="../../d2l-menu/d2l-menu.html">
 <link rel="import" href="../../d2l-typography/d2l-typography-shared-styles.html">
 <link rel="import" href="../d2l-utility-behavior.html">
@@ -58,6 +59,7 @@
 				}
 			},
 			behaviors: [
+				window.D2L.Hypermedia.HMConstantsBehavior,
 				window.D2L.MyCourses.UtilityBehavior
 			],
 			listeners: {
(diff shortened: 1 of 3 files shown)


d2l-filter-menu/d2l-filter-menu-tab-roles.html
```diff
@@ -103,34 +107,81 @@
...
+			},
+			_parseFilterItems: function(roleFiltersEntity) {
+				this._roleFiltersEntity = roleFiltersEntity;
+				this._filterEntities = this._roleFiltersEntity.entities || [];
+
+				var filters = [];
+				// DE27982 - Filters with the same title should be combined into one item
+				this._filterEntities.forEach(function(filterEntity) {
+					var existingFilter = filters.find(function(filter) {
+						return filter.title === filterEntity.title;
```
jstefaniuk-d2l: Does case matter here? `Student`/`student`?

d2l-filter-menu/d2l-filter-menu-tab-roles.html
```diff
@@ -103,34 +107,81 @@
...
-						var roles = applyAction.getFieldByName('roles').value;
-						this.fire('role-filters-changed', {
-							url: searchUrl,
-							filterCount: roles.length ? roles.split(',').length : 0
+				for (var i = 1; i < filters.length; i++) {
+					var filter = filters[i];
+					request = request.then(function(newFiltersEntity) {
+						// The first fetchSirenEntity will change the entities, so grab the updated
+						// filter from the newly-updated role filters (based on role href)
+						var selfHref = filter.getLinkByRel('related').href;
```
njostonehouse: I _believe_ this may be absent if the user doesn't have access to the system roles API. It seems wrong to provide a link if following it is Forbidden. Not sure if that's implemented currently, but it seems plausible to have that be true at some point.

njostonehouse: I think we are confident that a user has the ability to "see" this role based on the way our system usually determines that. It seems we equate "this user can search for users with this role" with "can see role." I don't think this necessarily means that the link itself will work, though. So I want to see what determines _that_.

d2l-filter-menu/d2l-filter-menu-tab-roles.html
njostonehouse: @ryantmer agreed

PR #414, opened 2017-11-02
De27941: Unpinning course in My Courses doesn't update course list dropdown

trevorjonesd2l:
This more or less undoes [this commit](https://github.com/Brightspace/d2l-my-courses-ui/commit/1d331908094119ac69ff175d13eb7fcb24e09634#diff-37608caf11263d278dd1c2fa6da3bc3a) but updated to work with `enrolledUser`s.

This isn't a perfect or long term solution but it addresses the issue of pin state not syncing between the course selector and the my courses widget. The perfect solution would be for the course selector to pass `enrolledUser` links with the `pin-event` but that got into a mess of dependencies which would require some refactoring to the `D2L.LP.Tools` solution to work with the hypermedia/enrolled user stuff in the course selector. Since that will require a bigger refactor and it seems like we're heading into some course selector updates I'm going to create a story to work with `enrolledUser` in the Course Selector.

For now, what this fix does:
- Brings back the events listening for the `'d2l-course-pinned-change'`
- Added a map of the enrollments keyed with the `orgUnitId` so that the enrollments can be referenced to adding the the `_moveEnrollmentToPinnedList` and `_moveEnrollmentToUnpinnedList`.
- Added some code for refetching the enrollments, such that it doesn't get back the cached result. This is for when a course was pinned but not yet fetched, and since we can't call it back directly yet without the `enrolledUser` link we will need to fetch enrollments again.

What kinda stinks:
- No animation when unfetched course is pinned.
- Refetching every time a unfetched course is pinned.
- Still parsing out orgUnitId.

What it should do in the future (after `D2L.LP.Tools` refactor):
- Get enrolledUser link instead of orgUnitId.
- Get rid of orgUnitId map
- Get rid of refetch stuff

I will update the tests to this code as this goes through reviews

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -281,7 +281,7 @@
 				if (!this._setImageOrg) {
 					return;
 				}
-				return this._getOrgUnitIdFromLink(this.getEntityIdentifier(this._setImageOrg));
+				return this._getOrgUnitIdFromHref(this.getEntityIdentifier(this._setImageOrg));
 			},
 
 			/*
@@ -302,7 +302,7 @@
 			},
 			_onEnrollmentPinAction: function(e) {
 				var isPinned = e.type === 'enrollment-pinned';
-				var orgUnitId = this._getOrgUnitIdFromLink(this.getEntityIdentifier(e.detail.organization));
+				var orgUnitId = this._getOrgUnitIdFromHref(this.getEntityIdentifier(e.detail.organization));
 
 				if (!orgUnitId) {
 					return;
@@ -528,8 +528,8 @@
 						}
 					}
 
-					var orgLink = enrollment.getLinkByRel(this.HypermediaRels.organization);
-					var orgUnitId = this._getOrgUnitIdFromLink(orgLink);
+					var orgHref = (enrollment.getLinkByRel(this.HypermediaRels.organization) || {}).href;
+					var orgUnitId = this._getOrgUnitIdFromHref(orgHref);
 					if (!this.orgUnitIdMap.hasOwnProperty(orgUnitId)) {
 						this.orgUnitIdMap[orgUnitId] = enrollment;
 					}
@@ -625,8 +625,8 @@
 			_refreshPage: function() {
 				document.location.reload(true);
 			},
-			_getOrgUnitIdFromLink: function(organizationHref) {
-				var match = /[0-9]+$/.exec((organizationHref || {}).href);
+			_getOrgUnitIdFromHref: function(organizationHref) {
+				var match = /[0-9]+$/.exec(organizationHref);
 
 				if (!match) {
 					return;
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -239,13 +239,14 @@
 				'_updateEnrollmentAlerts(_hasEnrollments, _hasPinnedEnrollments)'
 			],
 			ready: function() {
+				this._updateEnrollmentAlerts(this._hasEnrollments, this._hasPinnedEnrollments);
 				this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
 			},
 			attached: function() {
 				this.performanceMark('d2l.my-courses.attached');
 				this._fetchRoot();
 
-				document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage.bind(this), true);
+				document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
 				this.$$('d2l-course-tile-grid').addEventListener('startedInactiveAlert', this._onStartedInactiveAlert.bind(this));
 				this.$['image-selector-threshold'].scrollTarget = this.$['basic-image-selector-overlay'].scrollRegion;
 			},
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -316,20 +316,19 @@
 				);
 			},
 			_onEnrollmentPinnedMessage: function(e) {
-				if (e.target !== this) {
-					var enrollment = this.orgUnitIdMap[e.detail.orgUnitId];
+				if (e.target === this) return;
 
-					if (enrollment) {
-						if (e.detail.isPinned) {
-							this._moveEnrollmentToPinnedList(enrollment);
-						} else {
-							this._moveEnrollmentToUnpinnedList(enrollment);
-						}
-						setTimeout(this._onStartedInactiveAlert.bind(this), 50);
+				var enrollment = this.orgUnitIdMap[e.detail.orgUnitId];
+				if (enrollment) {
+					if (e.detail.isPinned) {
+						this._moveEnrollmentToPinnedList(enrollment);
 					} else {
-						this.fetchSirenEntity(this.enrollmentsUrl)
-							.then(this._refetchEnrollments.bind(this));
+						this._moveEnrollmentToUnpinnedList(enrollment);
 					}
+					setTimeout(this._onStartedInactiveAlert.bind(this), 50);
+				} else {
+					this.fetchSirenEntity(this.enrollmentsUrl)
+						.then(this._refetchEnrollments.bind(this));
 				}
 			},
 			_onStartedInactiveAlert: function(e) {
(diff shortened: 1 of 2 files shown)


d2l-my-courses.html
```diff
@@ -447,15 +480,32 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
+				var enrollmentsSearchUrl = this.createActionUrl(searchAction, query);
+				if (bustCache) enrollmentsSearchUrl += '&bustCache=' + Math.random();
 
-				this.performanceMark('d2l.my-courses.search-enrollments.request');
-				return this.fetchSirenEntity(this._enrollmentsSearchUrl)
+				return enrollmentsSearchUrl;
+			},
+			_refetchEnrollments: function(enrollmentsRootEntity) {
+
+				this.performanceMark('d2l.my-courses.root-enrollments.response');
```
trevorjonesd2l: I'm guessing this performance event name should probably be different? Maybe something like `'d2l.my-courses.update-enrollments.response'`

d2l-my-courses.html
```diff
@@ -306,6 +315,23 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 					}
 				);
 			},
+			_onEnrollmentPinnedMessage: function(e) {
+				if (e.target !== this) {
```
dlockhart: Nit: reduce nesting by bailing early: `if (e.target === this) return;`

d2l-my-courses.html
```diff
@@ -447,15 +480,32 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
+				var enrollmentsSearchUrl = this._createFetchEnrollmentsUrl(enrollmentsRootEntity, true);
+
+				this.performanceMark('d2l.my-courses..search-enrollments.request');
+				return this.fetchSirenEntity(enrollmentsSearchUrl)
 					.then(this._populateEnrollments.bind(this));
 			},
 			_populateEnrollments: function(enrollmentsEntity) {
-
-				this.performanceMark('d2l.my-courses.search-enrollments.response');
+				this.performanceMark('d2l.my-courses..search-enrollments.response');
```
dlockhart: Looks like you have two decimals in here after "my-courses"

d2l-my-courses.html
```diff
@@ -447,15 +480,32 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
 				};
-				this._enrollmentsSearchUrl = this.createActionUrl(searchAction, query);
+				var enrollmentsSearchUrl = this.createActionUrl(searchAction, query);
+				if (bustCache) enrollmentsSearchUrl += '&bustCache=' + Math.random();
 
-				this.performanceMark('d2l.my-courses.search-enrollments.request');
-				return this.fetchSirenEntity(this._enrollmentsSearchUrl)
+				return enrollmentsSearchUrl;
+			},
+			_refetchEnrollments: function(enrollmentsRootEntity) {
```
dlockhart: Will this be the second (or 3rd, 4th) time the enrollments have been fetched? If so, I don't think we want to make more marks & measures -- typically we only want to end up with 1 set otherwise it gets really confusing looking at the data. We only really care about the initial load performance (for now).

trevorjonesd2l: OK yeah that makes sense I can remove this then

d2l-my-courses.html
```diff
@@ -236,14 +240,19 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 			],
 			ready: function() {
 				this._updateEnrollmentAlerts(this._hasEnrollments, this._hasPinnedEnrollments);
+				this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
```
ryantmer: Is this necessary if you change the line in `attached` to use `this._onEnrollmentPinnedMessage.bind(this)`, like the startedInactiveAlert one?

d2l-my-courses.html
```diff
@@ -447,21 +490,24 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 					sort: '-PinDate,OrgUnitName,OrgUnitId',
 					autoPinCourses: true
 				};
-				this._enrollmentsSearchUrl = this.createActionUrl(searchAction, query);
+				var enrollmentsSearchUrl = this.createActionUrl(searchAction, query);
+				if (bustCache) enrollmentsSearchUrl += '&bustCache=' + Math.random();
```
ryantmer: Good to know that this does indeed work!

d2l-my-courses.html
```diff
@@ -580,6 +632,14 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 					return;
 				}
 				return match[0];
+			},
+			_getOrgUnitIdFromLink: function(organizationHref) {
+				var match = /[0-9]+$/.exec((organizationHref || {}).href);
```
ryantmer: It's baaaack! 😱 

(but only hopefully-temporarily so I'm fine with it)

d2l-my-courses.html
```diff
@@ -481,6 +527,12 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 							this._unpinnedCoursesMap[enrollmentId] = true;
 						}
 					}
+
+					var orgLink = enrollment.getLinkByRel(this.HypermediaRels.organization);
```
ryantmer: If you just populate `orgUnitIdMap` the first time such an event occurs within `_onEnrollmentPinnedMessage` we can not do it in the vast majority of page loads.

(i.e. only bother when/if somebody pins something from the course selector, which my definitely-correct and certainly-not-made-up statistics show happens on only 0.00008% of page loads)

d2l-my-courses.html
trevorjonesd2l: tis not

d2l-my-courses.html
trevorjonesd2l: We talked about this it might not be worth it to move since the perf effect of this is negligible and this would mess with the animations on the first pin/unpin action

d2l-my-courses.html
```diff
@@ -235,15 +239,19 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 				'_updateEnrollmentAlerts(_hasEnrollments, _hasPinnedEnrollments)'
 			],
 			ready: function() {
-				this._updateEnrollmentAlerts(this._hasEnrollments, this._hasPinnedEnrollments);
```
ryantmer: Was this line supposed to disappear? Shouldn't it have been the other one? 😮

d2l-my-courses.html
```diff
@@ -481,6 +526,12 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 							this._unpinnedCoursesMap[enrollmentId] = true;
 						}
 					}
+
+					var orgLink = enrollment.getLinkByRel(this.HypermediaRels.organization);
+					var orgUnitId = this._getOrgUnitIdFromLink(orgLink);
```
ryantmer: Feels like something is a bit disconnected here - the `orgLink` you're passing in here is a Siren Link, but the signature of the function indicates that it's taking in an `organizationHref` - I think you can leave the regexing function unchanged for this PR, since in the existing cases it was passing in the organization Link, and you can do the same (`orgLink`, one line up).

d2l-my-courses.html
```diff
@@ -235,15 +239,19 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
...
 			],
 			ready: function() {
-				this._updateEnrollmentAlerts(this._hasEnrollments, this._hasPinnedEnrollments);
+				this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
 			},
 			attached: function() {
 				this.performanceMark('d2l.my-courses.attached');
 				this._fetchRoot();
 
+				document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage.bind(this), true);
```
jstefaniuk-d2l: .bind(this) returns a new function, so your `removeEventListener` won't actually remove the function

jstefaniuk-d2l: maybe do:
```
this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
 
```

PR #416, opened 2017-11-14
fixing a lint error with slots

dlockhart:


dlockhart:
`polymer lint` must have gotten smarter with a recent release, because without this fix it fails now.

PR #417, opened 2017-11-14
Remove telemetry

dlockhart:
This removed user telemetry from my-courses, but also upgrades to the image-selector version that no longer has telemetry.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -35,7 +35,7 @@
     "d2l-fetch": "Brightspace/d2l-fetch#^1.5.1",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^2.10.0",
     "d2l-icons": "^3.2.0",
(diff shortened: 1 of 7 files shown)


bower.json
```diff
@@ -35,7 +35,7 @@
     "d2l-fetch": "Brightspace/d2l-fetch#^1.5.1",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^2.10.0",
     "d2l-icons": "^3.2.0",
-    "d2l-image-selector": "git://github.com/Brightspace/image-selector#^2.0.1",
+    "d2l-image-selector": "git://github.com/Brightspace/image-selector#^2.1.0",
```
ryantmer: Huh, I guess this was technically a breaking change. Granted, setting no-longer-existing polymer properties is completely harmless afaik!

dlockhart: Yeah so it's a breaking change if we don't upgrade to the version that doesn't need them anymore.

PR #419, opened 2017-11-16
US90521 - Show pin indicator on course tile

trevorjonesd2l:
Yeah there's a pin indicator now. It's behind the feature flag for updated sort.
I added a new lang term for the label of the indicator button for a11y stuff, not translated yet but the translation has been submitted.

image

### DIFF
--- a/d2l-course-tile-styles.html
+++ b/d2l-course-tile-styles.html
@@ -181,6 +181,11 @@
 				display: flex;
 			}
 
+			:host-context([dir="rtl"]) .menu-bar {
+				left: 10px;
+				right: auto;
+			}
+
 			.hover-menu {
 				opacity: 0;
 				margin-top: 0px;
@@ -207,6 +212,10 @@
 				width: 35px;
 				height: 35px;
 				margin-right: 10px;
+			}
+			:host-context([dir="rtl"]) .menu-item {
+				margin-left: 10px;
+				margin-right: auto;
 			}
 			.menu-item.static {
 				margin-top: 10px;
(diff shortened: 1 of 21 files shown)


d2l-course-tile-styles.html
```diff
@@ -202,6 +206,10 @@
 				cursor: pointer;
 				width: 35px;
 				height: 35px;
+				margin-right: 10px;
```
dlockhart: I know this was existing CSS, but does this right margin need a reversal for RTL?

d2l-course-tile.html
```diff
@@ -456,6 +478,11 @@
 				this._onUnpinHover({detail: {hoverState: true}});
 				this._hoverCourseTile();
 			},
+			_pinPressHandler: function(e) {
+				if ( e.code === 'Space' || e.code === 'Enter' ) {
```
dlockhart: This will fail `eslint` when you refresh from `master`. Should have no spaces between the `(` and `)`.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -614,6 +614,61 @@ describe('<d2l-course-tile>', function() {
...
+		it('should show the pin indicator button when a course is pinned and feature flag is on', function(done) {
+			widget.pinned = true;
+			widget.updatedSortLogic = true;
+			setTimeout(function() {
+				expect(widget.pinned).to.be.true;
+
+				var pinIndicatorButton = widget.$$('#pin-indicator-button');
+				expect(pinIndicatorButton).to.exist;
+				done();
+			});
```
dlockhart: Are all the timeouts needed? Maybe `flush` with a callback might be less hacky? More info under "Test DOM mutations" here:
https://www.polymer-project.org/2.0/docs/tools/tests

PR #420, opened 2017-11-16
upgrading to latest eslint and using the brightspace config files

dlockhart:


dlockhart:
Aside: `use strict` isn't needed inside WCs, and neither are anonymous JS functions. I'll go through a second pass in a separate PR to remove those.

ryantmer:
Ah, yeah I think the only reason `use strict` was in there was because of eslint, which I guess isn't super WC-aware?

package.json
```diff
@@ -25,12 +25,10 @@
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
-    "bower": "^1.8.0",
```
ryantmer: 🙌 one step closer to not needing bower at all anymore!

PR #421, opened 2017-11-16
Cleanup

dlockhart:
- removes unneeded `use strict`s
- removed anonymous methods around behaviours
- upgrades to latest Bower dependencies


bower.json
```diff
@@ -28,38 +28,37 @@
...
+    "d2l-polymer-behaviors": "^1.1.0",
     "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#^2.0.1",
-    "d2l-simple-overlay": "git://github.com/Brightspace/simple-overlay#^1.1.1",
-    "d2l-typography": "^5.4.0",
-    "d2l-organization-hm-behavior": "git://github.com/Brightspace/organization-hm-behavior.git#^1.0.1",
-    "iron-a11y-announcer": "^1.0.5",
-    "iron-a11y-keys": "^1.0.6",
-    "iron-input": "^1.0.10",
-    "iron-menu-behavior": "^1.1.9",
-    "iron-overlay-behavior": "^1.10.1",
```
dlockhart: `iron-overlay-behavior` wasn't used anywhere

ryantmer: Ah yeah, probably not since https://github.com/Brightspace/simple-overlay/ became a thing

d2l-alert-behavior.html
```diff
@@ -1,52 +1,47 @@
 <link rel="import" href="../polymer/polymer.html">
-
 <script>
-	'use strict';
-
-	(function() {
```
ryantmer: So anonymous functions aren't required? Were they the recommended way to do it at some point or something, for some reason they seem correct in my head 😕

dlockhart: Not from what I can see in any of the old documentation, but it's possible they were once needed. I think Polymer likely executes all this in an anonymous context on your behalf anyway.

ryantmer: Huh, must have been a mistake at some point that just got carried across through many places... :/

PR #423, opened 2017-11-17
DE28251 Remove URLSearchParams

ryantmer:
Edge doesn't work with URLSearchParams yet, so can't use it. Instead, just building a string.

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -266,16 +266,16 @@ the user has in that organization - student, teacher, TA, etc.
 				// handler (to this same value), but that could take a few hundred ms, so do it before the request too.
 				this.pinned = !this.pinned;
 
-				var body = {};
+				var body = '';
 				var fields = pinAction.fields || [];
 				fields.forEach(function(field) {
-					body[field.name] = field.value;
+					body = body + field.name + '=' + field.value + '&';
 				});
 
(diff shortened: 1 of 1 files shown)


d2l-course-tile.html
```diff
@@ -266,16 +266,16 @@
 				// handler (to this same value), but that could take a few hundred ms, so do it before the request too.
 				this.pinned = !this.pinned;
 
-				var body = {};
+				var body = '';
 				var fields = pinAction.fields || [];
 				fields.forEach(function(field) {
-					body[field.name] = field.value;
+					body = body + field.name + '=' + field.value + '&';
```
ryantmer: This really is just creating what `URLSearchParams.toString()` would generate anyway.

dlockhart: Double check... I think you may want to be [URL-encoding](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) the field name and values.

Here's an example from the polyfill:
https://github.com/WebReflection/url-search-params/blob/master/src/url-search-params.js#L140

ryantmer: Ah, yeah good call. We're only ever sending true/false right now, but that doesn't mean we'll always be. I'll update, thanks!

PR #425, opened 2017-11-20
Re-enabling troublesome tests

ryantmer:
These tests were failing on Sauce on OSX browsers - sometimes Chrome, sometimes Firefox, sometimes Safari. Not entirely sure why, but this change was initially just intended to move the getFutureDate, getPastDate, and verifyOverlay functions into the inner test scope. With a bit of further poking, it seems that the root of the problem was the Date.now() happening only once, and the future/end dates being only 8 seconds off that. One method that worked was making the future/end dates 8,000 seconds off the current date, but for an even cleaner approach, we can re-make the future/end date for each test.

### DIFF
--- a/test/d2l-course-tile/d2l-course-tile.js
+++ b/test/d2l-course-tile/d2l-course-tile.js
@@ -658,47 +658,43 @@ describe('<d2l-course-tile>', function() {
 		});
 	});
 
-	var curDate = Date.now();
-	var formattedDate = 'FORMATTED_DATE';
-	var inactiveText = '(Inactive)';
+	describe('Notification Overlay', function() {
+		var org,
+			curDate,
+			futureDate,
+			pastDate,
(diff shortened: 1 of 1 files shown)


test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -658,47 +658,43 @@ describe('<d2l-course-tile>', function() {
...
-		} else {
-			expect(overlayInactive.textContent).to.not.equal(inactiveText);
-		}
-	}
-
-	describe.skip('Notification Overlay', function() {
-		var org;
-
 		beforeEach(function() {
+			curDate = Date.now();
```
ryantmer: This diff is kinda gross, but ultimately what I did here was:

- move `verifyOverlay` into the inner scope
- remove `getFutureDate` and `getPastDate` and replace them with variables set in `beforeEach`

PR #428, opened 2017-11-22
US91774 Use IntersectionObserver to lazy-load course tiles

ryantmer:
A primitive version of this behaviour already existed via the `delay-load` attribute, which was used to defer loading of course information for the tiles in the first fetch that weren't pinned. With `IntersectionObserver`, we are able to instead be more intelligent about this deferral, and simply load the course tile's information whenever it becomes partially visible to the user. It's quite neat and kinda makes me want to redo paging a bit, so that it feels more "infinite scroll"-y.

This includes a polyfill for Edge/IE (Firefox and Chrome support it natively).

TODO:

- [x] Get some performance numbers on this w.r.t. initial load

### DIFF
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -218,7 +218,7 @@ the user has in that organization - student, teacher, TA, etc.
 					computed: '_computeShowPinIndicator(pinned, updatedSortLogic)'
 				},
 				// Set by the IntersectionObserver when tile is first visible in viewport
-				_isVisible: Boolean
+				_load: Boolean
 			},
 			behaviors: [
 				window.D2L.Hypermedia.HMConstantsBehavior,
@@ -230,7 +230,7 @@ the user has in that organization - student, teacher, TA, etc.
 				'animationend': '_onAnimationComplete'
 			},
 			observers: [
-				'_fetchEnrollmentData(_isVisible, enrollment)'
+				'_fetchEnrollmentData(_load, enrollment)'
 			],
 			attached: function() {
 				Polymer.RenderStatus.afterNextRender(this, function() {
@@ -245,14 +245,27 @@ the user has in that organization - student, teacher, TA, etc.
 							// so we need to also make sure the tile is visible for that first run
 							// see https://bugs.chromium.org/p/chromium/issues/detail?id=713819
 							if (entries[i].intersectionRatio > 0) {
-								this._isVisible = true;
-								observer.unobserve(this.$$('.tile-container'));
+								observer.unobserve(tileContainerEl);
+								this._load = true;
+								break;
 							}
+						}
+
+						if (!this._load) {
+							// Tile isn't visible - schedule it to load via requestIdleCallback
+
+							// Small shim for Edge/IE/Safari
+							var delayFunction = window.requestIdleCallback || setTimeout;
+
+							delayFunction(function() {
+								observer.unobserve(tileContainerEl);
+								this._load = true;
+							}.bind(this));
 						}
 					};
 
 					var observer = new IntersectionObserver(observerCallback.bind(this));
-					observer.observe(this.$$('.tile-container'));
+					observer.observe(tileContainerEl);
 				});
 			},
 			ready: function() {
@@ -405,7 +418,7 @@ the user has in that organization - student, teacher, TA, etc.
 				return Promise.resolve();
 			},
 			_fetchEnrollmentData: function() {
-				if (!this.enrollment || !this._isVisible) {
+				if (!this.enrollment || !this._load) {
 					return;
 				}
 
--- a/d2l-course-tile.html
+++ b/d2l-course-tile.html
@@ -127,20 +127,8 @@ the user has in that organization - student, teacher, TA, etc.
 					type: Boolean,
 					reflectToAttribute: true
 				},
-				// Delays initial loading of the organization until explicitly triggered by `loadOrganization`
-				delayLoad: {
-					type: Boolean,
-					observer: '_delayLoadChanged'
-				},
-				/*
-				* Course tile enrollment Entity. When the enrollment is set, the course
-				* tile will automatically fetch information about the organization the
-				* enrollment is in.
-				*/
-				enrollment: {
-					type: Object,
-					observer: '_enrollmentChanged'
-				},
+				// Course tile enrollment entity
+				enrollment: Object,
 				enrollmentId: String,
 				// Specifies whether the (desktop) hover menu is enabled on the course tile
 				hoverEnabled: {
@@ -227,7 +215,9 @@ the user has in that organization - student, teacher, TA, etc.
 					type: Boolean,
 					value: false,
 					computed: '_computeShowPinIndicator(pinned, updatedSortLogic)'
-				}
+				},
+				// Set by the IntersectionObserver when tile is first visible in viewport
+				_isVisible: Boolean
 			},
 			behaviors: [
 				window.D2L.Hypermedia.HMConstantsBehavior,
@@ -238,13 +228,22 @@ the user has in that organization - student, teacher, TA, etc.
 			listeners: {
 				'animationend': '_onAnimationComplete'
 			},
+			observers: [
+				'_fetchEnrollmentData(_isVisible, enrollment)'
+			],
 			attached: function() {
 				Polymer.RenderStatus.afterNextRender(this, function() {
 					this.fire('course-tile-first-render');
 					var tileContainerEl = this.$$('.tile-container');
 					document.body.addEventListener('focus', this._onFocus, true);
 					tileContainerEl.addEventListener('focus', this._hoverCourseTile, true);
-				}.bind(this));
+
+					var observer = new IntersectionObserver(function() {
+						this._isVisible = true;
+						observer.unobserve(this.$$('.tile-container'));
+					}.bind(this));
+					observer.observe(this.$$('.tile-container'));
+				});
 			},
 			ready: function() {
 				this._onFocus = this._onFocus.bind(this);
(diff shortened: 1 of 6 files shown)


dbatiste:
Where does Safari fall support-wise?

d2l-course-tile.html
```diff
@@ -118,6 +118,7 @@
 
 		</div>
 	</template>
+	<script src="polyfill/IntersectionObserver.js"></script>
```
dbatiste: As coded, if this were included by multiple components, it would be executed more than once.  I think we want this to be wrapped as an import.  Fastdom is one such example were we've done that (but there are a few others).

dbatiste:
This is awesome Ryan.  I'm not sure if this is an issue or not... so I'll just throw it out there and maybe you'll know right away if it's a concern...

If the user is using the keyboard, moving focus through the tiles, and then they get to a point where hitting tab one more time _should_ cause scroll to happen, does it still?  i.e. do the tiles below the fold have an element that is focusable before the intersection observer runs?

ryantmer:
Good call on Safari - just checked it now and it's good, although I forgot to do so initially... support is listed as a bunch of question marks on mozilla.org, and caniuse indicates no support, so I suspect it's using the polyfill (polyfill.io, OTOH, says is supported in Safari 10... so nobody really knows, apparently).

In terms of keyboard navigation, that's an excellent catch - it actually ends up on the menu, since the tile itself doesn't appear to focusable until it has a course name. I'll see if I can figure that one out - might be as simple as loading the n+1th row using the IntersectionObserver's offset, so that by the time a user has tabbed to it, it has loaded and is focusable.

d2l-course-tile.html
ryantmer: Right! I have a note to import-ify node-siren-parser for that very reason. There's gotta be a bower-based one somewhere, I'll see what I can find.

dlockhart: Agreed, and Ryan's done it with `d2lfetch` and associated plugins, so should be relatively easy to apply here as well.

I'm curious how Polymer 3 + JS modules will handle this... maybe it'll magically be solved.

d2l-course-tile.html
```diff
@@ -238,13 +229,31 @@
...
 					tileContainerEl.addEventListener('focus', this._hoverCourseTile, true);
-				}.bind(this));
+
+					var observerCallback = function(entries, observer) {
+						for (var i = 0; i < entries.length; i++) {
+							// Chrome/FF immediately call the callback when we observer.observe()
+							// so we need to also make sure the tile is visible for that first run
+							// see https://bugs.chromium.org/p/chromium/issues/detail?id=713819
+							if (entries[i].intersectionRatio > 0) {
+								this._isVisible = true;
```
dlockhart: Do you not need a `.bind(this)` on your callback for `this._isVisible` to be setting itself on the right thing?

dlockhart: Nevermind, I see that's being done down below. Ignore me.

d2l-course-tile.html
```diff
@@ -238,13 +229,31 @@
...
+								observer.unobserve(this.$$('.tile-container'));
+							}
+						}
+					};
+
+					var observer = new IntersectionObserver(observerCallback.bind(this));
+					observer.observe(this.$$('.tile-container'));
```
dlockhart: `this.$$('.tile-container')` is called a bunch of times in this method, including up above inside the `afterNextRender` callback. Maybe pull `tileContainerEl` outside of the if-statements and reuse it?

d2l-course-tile.html
```diff
@@ -238,13 +229,31 @@
...
-				}.bind(this));
+
+					var observerCallback = function(entries, observer) {
+						for (var i = 0; i < entries.length; i++) {
+							// Chrome/FF immediately call the callback when we observer.observe()
+							// so we need to also make sure the tile is visible for that first run
+							// see https://bugs.chromium.org/p/chromium/issues/detail?id=713819
+							if (entries[i].intersectionRatio > 0) {
+								this._isVisible = true;
+								observer.unobserve(this.$$('.tile-container'));
```
dlockhart: Maybe worth flipping these 2 lines so the listener is removed right away? Probably also want to break out of the `for`-loop once this is handled once.

dbatiste:
Re keyboard, yeah, interaction observer is really cool and offers a nice way to defer loading some things.  It being the first time we've used it, it seems like we should be careful when using it to load focusable elements.

I think it might also be possible to place the focus once the tile becomes interactive, but we'd have to be careful not to steal focus away unexpectedly.

d2l-course-tile.html
ryantmer: Wow yeah, totally. Coincidentally was always the last item in testing it out, but that's not guaranteed.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -340,7 +304,9 @@ describe('<d2l-course-tile>', function() {
 			widget._pinClickHandler(event);
 
 			setTimeout(function() {
-				expect(window.d2lfetch.fetch).to.have.been.calledOnce;
+				expect(window.d2lfetch.fetch).to.have.been
+					.calledWith(sinon.match.has('url', sinon.match('/enrollments/users/169/organizations/1'))
+						.and(sinon.match.has('method', 'PUT')));
 				done();
```
dlockhart: Not sure how easy it would be to add a test that actually has a course tile offscreen, verifies that it hasn't loaded, scrolls the page until its visible and then verifies that the load APIs were called? Probably hard.

ryantmer: Ah yeah, should have added that in the description - hard indeed. Gave it an earnest effort, but ended up being a fairly big mess 😢

ryantmer:
Good news! After some discussion with @dlockhart yesterday, he came up with the idea to use `requestIdleCallback` in conjunction with `IntersectionObserver` to take a slightly different approach that is screenreader-friendly.

In short, we immediately load any visible tiles, and use `requestIdleCallback` (or `setTimeout` if not supported) to load all non-visible tiles. This means the screenreader gets the updated tiles and navigation works as expected (NVDA, Chrome and IE11). I've also semi-manually confirmed that this works as intended - unfortunately trying to write tests for this behaviour proved to be rather cumbersome.

I'm still planning to get some performance numbers on this, although also in the conversation last night we realized that this change kinda skews what those performance numbers mean - so that change is upcoming in this PR, otherwise merging this would end up changing what our current perf metrics mean.

dbatiste:
Sounds good.  I think `requestIdleCallback` is closer to the original behavior.  I am curious where `IntersectionObserver` comes into play with the approach..?  

If you decide to set a timeout for the idle callback, you'll want it long enough to allow it to be run in idle-time.  If the main thread is blocked and the timeout is hit, the idle callback will be invoked regardless of any negative performance that may result.  

I'm not sure how much work you might be doing in the callback, but it too should not cause delay.  The callback is passed an API for requesting the estimated idle time available, and if you have a lot of work to do, you may do some, and request a subsequent idle callback to do more work.

As an aside, I've seen recommendation to use `requestIdleCallback` (in demanding scenarios) inside of Intersection observers so that they don't block the main thread.  Rationale being, IntersectionObserver is going to kick in during things like scroll and resize, where it's really important not to block the main thread.

ryantmer:
Yeah, I think the difference here is fairly small, but possibly worth it? The idea was to defer loading of any non-visible course tiles, but anything that;'s visible in the current viewport we want as son as possible. From what I understand, using only requestIdleCallback we'd be delaying the fetching of information of the visible tiles as well - but we want to load those as soon as possible, so the widget is meaningful as soon as possible.

As for the callback, it's a promise of three sequential network requests - not sure how that gets scheduled though tbh; it's obviously a lot of work, but I just kinda assumed that it would be non-blocking while it's waiting for networks responses, is that correct?

dbatiste:
Correct, those network requests are async - they'll get fired and the main thread will continue.  When the responses come back, they will be processed on the main thread in whatever frame they come back in.  The DOM work needs to happen on the main thread, and if you found there was a lot (like more than 50ms), you could choose to chunk it up, for example by using requestAnimationFrame.

(I'm still not certain about "requestIdleCallback in conjunction with IntersectionObserver".  From the description, I wasn't sure where it fits into the solution.  It sounded like the new solution just relies on requestIdleCallback.

ryantmer:
Huh, re-reading my change I realize what you're saying, my implementation makes no sense at the moment. It acted as I expected, delaying non-visible tiles and not delaying visible ones, which I can make sense of in Chrome/Firefox, but unexpected in the polyfilled browsers.

I'll take another look - the ultimate goal is

1. if the course tile is visible, load it immediately
2. if the course tile is not currently visible, load it "later", hopefully causing less of an impact on the tiles that qualify for 1.

ryantmer:
More changes! Results in the following:

image

This is when I've got the window fairly small, so only the first two tiles are visible - those two requests fire immediately, since the tiles are visible, and the other 8 tiles I've got pinned fire a little later (only about 100ms later in this case, granted).

I'm seeing similar results across browsers. I tried the method from the previous commit again and it didn't give this sort of timing graph - everything was actually firing immediately. But I believe the image above should show that this accomplishes both items 1. and 2. from my previous comment :)

dlockhart:
That's really sweet. And when we adjust our performance metrics timing, we'll consider the widget "done" after those first two tiles have rendered, and ignore the other 8.

ryantmer:
Looking into that (and the failed Edge test) right now, in fact!

Although, I've also realized with my current method that for visible tiles, we're setting up the `requestIdleCallback` anyway... but not sure if that's avoidable, as we don't immediately know if the tile is visible, until the observer is called. I'll add a check in there to not bother re-`unobserve`-ing if we've already got `_load = true`.

d2l-course-tile.html
```diff
@@ -238,13 +229,45 @@
...
+							return;
+						}
+						// Whether we load because the tile became visible, or because we got some
+						// idle time, we want to disconnect the observer either way
+						observer.unobserve(tileContainerEl);
+						this._load = true;
+					}.bind(this));
+
+					var observer = new IntersectionObserver(observerCallback.bind(this));
+					observer.observe(tileContainerEl);
```
dlockhart: `tileContainerEl` is defined _inside_ another callback function... which seems not quite right to me. I think it should be declared in the same global function scope.

ryantmer:
The promised performance numbers:

image

In summary, this adds ~100ms on average to the `d2l.my-courses` measure, but since we're not yet sending it anywhere, no biggie. Already started in on making those measurements lazy-loading-aware, which will hopefully bring a new, lower box to the chart.

PR #430, opened 2017-11-23
US91228 - View all courses count

trevorjonesd2l:
This sets the initial page size when loading the widget to 50. If there are less than 50 enrollments it will show the true number, if there are more it'll show 50+. If there are zero it will just show View All Courses, but also it doesn't show that link with zero enrollments because of the alert but hey.

...when ze feature flag is on.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -78,7 +78,7 @@
 				on-mouseover="_createAllCourses"
 				on-focus="_createAllCourses"
 				tabindex="0" >
-				<h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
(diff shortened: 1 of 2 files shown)


d2l-my-courses.html
```diff
@@ -495,6 +501,7 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 
 				this._lastEnrollmentsSearchResponse = enrollmentsEntity;
 				var enrollmentEntities = enrollmentsEntity.getSubEntitiesByClass('enrollment');
+				this._hasMoreEnrollments = enrollmentsEntity.hasLinkByRel('next');
```
njostonehouse: Nice!

PR #432, opened 2017-11-23
US91774 Improve performance measures

ryantmer:
Improves performance measuring with the addition of lazy loading.

TODO:

- [x] Update performance-timings.md
- [x] Handle case where widget isn't initially visible
- [x] Better way of detecting last visible tile?

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -222,7 +222,8 @@
 				'tile-remove-complete': '_onTileRemoveComplete',
 				'course-tile-first-render': '_onPerformanceEvent',
 				'course-tile-organization': '_onPerformanceEvent',
-				'course-image-loaded': '_onPerformanceEvent'
+				'course-image-loaded': '_onPerformanceEvent',
+				'initially-visible-course-tile': '_onInitiallyVisibleCourseTile'
 			},
 			observers: [
 				'_enrollmentsChanged(pinnedEnrollments.length, unpinnedEnrollments.length)',
@@ -279,6 +280,7 @@
 			*/
 
 			_allCoursesCreated: false,
+			_initiallyVisibleCourseTileCount: 0,
 
 			/*
 			* Listeners
@@ -362,6 +364,9 @@
 					this._moveEnrollmentToUnpinnedList(e.detail.enrollment);
 				}
 			},
+			_onInitiallyVisibleCourseTile: function() {
+				this._initiallyVisibleCourseTileCount++;
+			},
 			_onPerformanceEvent: function(e) {
 
 				var data = this._performanceTracking[e.type];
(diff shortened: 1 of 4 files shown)


ryantmer:
Not ready to merge yet, but wanted to get up a graph similar to the one posted in #428, but with improved marks/measures:

image

(Both this graph and the one in the other PR were created using local measurements, over an average of 40 page loads.)

The good news is that this does indeed improve load time, by an average of about 100ms, over not lazy loading! For the graph above, the _No lazy load_ and _Lazy (all tiles)_ data was collected using the `d2l.my-courses` measure; _Lazy (visible tiles)_ used the `d2l.my-courses.visible-images` introduced in this PR.

d2l-my-courses.html
```diff
@@ -380,6 +390,13 @@ <h3 class="d2l-body-standard">[[localize('viewAllCourses')]]</h3>
 				this.performanceMark('d2l.my-courses.complete');
 				this.performanceMeasure('d2l.my-courses', 'd2l.my-courses.attached', 'd2l.my-courses.complete');
 
+				this.performanceMeasure(
+					'd2l.my-courses.visible-images',
+					'd2l.my-courses.attached',
+					'd2l.my-courses.tiles.visible-images.complete'
+				);
+				console.log(window.performance.getEntriesByName('d2l.my-courses.visible-images')[0].duration);
```
dlockhart: Don't forget this guy ^ :)

ryantmer: Hah, whoops. You have seen behind the curtain of how I got the chart above 😜

d2l-my-courses.html
```diff
@@ -361,6 +361,23 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
...
+						'd2l.my-courses.meaningful.visible',
+						'd2l.my-courses.attached',
+						'd2l.my-courses.visible-organizations-complete'
+					);
+					this.performanceMark('d2l.my-courses.visible-images-complete');
+					this.performanceMeasure(
+						'd2l.my-courses',
+						'd2l.my-courses.attached',
+						'd2l.my-courses.visible-images-complete'
+					);
```
dlockhart: Is there a possibility for either of these to fire twice? Like if everything is initially hidden, and then later the user scrolls down to show them all? I think the first one can't fire twice, but I'm less sure about `visible-images-complete`.

ryantmer:
Finished implementing the TODO items I had from Friday, so this is ostensibly ready for review. Things of note:

- using the new `d2l-course-image` which lazy-loads images only once they're in the viewport
- pretty significantly changed how we're tracking important points (see performance-timing.md for more)
- the method to determine when we've finished "visible" stuff makes two assumptions
  - we know how many tiles are visible before the first /organizations request finishes
  - we get responses in order (not a great assumption to make, but I think should be okay here?)

d2l-my-courses.html
ryantmer: `d2l.my-courses.visible-organizations-complete` does indeed fire twice if the widget isn't visible - the first time because `_initiallyVisibleCourseTileCount` is zero, the second when the `_courseTileOrganizationEventCount` hits `pinnedEnrollments.length`.

I do only get one value, though - weirdly, the [docs for `mark`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/mark) indicate that trying to call mark with the same name twice should throw a `SyntaxError`, although I'm not seeing that happen 😕 That said, relying on that is kinda 💩 anyway, so I'll restructure this a bit.

PR #433, opened 2017-11-27
De28292/show start end time on tile notifications

trevorjonesd2l:


ryantmer:
Those failing tests should fix themselves if you rebase, I made them a little less restrictive. Couldn't get them to work otherwise :(

d2l-course-tile.html
```diff
@@ -610,11 +610,13 @@
 
 				if (this._dateFormatter) {
 					var dateObj = new Date(date);
-					this._notificationDate = this._dateFormatter.format(dateObj);
+					var time = dateObj.toLocaleTimeString(this.locale, { hour: 'numeric', minute: 'numeric', hour12 : true });
```
ryantmer: Hmm. `hour12: true` seems... not very i10n-ized. (e.g. France uses 24h time almost exclusively) According to [ze docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString) the default for this will be locale-specific, which I suspect is what we want?

trevorjonesd2l: Good call

PR #436, opened 2017-12-01
switching to use imported d2lIntl library instead of relying on BSI

dlockhart:
This is a temporary first step towards a fix. Eventually, I want this to call into `d2l-localize-behavior` which will contain `d2lIntl`, but in the meantime I need to get it to stop referencing `BSI.Intl` because that's going away.


dlockhart:
Closing this for now as I'm going to delay this change until `d2l-localize-behavior` has better support for user overrides. Will re-open then.

PR #437, opened 2017-12-01
WIP - Unified all courses + some all courses refactoring

trevorjonesd2l:
Waiting for after branching to merge so this PR is going to pretty much be all the unified courses changes. Non ideal for sure. But WIP for now

---

This PR shouldn't change any actual behaviour, but here's the plan:

`d2l-all-courses` as it is right now, will be split up a bit so that two different views can be shown based on the updated my courses flag.

`d2l-all-courses-segregated-content` will be the current all courses view, with pinned and unpinned courses being separated and animations.

`d2l-all-courses-unified-content` will show the unified all courses view with no hard separation between pinned and unpinned courses, just an order.

This PR is the first step into separating `d2l-all-courses` to smaller component so that the two views can be switched easier. There will probably be more PRs coming out to encapsulate all necessary behaviors in the individual components, just trying to make smaller and more frequent PRs since merge conflicts would be nasty with this refactoring.


src/d2l-all-courses-segregated-content.html
```diff
@@ -0,0 +1,131 @@
+<link rel="import" href="../../polymer/polymer.html">
+<link rel="import" href="../../d2l-alert/d2l-alert.html">
+<link rel="import" href="../d2l-alert-behavior.html">
+<link rel="import" href="../d2l-all-courses-styles.html">
+<link rel="import" href="../d2l-course-tile-grid.html">
+<link rel="import" href="../d2l-course-tile-responsive-grid-behavior.html">
+<link rel="import" href="../localize-behavior.html">
+
+<dom-module id="d2l-all-courses-segregated-content">
+	<template>
```
dlockhart: I'd recommend setting `strip-whitespace` on new things you build, just to make sure extra textNodes aren't messing things up:
https://www.polymer-project.org/1.0/docs/devguide/local-dom#strip-whitespace

d2l-all-courses.html
```diff
@@ -211,36 +179,36 @@ <h2>{{localize('unpinnedCourses')}}</h2>
...
 				_hasMoreEnrollments: {
 					type: Boolean,
 					computed: '_computeHasMoreEnrollments(lastEnrollmentsSearchResponse)'
 				},
-				_noPinnedCoursesInSearch: Boolean,
-				_noPinnedCoursesInSelection: Boolean,
-				_noUnpinnedCoursesInSearch: Boolean,
-				_noUnpinnedCoursesInSelection: Boolean,
-				_noUnpinnedCourses: Boolean,
+				noPinnedCoursesInSearch: Boolean,
```
ryantmer: On another note, I wanna say a good chunk of these renames weren't necessary, since they're not part of the `d2l-all-courses` API (e.g. nothing is doing `this.$$('d2l-all-courses').noPinnedCoursesInSearch = true`, so they're still "private" to within `d2l-all-courses`)

trevorjonesd2l:
Unless there are any big conflicts that the refactoring will cause this week, I'm going to hold off merging this in until after branching Friday

d2l-my-courses.html
```diff
@@ -597,30 +598,21 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
...
+				allCourses.filterStandardDepartmentName = this.standardDepartmentName;
+				allCourses.courseUpdatesConfig = this.courseUpdatesConfig;
+				allCourses.updatedSortLogic = this.updatedSortLogic;
+
 				allCourses.open();
 
 				e.preventDefault();
 				e.stopPropagation();
-
-				this.async(function() {
```
dbatiste: Another way to go about this would be to wire up to `animationend` or `transitionend` (depending on which method is being used to do the animation).  In order to keep that implementation choice hidden away, the overlay could dispatch a custom event.  Depending on the scenario, it's even possible to wire-up to the event just-in-time.

trevorjonesd2l:
New PR will pop up later

ryantmer:
rip

PR #442, opened 2017-12-06
Decouple d2l-all-courses

ryantmer:
Rather than directly setting the pinned and unpinned courses, we can instead treat the connection between d2l-all-courses and d2l-my-courses like the connection between d2l-all-courses and the filter - just set the myEnrollmentsEntity, and let the child component take it from there. The downside to this is that we do need to re-fetch things, but a) caching, and b) we no longer require the 600ms delay to make the animation not stutter.

### DIFF
--- a/src/d2l-all-courses.html
+++ b/src/d2l-all-courses.html
@@ -462,13 +462,16 @@
 				this.$['all-courses-scroll-threshold'].clearTriggers();
 			},
 			_myEnrollmentsEntityChanged: function(entity) {
-				if (entity) {
-					var myEnrollmentsEntity = this.parseEntity(entity);
-
-					if (myEnrollmentsEntity.hasActionByName(this.HypermediaActions.enrollments.searchMyEnrollments)) {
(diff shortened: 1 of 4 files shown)


src/d2l-all-courses.html
```diff
@@ -461,6 +462,30 @@ <h2>{{localize('unpinnedCourses')}}</h2>
 				this._hasFilteredEnrollments = this._hasFilteredPinnedEnrollments || this._hasFilteredUnpinnedEnrollments;
 				this.$['all-courses-scroll-threshold'].clearTriggers();
 			},
+			_enrollmentsSearchUrlChanged: function(url) {
+				return this.fetchSirenEntity(url)
+					.then(function(enrollmentsEntity) {
+						this.myEnrollmentsEntity = enrollmentsEntity;
```
ryantmer: This is all fairly-similar logic to what exists in `d2l-my-courses` (but different enough that factoring it out would be rather pointless)

ryantmer:
Hokay, added a couple tests!

PR #446, opened 2017-12-11
US92691 Precursor

ryantmer:
This doesn't actually accomplish the goals of US92691 - instead, it simply moves the existing `d2l-my-courses` component functionality into `d2l-my-courses-content-animated`, and makes `d2l-my-courses` just a passthrough component.

TODO:

- [x] merge #445 and rebase this


src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -0,0 +1,595 @@
+<link rel="import" href="../../../polymer/polymer.html">
```
ryantmer: oh... what the what! I even used `git mv` to attempt to avoid this awfulness... Well, short version is that this is exactly the content of the former `d2l-my-courses.html`, just moved here and with the `dom-module id` and `Polymer is` changed.

PR #449, opened 2017-12-13
US92691 Add new d2l-my-courses-content

ryantmer:
This new content component accomplishes the goal of the US, which was to fill up to 12 enrollments for a user. In particular, we will now fetch all of a user's pinned enrollments, then continue to add enrollments until we hit 12, or they no longer have any enrollments. There are other "new" things in here as well, for example using the new query parameters `?sort=current&promotePins=true`. We also no longer auto-pin in the new logic, so we explicitly set it to false to be safe for the initial request.

Aside from those changes, the component itself is largely the same as the older one, hence breaking out the behaviour in a previous PR.

The idea behind the structure here is that when the `updatedSortLogic` variable is removed, we can just remove `d2l-my-courses-content-animated`, and optionally re-integrate the functionality in `d2l-my-courses-content-behavior` right back into `d2l-my-courses-content` itself.

TODO:

- [x] Add tests
- [x] Wait for `promotePins` bookmarking fix to be merged in the LMS

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content.html
@@ -52,7 +52,7 @@
 					[[item.alertMessage]]
 					<a
 						is="d2l-link"
-						href="Javascript:void(0);"
+						href="javascript:void(0);"
 						hidden$="[[_hasEnrollments]]"
 						on-tap="_refreshPage">[[localize('refresh')]]</a>
 				</d2l-alert>
--- a/src/d2l-my-courses-content/d2l-my-courses-content.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content.html
@@ -6,14 +6,11 @@
 <link rel="import" href="../../../d2l-loading-spinner/d2l-loading-spinner.html">
 <link rel="import" href="../../../d2l-simple-overlay/d2l-simple-overlay.html">
 <link rel="import" href="../../../d2l-image-selector/d2l-basic-image-selector.html">
-<link rel="import" href="../../../d2l-typography/d2l-typography-shared-styles.html">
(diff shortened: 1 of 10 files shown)


src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -0,0 +1,271 @@
...
+			</d2l-loading-spinner>
+		</div>
+
+		<div hidden$="[[!_showContent]]" class="my-courses-content">
+			<template is="dom-repeat" items="[[_alerts]]">
+				<d2l-alert type="[[item.alertType]]">
+					[[item.alertMessage]]
+					<a
+						is="d2l-link"
+						href="Javascript:void(0);"
```
trevorjonesd2l: _mega nit_ `Javascript:void(0);` vs `javascript:void(0);`  like 10 lines down

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -0,0 +1,271 @@
...
+				on-focus="_createAllCourses"
+				tabindex="0">
+				<h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
+			</a>
+		</div>
+
+		<div id="allCoursesPlaceholder">
+		</div>
+
+		<d2l-simple-overlay
```
trevorjonesd2l: we make this twice now, maybe component?

ryantmer: As in create a separate component for this? If so, the one in `d2l-my-courses-content-animated` will be short-lived anyway

trevorjonesd2l:
_I still approve_

PR #450, opened 2017-12-14
Unified/Segregated All Courses

trevorjonesd2l:
Yeah this guys is massive but here's what's going on

Added two new components: `d2l-all-courses-unified-content` and `d2l-all-courses-segregated-content` which the current `d2l-all-courses` will switch between, depending on the value of `updatedSortLogic` (which is controlled by combo LaunchDarkly and config flag).

`d2l-all-courses` will still handle the fetching of enrollments, sorting, searching and whatnot and those enrollments will be passed into the `content` components. This isn't super ideal to me it'd be nice if they could just take the URL so they could fetch the enrollments in the actual `content` components, but that refactor seemed a bit too ambitious for this release - perhaps that'll be a dragon, maybe even a lizard.

`d2l-all-courses-unified-content`/`d2l-all-courses-segregated-content` will each work with the enrollments they're passed, to figure out things like animations, grid sizes, empty states and whatnot. Also alerts.

The new sorts were added into `d2l-all-courses`. They behave differently depending on the `updatedSortLogic` flag.

Tests are a mess, I'd just ignore those for now - I'm working on adding/refactoring them now for all the changes to `d2l-all-courses`. I'll either add those tests as a commit later on this PR or as a follow up PR. I mean they work, but like, they're not organized, and more are needed.

I'm sorry this PR is massive - I wanted to break it into smaller PR chunks initially but fear of merging things before branching last week just made this PR grow. The commits might be more bite sized.


src/d2l-all-courses.html
```diff
@@ -91,57 +93,34 @@
...
-					updated-sort-logic="[[updatedSortLogic]]">
-				</d2l-course-tile-grid>
-
+				<div>
+					<template is="dom-if" if="[[updatedSortLogic]]">
+						<d2l-all-courses-unified-content
+							show-course-code="[[showCourseCode]]"
+							show-semester="[[showSemester]]"
+							course-updates-config="[[courseUpdatesConfig]]"
+							updated-sort-logic="[[updatedSortLogic]]"
```
njostonehouse: We only ever show this view if `updatedSortLogic` is true, right? We probably don't need to pass this in, then?

src/d2l-all-courses.html
```diff
@@ -392,86 +352,98 @@ <h2>{{localize('unpinnedCourses')}}</h2>
...
 					case 'PinDate':
 						langterm = 'sorting.sortDatePinned';
 						queryParameter = '-PinDate,OrgUnitId';
+						promotePins = this.updatedSortLogic;
 						break;
 					case 'LastAccessed':
 						langterm = 'sorting.sortLastAccessed';
-						// Note: Sorting by LastAccessed is done differently LMS-side, hence this being different
-						queryParameter = 'LastAccessed';
+						queryParameter = this.updatedSortLogic ? 'LastAccessed,OrgUnitId' : 'LastAccessed';
```
njostonehouse: I don't think `,OrgUnitId` here will have an effect. `LastAccessed` makes up its own bookmark, so no worries there.

src/d2l-all-courses.html
```diff
@@ -392,86 +352,98 @@ <h2>{{localize('unpinnedCourses')}}</h2>
...
-					default: // Includes 'OrgUnitName', which is the default sort order
-						langterm = 'sorting.sortCourseName';
-						queryParameter = '-PinDate,OrgUnitName,OrgUnitId';
+					case 'EnrollmentDate':
+						langterm = 'sorting.enrollmentDate';
+						queryParameter = 'LastModifiedDate,OrgUnitId';
+						break;
+					case 'Default':
+						langterm = 'sorting.default';
+						queryParameter = 'Current,OrgUnitId';
```
njostonehouse: Similarly I don't think `,OrgUnitId` will make a difference here. `LastAccessed` and `Current` ignore the rest of the sort string.

src/d2l-all-courses.html
```diff
@@ -323,20 +287,21 @@ <h2>{{localize('unpinnedCourses')}}</h2>
...
+				if (this.updatedSortLogic) {
+					this.filteredEnrollments.forEach(function(enrollment) {
+						this._coursesMap[this.getEntityIdentifier(enrollment)] = true;
+					}, this);
+				} else {
+					this.filteredPinnedEnrollments.forEach(function(enrollment) {
+						this._pinnedCoursesMap[this.getEntityIdentifier(enrollment)] = true;
+					}, this);
+					this.filteredPinnedEnrollments.forEach(function(enrollment) {
+						this._unpinnedCoursesMap[this.getEntityIdentifier(enrollment)] = true;
```
ryantmer: ~This mystified me as to why this was like this before - why are we looping over `filteredPinnedEnrollments` twice?~

OH this is actually a bug! This second one should be `filteredUnpinnedEnrollments`, not pinned!

src/d2l-all-courses.html
```diff
@@ -392,86 +352,98 @@ <h2>{{localize('unpinnedCourses')}}</h2>
 			},
 			_onFilterChanged: function(e) {
 				this._searchUrl = e.detail.url;
-				this._filterCount = e.detail.filterCount;
```
ryantmer: By whyfore the rename? 😜

src/d2l-all-courses.html
```diff
@@ -392,86 +352,98 @@ <h2>{{localize('unpinnedCourses')}}</h2>
...
-				this._updateTileSizes();
-			},
 			_onSortOrderChanged: function(e) {
 				var queryParameter, langterm;
+				var promotePins = false;
 
 				switch (e.detail.value) {
+					case 'OrgUnitName':
+						langterm = 'sorting.sortCourseName';
+						queryParameter = this.updatedSortLogic ? 'OrgUnitName,OrgUnitId' : '-PinDate,OrgUnitName,OrgUnitId';
```
ryantmer: With the changes in here (and arguably before), `queryParameter` doesn't make much sense anymore - maybe `sortParameter` or something?

lang/en.html
```diff
@@ -72,6 +74,9 @@
 			'sorting.datePinned': 'Date Pinned',
 			'sorting.department': 'Department',
 			'sorting.lastAccessed': 'Last Accessed',
+			'sorting.enrollmentDate': 'Enrollment Date',
```
ryantmer: Oh, these should have "Sort: " in them - otherwise when the sort menu is closed, it just reads "Default", which is somewhat confusing.

PR #451, opened 2017-12-15
Unified all courses follow-up

trevorjonesd2l:
Pretty much all just PR feedback from last time. No tests yet, another PR.


src/d2l-all-courses-unified-content.html
```diff
@@ -113,10 +88,30 @@
...
+			},
+
+			_updateTileSizes: function() {
+				this._rescaleCourseTileRegions();
+				this._tileSizes = Math.floor(100 / this._numColsOverlay) + 'vw';
+			},
+			_enrollmentsChanged: function() {
+				this._noCoursesInSearch = false;
+				this._noCoursesInSelection = false;
+				if (!(this.filteredEnrollments.length > 0)) {
```
ryantmer: Could ditch the `!` here, to just `if (this.filteredEnrollments.length === 0)` (similarly below)

lang/en.html
```diff
@@ -74,9 +74,8 @@
 			'sorting.datePinned': 'Date Pinned',
 			'sorting.department': 'Department',
 			'sorting.lastAccessed': 'Last Accessed',
-			'sorting.enrollmentDate': 'Enrollment Date',
-			'sorting.current': 'Current',
```
ryantmer: Why no more current sort? We're just using "Default" in that case, ja?

src/d2l-all-courses.html
```diff
@@ -77,12 +77,12 @@
 										<div class="dropdown-content-header">
 											<span>{{localize('sorting.sortBy')}}</span>
 										</div>
-										<d2l-menu-item-radio hidden$="[[!updatedSortLogic]]" class="dropdown-content-gradient" value="Default" text="{{localize('sorting.default')}}"></d2l-menu-item-radio>
-										<d2l-menu-item-radio value="OrgUnitName" class$="{{_showDropdownGradient(updatedSortLogic)}}" text="{{localize('sorting.sortCourseName')}}"></d2l-menu-item-radio>
+										<d2l-menu-item-radio hidden$="[[!updatedSortLogic]]" class="dropdown-content-gradient" value="Default" text="{{localize('sorting.sortDefault')}}"></d2l-menu-item-radio>
```
ryantmer: nbd, but these `localize`s can also be one-way bindings

lang/en.html
trevorjonesd2l: ja

PR #452, opened 2017-12-18
Small d2l-my-courses-content changes

ryantmer:
Just a few smaller things that weren't required for the US, but seemed like a good idea. Notably,

- Move most functionality from `d2l-my-courses-content` into the behavior, so that the `-content-animated` just overrides the few things it needs to override (and extends otherwise)
- Remove now-unnecessary attached/detached functions, since both implementations have the same attached/detached behaviour


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -23,6 +24,21 @@
 				type: Array,
 				value: function() { return []; }
 			},
+			// Array of all enrollments fetched from the API (including those not displayed)
+			_allEnrollments: {
```
ryantmer: This looks far worse than it is - all new code in this file is just stuff I moved over from `d2l-my-courses-content` (so all the red in that file is the green here, without any real modifications).

PR #461, opened 2017-12-21
If user has no enrollments, immediately show content

ryantmer:
Missed this case when I moved when/how we _showContent. In the case where a user doesn't have any enrollments, we'll never hit _onCourseTileOrganization (obviously), so we should immediately show content in that case.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -332,6 +332,10 @@
 			}
 		},
 		_fetchRoot: function() {
+			if (!this.enrollmentsUrl) {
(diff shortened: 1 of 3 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -435,6 +439,12 @@
 
 			this._enrollments = this._enrollments.concat(newEnrollments);
 
+			if (this._enrollments.length === 0) {
```
ryantmer: This is the actual fix here - normally we'd set `_showContent` in `_onCourseTileOrganization`, but for a user without any enrollments, we'll never hit that case.

PR #465, opened 2018-01-02
DE28675 Fix "no unpinned" message when config is off

ryantmer:
Accidentally broke this when the config is off and the user has only unpinned courses. Previously _showContent = true was only happening in the handler when organization requests had finished, but this wasn't working when the config was off (since none of said organizations would be visible, as they're unpinned). Moved the handler to be a bit more like it was before, but still with the worst-case-one-second setTimeout. In other words, the content should become visible either after the last visible organization has loaded, or one second after the enrollments request has finished.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -183,11 +183,6 @@
 					'd2l.my-courses.attached',
 					'd2l.my-courses.visible-images-complete'
 				);
-
-				setTimeout(function() {
-					// At worst, we show content 1s after the first organization has loaded
-					this._showContent = true;
-				}.bind(this), 1000);
 			}
 
 			this._courseTileOrganizationEventCount++;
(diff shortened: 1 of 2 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -357,11 +352,18 @@
...
 				.then(this._fetchEnrollments.bind(this))
-				.catch(function() {
-					this._showContent = true;
-				}.bind(this));
+				.then(function() {
+					// At worst, display content 1s after we fetch enrollments
+					// (Usually set to true before that, in _onCourseTileOrganization)
+					setTimeout(showContent, 1000);
+				})
+				.catch(showContent);
```
njostonehouse: Is this more like a `finally`?

njostonehouse: nvm: It's possible `finally` isn't supported in all browsers.

ryantmer: Huh, I was fairly certain it wasn't a thing outside of third-party implementations, but looks like `Promise.prototype.finally` [is in TC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally) right now. But either way yeah, not supported by IE11/Edge :(

PR #466, opened 2018-01-04
Pin Indicator Animation

trevorjonesd2l:
When pinned, the pin indicator will moonwalk in from the right. The overflow menu will slide out of the way too. Also like, sometimes from the left

### DIFF
--- a/src/d2l-course-tile-styles.html
+++ b/src/d2l-course-tile-styles.html
@@ -194,14 +194,14 @@
 				padding-left: 0;
 				padding-right: 0;
 			}
+			:host-context([dir="rtl"]):not([pinned]) .pin-indicator {
+				margin-left: 0;
+			}
 			.pin-indicator {
 				opacity: 0;
 				margin-top: 10px;
 				line-height: 20px;
-			}
-			:host([pinned]) .pin-indicator {
 				opacity: 1;
-				margin-top: 10px;
 			}
 			.hover-menu {
 				opacity: 0;
--- a/src/d2l-course-tile-styles.html
+++ b/src/d2l-course-tile-styles.html
@@ -14,7 +14,6 @@
 				--scale-fade-min-scale: 1.05;
 			}
 			:host:not([updated-sort-logic]) #pin-indicator-button,
-			:host:not([pinned]) #pin-indicator-button,
 			.d2l-course-tile-hidden {
 				display: none;
 			}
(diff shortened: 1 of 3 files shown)


src/d2l-course-tile-styles.html
```diff
@@ -209,7 +224,7 @@
 				border: none;
 				border-radius: 5px;
 				float: left;
-				transition: color 0.5s, background 0.5s;
+				transition: color 0.5s, background 0.5s, opacity 0.25s, width 0.15s, margin-right 0.25s, padding-left 0.25s, padding-right 0.25s;
```
dbatiste: `width`, `margin`, and `padding` are all layout triggers (which of course results in repainting too), making animation expensive.  If possible, triggering layout should be avoided in animations.  I'm not sure what the required visual effect is, but it might be possible to achieve the same result with transform options such as translate.

https://csstriggers.com/

src/d2l-course-tile-styles.html
```diff
@@ -188,6 +187,22 @@
...
+				padding-right: 0;
+			}
+			:host-context([dir="rtl"]):not([pinned]) .pin-indicator {
+				margin-left: 0;
+			}
+			.pin-indicator {
+				opacity: 0;
+				margin-top: 10px;
+				line-height: 20px;
+				opacity: 1;
```
dbatiste: Looks like `opacity` is specified here twice.

src/d2l-course-tile-styles.html
trevorjonesd2l: Unfortunately after looking that doc I don't think it's possible to achieve this sort of animation without a layout trigger

src/d2l-course-tile-styles.html
```diff
@@ -14,7 +14,6 @@
 				--scale-fade-min-scale: 1.05;
 			}
 			:host:not([updated-sort-logic]) #pin-indicator-button,
-			:host:not([pinned]) #pin-indicator-button,
```
ryantmer: Is this ID no longer being used? If so, could be removed. Not a biggie tho

trevorjonesd2l: When the flag is `off` it will `display: none`. When it is `on` but `not pinned` it `visibility: hidden`. I did some a11y testing and this and with things switched to `visibility: hidden` I couldn't access the pin indicator with keyboard/screen reader.

trevorjonesd2l: So I just moved `:host:not([pinned]) #pin-indicator-button` somewhere else pretty much since it's won't `display: none` anymore

PR #467, opened 2018-01-04
DE28697 Fire d2l-pinned-course-change from View All Courses

ryantmer:
This allows the main widget to update its list of courses when a user pins or unpins something in the View All Courses overlay. The good news is that in this case, we can pass the updated enrollment directly via the event, which means we don't need to do the org unit ID fiddling. #nice.


njostonehouse:
Are we targeting this change for the current release?

ryantmer:
I do believe so! The other half of this I don't think so, since we're undecided on what we want to do in the case where you pin/unpin in the widget itself.

PR #469, opened 2018-01-05
app-localize-behavior for dates/times formatting

dlockhart:
A small change, but 65 million years in the making. Almost.

The major change here is to switch to using `d2l-app-localize-behavior` instead of `d2lIntl` to format the dates and times. It uses `d2lIntl` under the covers, but will also automatically pick up any org/user custom locale overrides, such as 24-hour clock and D2L-specific date/time formats.

In addition, the "{date} at {time}" format was removed in favour of `.formatDateTime()`. By using our built-in date-time formatting, we're not introducing unnecessary language translations or setting any precedents. It'll also "just work" in translations instead of relying on our translators to do the right thing.


lang/ar.html
```diff
@@ -83,9 +82,9 @@
 			'courseTile.updates': 'التحديثات:',
 			'coursePinButton': 'تم تثبيت {course}. إلغاء تثبيت المقرر التعليمي',
 			'overlay.courseClosed': 'Closed',
-			'overlay.courseClosedOn': 'Closed on {date} at {time}',
+			'overlay.courseClosedOn': 'Closed {date} {time}',
```
dlockhart: These will eventually get replaced with actual translations, I just made this consistent with the English value.

src/d2l-course-tile.html
```diff
@@ -613,30 +610,34 @@
...
+					dateObj,
+					{'format': 'medium'}
+				);
+				var courseEndedTerm = this.updatedSortLogic
+					? (type === 'courseEnded') ? 'overlay.courseClosedOn' : 'overlay.courseOpeningOn'
+					: (type === 'courseEnded') ? 'overlay.courseEndedOn' : 'overlay.courseStartingOn';
+				this._courseNotificationLabel = this.localize(
+					courseEndedTerm,
+					'date',
+					this.formatDate(dateObj, {format: 'MMMM d, yyyy'}),
```
dlockhart: The reason I'm not using `.formatDateTime()` here and still passing separate dates and times (but without the "at") is that it's using a custom date format with the long "MMMM" (e.g. January) month name. If I use the "long" date-time format to get the long month name, you also get a timezone name unfortunately.

test/d2l-course-tile/d2l-course-tile.js
```diff
@@ -663,23 +661,17 @@ describe('<d2l-course-tile>', function() {
...
 		beforeEach(function() {
-			curDate = Date.now();
-			futureDate = new Date(curDate + 8000).toISOString();
-			pastDate = new Date(curDate - 8000).toISOString();
-			formattedFutureDateTime = formattedDate + ' at ' + (new Date(futureDate)).toLocaleTimeString('en', { hour: 'numeric', minute: 'numeric', hour12 : true });
-			formattedPastDateTime = formattedDate + ' at ' + (new Date(pastDate)).toLocaleTimeString('en', { hour: 'numeric', minute: 'numeric', hour12 : true });
+			futureDate = new Date(3000, 0, 1, 15, 5).toISOString();
+			pastDate = new Date(1900, 3, 30, 4, 38).toISOString();
+			formattedFutureDateTime = 'Jan 1, 3000 at 3:05 PM';
+			formattedPastDateTime = 'Apr 30, 1900 at 4:38 AM ';
```
dlockhart: Just simplified this a bit since it's a lot harder to mock out the Intl stuff. If we're still around when this test breaks in the year 3000, I volunteer to fix it.

bower.json
```diff
@@ -36,10 +36,9 @@
     "d2l-icons": "^3.7.0",
     "d2l-image-selector": "git://github.com/Brightspace/image-selector#^2.2.0",
     "d2l-intersection-observer-polyfill-import": "Brightspace/intersection-observer-polyfill-import#^1.0.0",
-    "d2l-intl-import": "^1.0.0",
```
ryantmer: Is this a change that is intended to be applied everywhere, meaning `d2l-intl-import` could be deprecated at some point? (Or does it still have value in some cases?)

dlockhart: In theory they could all go away, although `d2l-localize-behavior` still uses it, as does BSI itself, so the actual component can't ever go away.

dlockhart: Also, it's pretty new -- I added it as part of this exercise so I don't think very much is using it yet.

PR #471, opened 2018-01-08
DE28699 Include past courses in limit-to-12 count (sometimes)

ryantmer:
Bit of an oversight here - when a user has many enrollments, they are sorted by last accessed date. This means there's no `hide-past-courses` being applied to the course tile grid, which is correct. However, we were missing the CSS rule for including these (visible) past courses in the `limit-to-12` CSS logic. This meant that all recently-accessed ended courses were not contributing to the 12 limit, so the user ended up seeing more than 12.

### DIFF
--- a/src/d2l-course-tile-grid-styles.html
+++ b/src/d2l-course-tile-grid-styles.html
@@ -30,6 +30,7 @@
 		}
 
 		:host[hide-past-courses] d2l-course-tile[past-course]:not([pinned]),
+		:host[limit-to-12]:not([hide-past-courses]) d2l-course-tile:not([pinned]):nth-child(n+13),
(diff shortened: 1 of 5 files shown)


trevorjonesd2l:
Can we test this stuff?

ryantmer:
I will say yes we can! Coming up.

ryantmer:
Oh wow there are _no_ tests for `d2l-course-tile-grid`... I'll add some (especially for what I've added here), but my current Dragon Slayer stuff actually obsoletes it anyway.

ryantmer:
Travis is still being fussy - I'll wait until this PR (and master) are both back to green before merging.

PR #473, opened 2018-01-09
Fix Travis using sudo: required

ryantmer:
See https://github.com/travis-ci/travis-ci/issues/8836#issuecomment-356362524 for explanation.


ryantmer:
Yay, this fixes the issue we were seeing. Going to just merge.

PR #474, opened 2018-01-10
Plugged in the new translations

trevorjonesd2l:


lang/de.html
```diff
@@ -55,9 +55,13 @@
 			'filtering.noRoles': 'Sie haben keine Rollenfilter ausgewählt.',
 			'filtering.roles': 'Rollen',
 			'overlay.courseStarting': 'Kurs beginnt',
-			'overlay.courseStartingOn': 'Kurs beginnt am {date} {time}',
+			'overlay.courseStartingOn': 'Kurs beginnt am {date} um {time}',
```
dlockhart: Seems like they put an "at" between the date and time here? "um" in German -> "around" in English.

lang/fr.html
```diff
@@ -55,9 +55,13 @@
 			'filtering.noRoles': 'Vous n\'avez aucun filtre de rôles.',
 			'filtering.roles': 'Rôles',
 			'overlay.courseStarting': 'Le cours commence',
-			'overlay.courseStartingOn': 'Le cours commence le {date} {time}',
+			'overlay.courseStartingOn': 'Le cours commence le {date} à {time}',
```
dlockhart: French also put an "at" in here.

lang/de.html
trevorjonesd2l: Yeah, maybe we should just remove this?

dlockhart: I think so, yes, and same for the French version.

ryantmer: I don't speak German, but can confirm that removing it in French isn't problematic. It's the same level of weird as removing it in English, i.e. a little non-conversational but that's about it.

PR #477, opened 2018-01-12
Dragon Slayer 

ryantmer:


ryantmer:
Nope, this is ridiculous. Going to break this up into smaller chunks and hide this behind a config.

PR #478, opened 2018-01-13
First stage of new d2l-course-image-tile (US90524)

ryantmer:
This version includes most of the basic functionality that d2l-course-tile has, but excludes functionality around course updates (fetching/displaying), pin/unpin (actually changing the state, that is), inactive/ended/not-yet-started overlays, started-inactive-pinned notification and alert, and setting/uploading course images. These will come in stages over the next PR(s).

Currently, you must set the `updated-course-tile` attribute on `d2l-my-courses` itself in order for this new component to be used in place of the old one. (Upcoming LP PR for this, too.)

TODO:

- [x] Merge LP PR with LD flag (optional really, this just won't do anything without it)
- [x] Wait for release of `d2l-tile` and update bower.json

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content.html
@@ -80,29 +80,32 @@
 						on-tap="_refreshPage">[[localize('refresh')]]</a>
 				</d2l-alert>
 			</template>
-			<div class="course-tile-grid" hidden$="[[!cssGridView]]">
-				<template is="dom-repeat" items="[[_enrollments]]">
-					<d2l-course-image-tile
-						enrollment="[[item]]"
-						tile-sizes="[[_tileSizes]]"
-						show-course-code="[[showCourseCode]]"
-						show-semester="[[showSemester]]"
-						course-updates-config="[[courseUpdatesConfig]]">
-					</d2l-course-image-tile>
-				</template>
-			</div>
-			<d2l-course-tile-grid
-				hidden$="[[cssGridView]]"
-				limit-to-12
-				enrollments="[[_enrollments]]"
-				tile-sizes="[[_tileSizes]]"
-				locale="[[locale]]"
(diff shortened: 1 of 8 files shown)


src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -57,7 +80,19 @@
...
+						enrollment="[[item]]"
+						tile-sizes="[[_tileSizes]]"
+						show-course-code="[[showCourseCode]]"
+						show-semester="[[showSemester]]"
+						course-updates-config="[[courseUpdatesConfig]]">
+					</d2l-course-image-tile>
+				</template>
+			</div>
 			<d2l-course-tile-grid
+				hidden$="[[cssGridView]]"
```
trevorjonesd2l: Should this be `dom-if`d?

ryantmer: Could be, functionally the same(ish, minus the fact that this still gets rendered). Using the `dom-if` can cause annoyances with tests, though - in this case, [trying to select the grid in `attached`](https://github.com/Brightspace/d2l-my-courses-ui/pull/478/files#diff-452fd5cbd10e4a56bc65217e2d9fe72bR103) wasn't working, as it hadn't yet rendered 😢

trevorjonesd2l: What would the performance impact be of rendering both? :-/

ryantmer: Hm, I guess it would be as bad as rendering two tile grids. So, probably not a good idea. I'll fiddle with the `dom-if`! :)

ryantmer:
Hokay, merging - note that this behaviour is default-off for now, unless you enable the `us90524-my-courses-css-grid-layout` flag in LD. (Will remain default-off for at least this sprint.)

PR #481, opened 2018-01-15
Only skip tests on OSX

ryantmer:
These test randomly fail on all browsers on OSX for whatever reason - can't reproduce it locally, so most-recent fix was to just skip the problematic ones. A slightly-more-specific fix here is to only skip them when they're running on OSX, so that we can at least run them on the Windows browsers in Sauce.


ryantmer:
Apparently these tests are just destined to not work on Sauce 😢 Oh well.

PR #482, opened 2018-01-16
De28784 dont duplicate pinned enrollments

trevorjonesd2l:
That map got lost in the refactor (ironic), added back a map, hello map

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-animated.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-animated.html
@@ -128,6 +128,10 @@
 				_viewAllCoursesText: {
 					type: String,
 					computed: '_getViewAllCoursesText(_hasMoreEnrollments, _allEnrollments)'
+				},
+				_pinnedEnrollmentsMap: {
+					type: Object,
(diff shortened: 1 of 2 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -318,6 +328,7 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 					var unpinnedEnrollmentId = this.getEntityIdentifier(this._pinnedEnrollments[index]);
 					if (unpinnedEnrollmentId === enrollmentId) {
 						this.splice('_pinnedEnrollments', index, 1);
+						if (this._pinnedEnrollmentsMap.hasOwnProperty(enrollmentId)) delete this._pinnedEnrollmentsMap[enrollmentId];
```
ryantmer: Not a biggie probably, but iirc `delete` is slower than `= undefined` or `= null`, so instead of `hasOwnProperty/delete`, you can just use `map[enrollmentId]/= undefined`.

PR #483, opened 2018-01-17
Add Finnish translation

ryantmer:
This has been requested by a client; the initial set of translations went out a while ago, hence the new ones being English. I've checked this locally and setting `lang="fi"` (or with a locale, e.g. `lang="fi-fi"`) on the `<html> element does indeed work.


ryantmer:
Kuinka paljon?

PR #488, opened 2018-01-22
(DE28747) Update started inactive alert

trevorjonesd2l:
This makes it so we only show the started but inactive alert if the user has access to the course info link

### DIFF
--- a/src/d2l-course-tile.html
+++ b/src/d2l-course-tile.html
@@ -168,6 +168,12 @@ the user has in that organization - student, teacher, TA, etc.
 					value: false,
 					observer: '_updatedSortLogicChanged'
 				},
+				// Indicates whether the course has the info urls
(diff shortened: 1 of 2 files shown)


src/d2l-course-tile.html
```diff
@@ -532,10 +538,13 @@
...
 
-				this._checkDateBounds(organization);
-
 				var courseInfoUrl = organization && organization.getLinkByRel(this.HypermediaRels.courseOfferingInfoPage);
 				this._courseInfoUrl = courseInfoUrl ? courseInfoUrl.href : null;
+				if (this._courseInfoUrl) {
+					this.hasCourseInfoUrl = true;
+				}
+
+				this._checkDateBounds(organization);
```
ryantmer: Tired-brain is not allowing me to figure this out - did this need to be moved?

PR #489, opened 2018-01-23
US90524 IE11 support (CSS grid)

ryantmer:
This allows for the CSS grid layout to work in IE11. Unfortunately IE11 uses an old CSS grid spec, so we only use this when required, but it works, and is ignored by all other browsers. This does include a bit of a hack, but only in IE - sometimes it takes a moment for the _enrollments array to get populated, and therefore there are no course tiles to select. In this case, we simply re-_onResize a few times (for up to a second) until things get properly rendered. (Tested this fairly extensively - it only occurs once in a while, but the provided solution worked in all cases.)

This change did require ditching the `grid-gap`, in favour of `grid-gap-column` and `margin-bottom` on the course tiles. This is because the method by which below-the-fold tiles are hidden has changed slightly. If we continue to use grid-gap, it adds space between the last visible course tile and the View All Course link, since the tile-surrounding divs are still technically visible (albeit empty). The new method is pretty much visually identical, and prevents the extra space.

Two other small changes:

- Increased the tile width to 350px for breaking - this avoids the circumstance where a two-column layout occurs and the "Course Closed <date>"-type overlay is too large for the tile.
- Added _pinPressHandler - otherwise tabbing off the pin button will always unpin the course (forgot this in the initial version of d2l-course-image-tile)

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -308,16 +308,43 @@
 					}
 				}.bind(this));
 		},
-		_onResize: function() {
+		_onResize: function(ie11retryCount) {
 			var courseTileGrid = this.$$('.course-tile-grid');
 
 			var containerWidth = this.$$('.spinner-container').offsetWidth;
-			var numColumns = Math.min(Math.floor(containerWidth / 300), 4) + 1;
+			var numColumns = Math.min(Math.floor(containerWidth / 350), 4) + 1;
 			courseTileGrid.classList.remove('columns-1');
 			courseTileGrid.classList.remove('columns-2');
 			courseTileGrid.classList.remove('columns-3');
 			courseTileGrid.classList.remove('columns-4');
 			courseTileGrid.classList.add('columns-' + numColumns);
+
+			var cssGridStyle = document.body.style['grid-template-columns'];
+			// Can be empty string, hence the strict comparison
+			if (cssGridStyle !== undefined) {
+				return;
+			}
+
+			var courseTileDivs = this.querySelectorAll('.course-tile-grid > div');
+			ie11retryCount = ie11retryCount || 0;
+			if (
+				ie11retryCount < 10
+				&& courseTileDivs.length === 0
+			) {
+				// If course tiles haven't yet rendered, try again for up to one second
+				// (only happens sometimes, only in IE)
+				setTimeout(this._onResize.bind(this, ie11retryCount++), 100);
+				return;
+			}
+
+			for (var i = 0; i < courseTileDivs.length; i++) {
+				var div = courseTileDivs[i];
+				// The (* 2 - 1) accounts for the grid-gap-esque columns
+				var column = (i % numColumns + 1) * 2 - 1;
+				var row = Math.floor(i / numColumns) + 1;
+				div.style['-ms-grid-column'] = column;
+				div.style['-ms-grid-row'] = row;
+			}
 		},
 		_onStartedInactiveAlert: function(e) {
 			this._removeAlert('startedInactiveCourses');
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -313,11 +313,14 @@
 
 			var containerWidth = this.$$('.spinner-container').offsetWidth;
 			var numColumns = Math.min(Math.floor(containerWidth / 350), 4) + 1;
-			courseTileGrid.classList.remove('columns-1');
-			courseTileGrid.classList.remove('columns-2');
-			courseTileGrid.classList.remove('columns-3');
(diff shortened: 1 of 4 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -306,16 +306,43 @@
...
+
+			var courseTileDivs = this.querySelectorAll('.course-tile-grid > div');
+			ie11retryCount = ie11retryCount || 0;
+			if (
+				ie11retryCount < 10
+				&& courseTileDivs.length === 0
+			) {
+				// If course tiles haven't yet rendered, try again for up to one second
+				// (only happens sometimes, only in IE)
+				setTimeout(this._onResize.bind(this, ie11retryCount++), 100);
```
jstefaniuk-d2l: Would it be better perhaps for tiles to say when they have been rendered (or before rendering) via an event, and then you give them a class at that point?

ryantmer: My inclination against that is that this is only ever hit by IE anyway, and it keeps this code fairly localized - I feel like using an event/class would touch more places.

That said, you've made me realize that perhaps I can just add a call to _onResize when `_enrollments` first goes from 0 to n elements... hmm...

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -306,16 +306,43 @@
...
 			var containerWidth = this.$$('.spinner-container').offsetWidth;
-			var numColumns = Math.min(Math.floor(containerWidth / 300), 4) + 1;
+			var numColumns = Math.min(Math.floor(containerWidth / 350), 4) + 1;
 			courseTileGrid.classList.remove('columns-1');
 			courseTileGrid.classList.remove('columns-2');
 			courseTileGrid.classList.remove('columns-3');
 			courseTileGrid.classList.remove('columns-4');
 			courseTileGrid.classList.add('columns-' + numColumns);
+
+			var cssGridStyle = document.body.style['grid-template-columns'];
```
dbatiste: I gather the idea here is that if the browser does not support CSS grid, we can detect via presence of `grid-template-columns`, and the rest of the code here is for browsers that do not support?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
ryantmer: Huh, turns out two things are slow actually, so my approach above won't work - the initial call to `_onResize` (from `attached`) happens before the `_enrollments` have been populated, so things won't properly update there. Calling it again when `_enrollments` changes also doesn't work, since the browser hasn't yet rendered the `d2l-course-image-tiles` themselves at that point. It _does_ work if I do something like

```
_enrollmentsChanged: function() {
    setTimeout(this._onResize.bind(this), 100);
}
```

but ultimately that's back to using a `setTimeout` anyway, and on a slower machine that 100ms might not be enough anyway 😢

(I'll fully admit the approach I've taken isn't great, but my goal was to keep it as not-affecting-other-browsers as possible.)

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -306,16 +306,43 @@
...
 			var courseTileGrid = this.$$('.course-tile-grid');
 
 			var containerWidth = this.$$('.spinner-container').offsetWidth;
-			var numColumns = Math.min(Math.floor(containerWidth / 300), 4) + 1;
+			var numColumns = Math.min(Math.floor(containerWidth / 350), 4) + 1;
 			courseTileGrid.classList.remove('columns-1');
 			courseTileGrid.classList.remove('columns-2');
 			courseTileGrid.classList.remove('columns-3');
 			courseTileGrid.classList.remove('columns-4');
 			courseTileGrid.classList.add('columns-' + numColumns);
```
dbatiste: It seems like we could have a scenario where we remove a class and then add the same class back again.  The browser may or may not be smart enough to interpret this as no-op.  I might be inclined to check for the class we're planning to add, and only doing the removal if necessary.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
ryantmer: Yep! Confirmed that the approach works in Firefox and Chrome. The downside to this approach would be if a browser went from failing this check to passing it at some point, but I don't see that happening with IE11 😜 Older versions of Edge will fail the check, and so run the IE stuff below, but I've confirmed on the newest Edge that it passes this check, doesn't run the IE stuff, and still displays properly.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
ryantmer: Good call, this can definitely be smarter about it. It's a pretty heavy change too, so worth avoiding if possible!

PR #491, opened 2018-01-24
US92183 - Use hybrid version of course-image and organization-hm-behavior

svanherk:
This will fail until Brightspace/image-selector#21 is merged and a minor release is made.


svanherk:
Re-ran after merging Brightspace/image-selector#21 and making a minor release, all is green now!

PR #493, opened 2018-01-29
Call _onResize initially to set correct tile size

ryantmer:
Not a huge deal, but forgot this in my previous PR. The 6rem default height is a bit off the correct 43% ratio in some cases (close enough for a default), but if possible then we should set the correct size right away.


ryantmer:
😢

PR #494, opened 2018-01-31
(DE28770) New empty state messages for no courses in semester/department/role.

trevorjonesd2l:
Show a few new messages for when there are no courses in department or semester - this seemed like the simpler approach to figuring out how to show/hide the semesters based on _visible_ courses within that department or semester. Also added this stuff for roles.

The basic new messages are below, but also there are similar messages for pinned/unpinned views
> There are no current or active courses in this department
> There are no current or active courses in this semester
> There are no current or active courses for this role

Changed up `filter-menu` a bit so it returns which specific things are filtered, so the appropriate message can be displayed - `filterCount` is now determined by the consumer (`all-courses`)

I'll add a commit after feedback comes in which will add tests for when each state is displayed.


src/d2l-all-courses.html
```diff
@@ -230,6 +236,18 @@
 					type: Number,
 					value: 0
 				},
+				_departmentFilterCount: {
```
ryantmer: Opinion: rather than three different properties, `_filterCount` could be an object with department/semester/role subproperties. Just means you'd only be dealing with one attribute rather than three (and attribute serialization/deserialization isn't really risky with something so simple)

src/d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -272,9 +276,12 @@
 					parentOrganizations: departmentSemesterFilters.join(',')
 				});
 
+				// include number of department and semester filters
```
ryantmer: Opinion: this comment seems unnecessary

trevorjonesd2l: that was def a TODO

PR #495, opened 2018-02-01
Close overlay when image is changed

ryantmer:
Previously, this was being accomplished by an event listener inside d2l-simple-overlay, but it was specific to this case. So, it will be removed from there, which means we have to explicitly close the overlay when a set-course-image event occurs.

(The change to `d2l-simple-overlay` to remove this behavior is in https://github.com/Brightspace/simple-overlay/pull/16, so this needs to merge + release before that merges + releases.)

See also https://github.com/Brightspace/d2l-image-banner-overlay/pull/54.


ryantmer:
Verified that this works both and without https://github.com/Brightspace/simple-overlay/pull/16

PR #497, opened 2018-02-05
Fix build maybe?

ryantmer:
Went red on master for unknown reasons, seeing

```
$ xvfb-run wct --skip-plugin sauce
/usr/bin/xvfb-run: 184: /usr/bin/xvfb-run: wct: not found
```


ryantmer:
YAY

PR #504, opened 2018-02-14
Hybridization

dlockhart:
The time has come! Hybridization of my-courses.


test/index.html
```diff
@@ -10,27 +10,48 @@
...
+				'./d2l-my-courses-content/d2l-my-courses-content-animated.html?wc-shadydom=true',
+				'./d2l-my-courses-content/d2l-my-courses-content-animated.html?dom=shadow',
+				'./d2l-search-widget-custom/d2l-search-widget-custom.html?wc-shadydom=true',
+				'./d2l-search-widget-custom/d2l-search-widget-custom.html?dom=shadow',
+				'./d2l-touch-menu-item/d2l-touch-menu-item.html?wc-shadydom=true',
+				'./d2l-touch-menu-item/d2l-touch-menu-item.html?dom=shadow',
+				'./d2l-utility-behavior/d2l-utility-behavior.html?wc-shadydom=true',
+				'./d2l-utility-behavior/d2l-utility-behavior.html?dom=shadow',
+				'./localize-behavior/localize-behavior.html?wc-shadydom=true',
+				'./localize-behavior/localize-behavior.html?dom=shadow'
```
dlockhart: Possible optimization here might be to pull the HTML files out into an array, then iterate it creating a new array with each item from the original in there twice, once with `?dom=shadow` appended at the end and the other with `wc-shadydom`.

test/d2l-course-image-tile/undefined
```diff
@@ -0,0 +1 @@
+placeholder to avoid GET errors
```
ryantmer: Hah, excellent idea!

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -362,6 +383,7 @@ describe('d2l-my-courses-content', () => {
 				});
 			});
 
+			/*
```
ryantmer: Oh, I take it `.skip` doesn't work for these array-based tests?

test/index.html
dbatiste: Good point.  :)

test/d2l-my-courses-content/d2l-my-courses-content.js
dbatiste: Hmmm yeah.  I am gonna need to look at this more.

test/d2l-course-image-tile/undefined
dbatiste: It felt hacky, and I wasn't exactly sure what I needed to do to give it an actual name.  My attempts at mocking it were not very fruitful.  I think this is one of those cases where the browser console was showing it was an error but CLI was not.

package.json
```diff
@@ -3,10 +3,11 @@
...
   "private": true,
   "scripts": {
-    "postinstall": "polymer install",
-    "test": "npm run test:lint:js && npm run test:lint:wc && polymer test --skip-plugin sauce",
-    "test:lint:js": "eslint --ext .js,.html .",
-    "test:lint:wc": "polymer lint"
+    "postinstall": "polymer install --variants",
+    "lint": "npm run lint:js && npm run lint:wc",
+    "lint:js": "eslint --ext .js,.html .",
+    "lint:wc": "polymer lint -i *.html ./src/* ./demo/* ./lang/* ./test/*",
```
dbatiste: I found this was necessary to avoid the polymer linter going through all the components in the bower directly.

src/d2l-all-courses.html
```diff
@@ -234,21 +244,26 @@
...
+							semesters: 0,
+							roles: 0
+						};
+					}
+				},
+				_updatedSortLogicInitallyObserved: {
+					type: Boolean,
+					value: false
 				},
-				_updatedSortLogicInitallyObserved: false,
```
dbatiste: This was causing lots of grief with Polymer 2 for lack of a useful error.  Something to be watchful for I think.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -1,5 +1,6 @@
 <link rel="import" href="../../../polymer/polymer.html">
 <link rel="import" href="../../../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
+<link rel="import" href="../d2l-all-courses.html">
```
dbatiste: This appeared to be a missing import.  While it was not causing an error, this component has some code that does `document.createElement('d2l-all-courses')` so it should include an import for it.

src/d2l-search-widget-custom.html
```diff
@@ -37,15 +43,16 @@
...
-				<input
-					is="iron-input"
-					bind-value="{{_searchInput}}"
-					role="combobox"
-					aria-autocomplete="list"
-					placeholder="{{localize('courseSearchInputPlaceholder')}}"
-					on-keydown="_onSearchInputKeyPressed"
-					on-focus="_onSearchInputFocused">
-				</input>
+				<iron-input bind-value="{{_searchInput}}">
```
dbatiste: I am not 100% confident in the value binding here.  This was the only way I seemed to get this to work as expected though.

src/d2l-search-widget-custom.html
```diff
@@ -161,7 +168,9 @@
 				document.body.addEventListener('focus', this._handleFocusBound, true);
 				document.body.addEventListener('click', this._handleClickBound, true);
 
-				this._listboxes = Polymer.dom(this.root).querySelectorAll('d2l-search-listbox').slice();
+				var nodes = Polymer.dom(this.root).querySelectorAll('d2l-search-listbox');
```
dbatiste: Something to be careful of here: in Polymer 2, this call returns a `NodeList` (which does not have a `slice` method) in the same way that the native call does, whereas Polymer 1 returns an array.  So, this code needed to change slightly to work for both Polymer 1 & 2.

src/d2l-all-courses.html
```diff
@@ -106,7 +116,7 @@
 							filtered-enrollments="[[_filteredEnrollments]]"
 						></d2l-all-courses-unified-content>
 					</template>
-					<template is="dom-if" if$="[[!updatedSortLogic]]">
```
dbatiste: Polymer 2 didn't seem to like this `if$=` very much.  ;)

test/d2l-my-courses-content/d2l-my-courses-content-animated.js
```diff
@@ -376,15 +382,15 @@ describe('d2l-my-courses-content-animated', function() {
 			expect(widget._alerts).not.to.include({ alertName: 'setCourseImageFailure', alertType: 'warning', alertMessage: 'Sorry, we\'re unable to change your image right now. Please try again later.' });
 		});
 
-		it('should show the number of enrollments when there are no new pages of enrollments with the View All Courses link', function() {
+		it.skip('should show the number of enrollments when there are no new pages of enrollments with the View All Courses link', function() {
```
svanherk: The only `d2l-my-courses-content-animated` tests now that are being skipped is this one and the two below, relating to the AllCoursesText.  This is highlighting an actual functional problem in Polymer 2 that I finally tracked down.

A change to `_allEnrollments` was not triggering a re-computation of `_viewAllCoursesText` because you can't override computed properties (https://github.com/Polymer/polymer/issues/4974).  The `_getViewAllCoursesText` function in  `d2l-my-courses-content-animated.html` is used, but it's using the definition in `d2l-my-courses-content-behavior.html` (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html#L73) and waiting for a change to `_hasEnrollments` or `_enrollments`.

@ryantmer Correct me if I'm wrong, but I'm not sure this will ever be seen anyways?  I had to hack `d2l-my-courses.html` file for testing, because this additional number of courses logic only runs if the updates flag is on (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content/d2l-my-courses-content-animated.html#L341).  But 
`d2l-my-courses-content-animated` is only used if the updates flag is off (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/d2l-my-courses.html#L8).

package.json
dlockhart: This might work as well:
https://github.com/Brightspace/d2l-image-banner-overlay/blob/master/package.json#L9

test/d2l-my-courses-content/d2l-my-courses-content-animated.js
ryantmer: You are indeed correct! Yeah I'm not too concerned about lack of tests in `d2l-my-courses-content-animated` anyway, since it's going the way of the dodo soon enough.

PR #506, opened 2018-02-23
US94430 - Use CSS grid in all courses overlay

ryantmer:
Kind of... kind of completely forgot about this. This first factors out the the CSS grid that was introduced in #478/#479/#480 into it's behaviour and styles, and then uses it within the `d2l-all-courses-unified-content` element.

If this is a pain to review, I can split it up - all three commits were actually distinct chunks of work.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content.html
@@ -49,7 +49,8 @@ This is only used if the `US90527-my-courses-updates` LD flag is ON
 				clear: both;
 			}
 			.course-tile-grid:not([hide-past-courses]) > div:nth-child(n+13) > d2l-course-image-tile:not([pinned]),
-			/* If hide-past-courses, past courses are already hidden (in d2l-css-grid-view-styles), so exclude them from the n+13 count */
+			.course-tile-grid[hide-past-courses] > div > d2l-course-image-tile[past-course]:not([pinned]),
+			/* If hide-past-courses, past courses are already hidden, so exclude them from the n+13 count */
 			.course-tile-grid[hide-past-courses] > div:nth-child(n+13) > d2l-course-image-tile:not([pinned]):not([past-course]) {
 				display: none;
 			}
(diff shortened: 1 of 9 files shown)


ryantmer:
Right - my comment in standup wasn't totally correct; this only applies to `d2l-all-courses-unified` anyway, so will have no effect when the config variable is off. Safe changes, whoop!

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -47,35 +48,10 @@
...
-			d2l-course-image-tile {
-				margin-bottom: 0.75rem;
-			}
-			:host([hide-past-courses]) d2l-course-image-tile[past-course]:not([pinned]),
-			div:nth-child(n+13) > d2l-course-image-tile:not([pinned]):not([past-course]) {
-				display:none;
+			.course-tile-grid:not([hide-past-courses]) > div:nth-child(n+13) > d2l-course-image-tile:not([pinned]),
+			/* If hide-past-courses, past courses are already hidden (in d2l-css-grid-view-styles), so exclude them from the n+13 count */
+			.course-tile-grid[hide-past-courses] > div:nth-child(n+13) > d2l-course-image-tile:not([pinned]):not([past-course]) {
+				display: none;
```
trevorjonesd2l: Sooo...

```
if ( ( !hidePastCourses & courseIndex > 12 & !coursePinned )
|| ( hidePastCourses & courseIndex > 12 & ( !coursePinned || !coursePast ) ) ) {
    hideCourse
}
```

Am I reading that right? Do we want to not hide past courses in that second condition?

ryantmer: We do indeed - that's already being done [in `d2l-css-grid-view-styles`](https://github.com/Brightspace/d2l-my-courses-ui/pull/506/files?diff=unified#diff-ef51acdb869f271355164db28f6779f6R34) though - see comment :)

I can duplicate the rule, granted - in hindsight, having the same fairly-complex-rule in two places is probably less-bad than having fairly-complex-rules coming from two different places.

PR #508, opened 2018-02-26
Fire perf measure event

dlockhart:
Two things happening here:

1. I've renamed the `d2l.my-courses` measure to `d2l.my-courses.hero`. Mostly this is so that all the measures are named under `d2l.my-courses.*`, but also because it roughly corresponds to the "hero element" timing event for the widget.
2. This is now firing the `D2LPerformanceMeasure` custom event for the hero measure. This will get captured by BSI's timing debug code, but also the Google Analytics plugin will pick it up (if enabled) and send this measure along to GA.


dlockhart:
Maybe? Like currently until the browsers add some kind of native support for identifying hero elements in your application (like `<d2l-my-element hero>`), tracking timing manually like this is the best we've got.

PR #516, opened 2018-03-27
DE29564 - Fix Polymer 2 Formatting in Edge

svanherk:
My Courses in Edge with Polymer 2 on currently looks like this:
image
This PR fixes that lol.

### DIFF
--- a/src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
+++ b/src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
@@ -57,7 +57,7 @@
 				return;
 			}
 
-			var courseTileDivs = this.querySelectorAll('.course-tile-grid > div');
+			var courseTileDivs = Polymer.dom(this.root).querySelectorAll('.course-tile-grid > div');
 			ie11retryCount = ie11retryCount || 0;
 			if (
 				ie11retryCount < 10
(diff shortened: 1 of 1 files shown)


src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
```diff
@@ -57,15 +57,15 @@
 				return;
 			}
 
-			var courseTileDivs = this.querySelectorAll('.course-tile-grid > div');
+			var courseTileDivs = Polymer.dom(this.root).querySelectorAll('.course-tile-grid > div');
```
svanherk: This was the main issue, the previous method was never finding the `div`s in Polymer 2, so the column and row styles weren't applied.

src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
```diff
@@ -57,15 +57,15 @@
...
 			ie11retryCount = ie11retryCount || 0;
 			if (
 				ie11retryCount < 10
 				&& courseTileDivs.length === 0
 			) {
 				// If course tiles haven't yet rendered, try again for up to one second
 				// (only happens sometimes, only in IE)
-				setTimeout(this._onResize.bind(this, ie11retryCount++), 100);
+				setTimeout(this._onResize.bind(this, ++ie11retryCount), 100);
```
svanherk: Found this while debugging, this was passing 0 every time and just kept looping infinitely trying to find the `div`s it would never find.

ryantmer:
idk, seems pretty "working as intended" to me... 😜

src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
ryantmer: I will admit, this is probably the first time in life and prefix/postfix has been an actual problem 😮

PR #517, opened 2018-03-28
Remove reliance on d2l-tile z-index

ryantmer:
d2l-tile has a z-index of 3 set on the menu to allow for things to be in the default slot, use a negative `top` value, and effectively slide in in between the image and menu layers. This was being done to show the various overlays on the course tiles here, but a better solution seems to be just having the overlay within the same slot as the image itself, thereby removing the need to fiddle with z-indexes in d2l-tile.

This fixes a bug (caused by the z-index: 3) where the menu for a different course tile could show on top of another course tile's menu.

Requires https://github.com/BrightspaceUI/tile/pull/54 to fix the bug, although this change is safe to go in without that.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -44,7 +44,7 @@
     "d2l-polymer-behaviors": "^1.2.0",
     "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#^3.0.2",
     "d2l-simple-overlay": "git://github.com/Brightspace/simple-overlay#^2.0.0",
-    "d2l-tile": "^3.0.2",
+    "d2l-tile": "^3.0.6",
     "d2l-typography": "^6.1.2",
     "d2l-organization-hm-behavior": "git://github.com/Brightspace/organization-hm-behavior.git#^2.0.1",
     "iron-a11y-announcer": "^2.0.0",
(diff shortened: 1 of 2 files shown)


bower.json
```diff
@@ -44,7 +44,7 @@
     "d2l-polymer-behaviors": "^1.2.0",
     "d2l-search-widget": "git://github.com/Brightspace/d2l-search-widget-ui.git#^3.0.2",
     "d2l-simple-overlay": "git://github.com/Brightspace/simple-overlay#^2.0.0",
-    "d2l-tile": "^3.0.2",
+    "d2l-tile": "^3.0.6",
```
ryantmer: This version removes the `z-index: 3` that we were relying on - in a sense it was a breaking change rather than a patch, but it also shouldn't have been there in the first place (my own error), so calling it a "fix".

src/d2l-course-image-tile.html
```diff
@@ -119,7 +119,7 @@
 				flex-direction: column;
 				justify-content: center;
 				position: absolute;
-				top: calc(-1 * var(--course-image-height));
+				top: 0;
```
ryantmer: Despite zero being the default value for `top`, this is for some reason required... without it, the overlay ends up below, on top of the text 😕

ryantmer:
Checked this yesterday with Folio and nothing seemed to break, so going to merge this in! 🤞

PR #519, opened 2018-04-05
DE29424 - Fix pinning in Polymer 2

svanherk:
This fixes the pinning animation when the new sort logic is on (and starts fixing it when it is off - the rest of that fix will be tracked in DE29828).

### DIFF
--- a/src/d2l-course-tile.html
+++ b/src/d2l-course-tile.html
@@ -232,9 +232,6 @@ the user has in that organization - student, teacher, TA, etc.
 				D2L.MyCourses.UtilityBehavior,
 				D2L.PolymerBehaviors.Hypermedia.OrganizationHMBehavior
 			],
-			listeners: {
-				'animationend': '_onAnimationComplete'
-			},
 			observers: [
 				'_fetchEnrollmentData(_load, enrollment)'
 			],
@@ -243,6 +240,7 @@ the user has in that organization - student, teacher, TA, etc.
 					var tileContainerEl = this.$$('.tile-container');
 					document.body.addEventListener('focus', this._onFocus, true);
 					tileContainerEl.addEventListener('focus', this._hoverCourseTile, true);
+					tileContainerEl.addEventListener('animationend', this._onAnimationComplete, true);
 
 					var observerCallback = function(entries, observer) {
 						if (this._load) {
@@ -284,6 +282,7 @@ the user has in that organization - student, teacher, TA, etc.
 				this._onFocus = this._onFocus.bind(this);
 				this._closeDropdown = this._closeDropdown.bind(this);
 				this._hoverCourseTile = this._hoverCourseTile.bind(this);
+				this._onAnimationComplete = this._onAnimationComplete.bind(this);
 				Polymer.IronA11yAnnouncer.requestAvailability();
 			},
 			focus: function() {
--- a/src/d2l-course-tile.html
+++ b/src/d2l-course-tile.html
@@ -278,6 +278,12 @@ the user has in that organization - student, teacher, TA, etc.
 					observer.observe(tileContainerEl);
 				});
 			},
+			detached: function() {
+				var tileContainerEl = this.$$('.tile-container');
+				tileContainerEl.removeEventListener('focus', this._hoverCourseTile, true);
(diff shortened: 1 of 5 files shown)


src/d2l-course-tile.html
```diff
@@ -232,9 +232,6 @@
 				D2L.MyCourses.UtilityBehavior,
 				D2L.PolymerBehaviors.Hypermedia.OrganizationHMBehavior
 			],
-			listeners: {
```
svanherk: The listener was being added, but `_onAnimationComplete` was never run when defined like this.  Adding it to the `attached` callback fixes this.  Will try to reproduce in a simple example to see if it's a Polymer 2 bug.

src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -169,7 +169,7 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 
 			// Override for MyCoursesContentBehavior._onEnrollmentPinnedMessage
 			_onEnrollmentPinnedMessage: function(e) {
-				if (e.target === this) return;
+				if (e.composedPath()[0] === this) return;
```
svanherk: This condition was never returning `false` because of event retargeting in Shadow DOM.  The change asks for the original target instead, which is what shady DOM was returning.

src/d2l-utility-behavior.html
```diff
@@ -14,6 +14,7 @@
 		// Creates a URL with a query from an Action and an object of required parameters
 		createActionUrl: function(action, parameters) {
 			parameters = parameters || {};
+			action = action || {};
```
svanherk: Not sure if this would ever be a problem in production, but I ran into it will debugging.

omsmith: This will just cause it to fail elsewhere (`action.href`). Definitely not valid for this to be passed a non-action.

src/d2l-course-tile.html
```diff
@@ -243,6 +240,7 @@
 					var tileContainerEl = this.$$('.tile-container');
 					document.body.addEventListener('focus', this._onFocus, true);
 					tileContainerEl.addEventListener('focus', this._hoverCourseTile, true);
+					tileContainerEl.addEventListener('animationend', this._onAnimationComplete, true);
```
dlockhart: I think you'll need to remove the listener now in `detached`, otherwise it'll continue to get called even after the element is removed from the DOM. Not necessary for the `listeners` collection but if you do it manually like this you need to clean it up. Probably true for the `focus` listeners here too...

src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -169,7 +169,7 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 
 			// Override for MyCoursesContentBehavior._onEnrollmentPinnedMessage
 			_onEnrollmentPinnedMessage: function(e) {
-				if (e.target === this) return;
+				if (Polymer.dom(e).rootTarget === this) return;
```
svanherk: Originally I had `composedPath()[0]` here, which is what we need to move to eventually once we are completely Polymer 2.  But for hybrid we need to use this still.

svanherk:
Ok, I think this is finally ready to go, sorry for the delay!

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -369,13 +369,21 @@ describe('d2l-my-courses-content', () => {
 						}
 					}
 				);
-			}
+			}*/
```
dlockhart: Looks like commented out function here that's not needed?

svanherk: Yeah linting was complaining that this function isn't used anywhere, but it's used in a bunch of tests that are commented out in this file.  I think they were commented out as part the hybridization work and the goal was to get them fixed up at some point (maybe as part of the upcoming my-courses work).

dlockhart: Gotcha 👍

PR #525, opened 2018-04-25
Fix node 10 travis problem

jstefaniuk-d2l:


ryantmer:
What happens with node 10? Would be good to capture that in the commit as well :octocat:

jstefaniuk-d2l:
It spits out the error: `TypeError: Cannot read property '1' of null`

jstefaniuk-d2l:
image

svanherk:
I did re-run it and it failed again, different timeout this time.  Trying again

PR #526, opened 2018-04-27
Capajon/de29878 ie11 grid overlapping

capajon:


ryantmer:
The failed build here should be fixed by #525

PR #527, opened 2018-05-10
DE29202 - Adjusting overlay date font size

capajon:


src/d2l-course-image-tile.html
```diff
@@ -136,7 +136,7 @@
 				font-weight: bold;
 			}
 			.overlay-date {
-				font-size: 0.8rem;
+				font-size: 0.7rem;
```
ryantmer: I noticed recently that the overlay text in general tends to overflow on small screens - is this going to exacerbate that?

That said, I was seeing a two-column layout, which I suspect means the updated sort was not on on the instance in question (Fleming College)...

PR #533, opened 2018-05-31
US97068 Update sort from enrollmentsSearchAction

ryantmer:
When a set of enrollments is returned, we were previously just updating the local `_enrollmentsSearchAction` from the response. Now, we also want to update the search/filter/sort with the Field values from the returned action. This commit adds this functionality for `sort`, so that an `_enrollmentsSearchAction`'s `sort` field value is applied to the sort menu when it's received.

Also included is a minor fix that is a followup to https://github.com/Brightspace/d2l-search-widget-ui/pull/44, which prevents the live-search-dropdown from opening if the search input value changes programmatically (i.e. by the `_enrollmentsSearchAction` being updated, rather than by somebody clicking/typing in it).

### DIFF
--- a/src/d2l-all-courses.html
+++ b/src/d2l-all-courses.html
@@ -462,7 +462,62 @@ If it is off and the attribute is not added, the `d2l-all-courses-segregated-con
 					return;
 				}
 
-				this._enrollmentsSearchAction = myEnrollmentsEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
+				var searchAction = myEnrollmentsEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
+				this._enrollmentsSearchAction = searchAction;
+
+				if (searchAction && searchAction.hasFieldByName('sort')) {
+					var sortParameter = searchAction.getFieldByName('sort').value;
+					if (!sortParameter) {
+						return;
+					}
+
+					var sortMap = {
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.html
```diff
@@ -470,7 +470,62 @@
...
+				var searchAction = myEnrollmentsEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
+				this._enrollmentsSearchAction = searchAction;
+
+				if (searchAction && searchAction.hasFieldByName('sort')) {
+					var sortParameter = searchAction.getFieldByName('sort').value
+					if (!sortParameter) {
+						return;
+					}
+
+					var sortMap = {
```
ryantmer: This is an unfortunate thing to exist, but not sure there's a way around it... the similar object in `_onSortOrderChanged` is _approximately_ the opposite of this, but ended up being different enough that trying to make them work together ended up being more complicated than it was worth.

ryantmer:
Oh, probably worth illustrating - there isn't a tabbed view here yet, this is more of a precursor-to-US97068, but doesn't make all that much sense without the requirements of US97068. Just splitting things up to make reviewing easier :)

src/d2l-all-courses.html
njostonehouse: Yep, gross. Something to consider pulling out later. The less we are referencing real data fields, it seems to me, the better.

PR #534, opened 2018-06-05
Tabs my courses

jlai-d2l:
Adding tabs to My Courses widget, and support for Department/Semester filtering.


d2l-my-courses.html
```diff
@@ -1,7 +1,11 @@
+<script src="../webcomponentsjs/webcomponents-lite.js"></script>
```
ryantmer: I suspect this is from testing ja? Shouldn't be required here.

d2l-my-courses.html
```diff
@@ -30,26 +34,37 @@
...
+								standard-semester-name="[[standardSemesterName]]"
+								course-updates-config="[[courseUpdatesConfig]]"
+								course-image-upload-cb="[[courseImageUploadCb]]"
+								css-grid-view="[[cssGridView]]"
+								updated-sort-logic="[[updatedSortLogic]]">
+							</d2l-my-courses-content>
+						</d2l-tab-panel>
+					</template>
+				</d2l-tabs>
+			<!-- </div> -->
```
ryantmer: Also from testing?

d2l-my-courses.html
```diff
@@ -30,26 +34,37 @@
...
 		</template>
 	</template>
 	<script>
 		Polymer({
 			is: 'd2l-my-courses',
 			behaviors: [
-				D2L.MyCourses.MyCoursesBehavior
+				D2L.PolymerBehaviors.MyCourses.LocalizeBehavior,
```
ryantmer: Can these other behaviours be added to `D2L.MyCourses.MyCoursesBehavior` itself instead? :)

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -33,12 +33,20 @@
 				type: Boolean,
 				value: false
 			},
+			promotedSearches: String,
+			tabSearchActions: {
```
ryantmer: It's purely convention, but since this is only accessed locally, underscore-prefixed (`_tabSearchActions`) marks it as "private" (it's not actually private, of course, #becauseJS). Same goes for `_promotedSearches` and `_fetchTabSearchActions` - anything that's not accessed by a parent is underscored, generally.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,31 @@
...
+						var searchAction = enrollmentsRootEntity.getActionByName('search-my-enrollments');
+						this.tabSearchActions = [{'name': searchAction.name, 'title': this.localize('allTab'), 'selected': false, 'enrollmentsSearchAction': searchAction}];
+					}
+					if (promotedSearchesEntity) {
+					 	var promotedSearchesArray = promotedSearchesEntity.actions.map(action => ({'name': action.name, 'title': action.title, 'selected': false, 'enrollmentsSearchAction': action}));
+						this.tabSearchActions = this.tabSearchActions.concat(promotedSearchesArray);
+					}
+				}.bind(this));
+		},
+		_fetchSirenEntityIfExists: function(url) {
```
ryantmer: Hm, I'm surprised by this - `fetchSirenEntity` would just end up passing a falsy URL to `window.fetch`, which would itself do more-or-less the same thing - what's the circumstance where this came up?

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,31 @@
 			return this.updatedSortLogic
 				? this.$$('d2l-my-courses-content').getLastOrgUnitId()
 				: this.$$('d2l-my-courses-content-animated').getLastOrgUnitId();
+		},
+		fetchTabSearchActions: function() {
+			return Promise.all([
+				this._fetchSirenEntityIfExists(this.enrollmentsUrl),
```
ryantmer: What's the rationale behind moving the fetch into this behaviour from `d2l-my-courses-content-behavior`? This behaviour and it's associated component are mostly just a side effect of having two distinct code paths (UpdatedSortLogic on/off), and all it really does is decide which to display, and marshal a few events/calls into the right one. Can `_fetchTabSearchActions` be moved to `d2l-my-courses-content-behavior` instead? (It would require adding the pass-through property, kind of like how `updatedSortLogic` currently works)

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,31 @@
...
+				this._fetchSirenEntityIfExists(this.promotedSearches)
+			])
+				.then(function([enrollmentsRootEntity, promotedSearchesEntity]) {
+					if (enrollmentsRootEntity && !enrollmentsRootEntity.hasAction('search-my-enrollments')) {
+						this.tabSearchActions = [];
+					}
+					else {
+						var searchAction = enrollmentsRootEntity.getActionByName('search-my-enrollments');
+						this.tabSearchActions = [{'name': searchAction.name, 'title': this.localize('allTab'), 'selected': false, 'enrollmentsSearchAction': searchAction}];
```
ryantmer: The call to `this.localize` adds a dependency on the `localize-behavior` - it just so happens that it'll always work here because of the components using this behaviour, but in a "more correct" sense, it should wrap that behaviour.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -245,12 +247,15 @@
 		_onEnrollmentPinnedMessage: function(e) {
 			if (Polymer.dom(e).rootTarget === this) return;
 
-			var enrollment = e.detail.enrollment || this._orgUnitIdMap[e.detail.orgUnitId];
+			var orgUnitId = e.detail.orgUnitId ?
+				e.detail.orgUnitId :
+				this._getOrgUnitIdFromHref(e.detail.enrollment.getLinkByRel(this.HypermediaRels.organization).href);
+
+			// Only want to move pinned/unpinned enrollment if it exists in the panel
+			var enrollment = orgUnitId ? this._orgUnitIdMap[orgUnitId] : null;
```
ryantmer: Not a biggie, but JS "fun": this can be `var enrollment = orgUnitId && this._orgUnitIdMap[orgUnitId]`

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -245,12 +247,15 @@
 		_onEnrollmentPinnedMessage: function(e) {
 			if (Polymer.dom(e).rootTarget === this) return;
 
-			var enrollment = e.detail.enrollment || this._orgUnitIdMap[e.detail.orgUnitId];
+			var orgUnitId = e.detail.orgUnitId ?
+				e.detail.orgUnitId :
+				this._getOrgUnitIdFromHref(e.detail.enrollment.getLinkByRel(this.HypermediaRels.organization).href);
```
ryantmer: It's overly-cautious in a case like this, but with HM responses, we should check `enrollment.hasLinkByRel` before grabbing it, otherwise the `.href` is going to blow up if that link isn't there.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -310,6 +315,23 @@
 				this._addAlert('warning', 'startedInactiveCourses', this.localize('startedInactiveAlert'));
 			}
 		},
+		_onTabSelected: function(e) {
+			// Only handle if tab selected corresponds to this panel
+			if (e.target.id === this.parentElement.id) {
```
ryantmer: Just a personal preference, but switching the condition and `return`ing early reduces the nesting for the rest of this function.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -310,6 +315,23 @@
...
+					this._fetchRoot()
+						.then(function() {
+							window.dispatchEvent(new Event('resize'));
+						});
+				}
+				else {
+					setTimeout(function() {
+						// Force redraw of course tiles.
+						window.dispatchEvent(new Event('resize'));
+					}.bind(this), 10);
```
ryantmer: nit: the `.bind(this)` isn't required here as far as I can tell

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,31 @@
...
 				? this.$$('d2l-my-courses-content').getLastOrgUnitId()
 				: this.$$('d2l-my-courses-content-animated').getLastOrgUnitId();
+		},
+		fetchTabSearchActions: function() {
+			return Promise.all([
+				this._fetchSirenEntityIfExists(this.enrollmentsUrl),
+				this._fetchSirenEntityIfExists(this.promotedSearches)
+			])
+				.then(function([enrollmentsRootEntity, promotedSearchesEntity]) {
+					if (enrollmentsRootEntity && !enrollmentsRootEntity.hasAction('search-my-enrollments')) {
```
ryantmer: I think this should be `!enrollmentsRootEntity || ...`? i.e. if `enrollmentsUrl` was not set, `enrollmentsRootEntity` will be undefined.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -21,6 +21,7 @@
 	*/
 	D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 		properties: {
+			enrollmentsSearchAction: Object,
```
ryantmer: Looking at the change generated by this, we should have done this ages ago - so much nicer than constantly checking if the action exists!

d2l-my-courses.html
jlai-d2l: From the README here, it looked like it was necessary? https://github.com/BrightspaceUI/tabs#component

d2l-my-courses.html
jlai-d2l: I'm an idiot.

d2l-my-courses.html
ryantmer: Oh gotcha - in this case we're okay without, since we operate within BSI, which includes it.

That said, it _is_ kind of an implicit dependency in that case, but I think our general approach is "we assume the place you use our components will include the needed things". In this case, that place is BSI, so I think we're safe not including it.

d2l-my-courses.html
jlai-d2l: So you want me to move the reference to LocalizeBehavior into MyCoursesBehavior?

ryantmer: Yeah, both the `LocalizeBehaviour` and `UtilityBehaviour`. Effectively, it'll make any anything that implements `MyCoursesBehaviour` also implement the other two by default, which is what we'd want. Otherwise, if something implemented `MyCoursesBehaviour` right now and _didn't_ do so, then it could blow up on any things that exist in those two behaviours.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
jlai-d2l: This `fetch` fetches the tabs data and searchActions required for the tabs component, which exists above the `d2l-my-courses-content-behavior.` The `d2l-my-courses-content-behavior` takes the `searchAction` from this fetch and uses it to fetch the actual enrollments...unless I'm misunderstanding you.

d2l-my-courses.html
```diff
@@ -30,26 +34,37 @@
...
-				show-semester="[[showSemester]]"
-				standard-department-name="[[standardDepartmentName]]"
-				standard-semester-name="[[standardSemesterName]]"
-				course-updates-config="[[courseUpdatesConfig]]"
-				course-image-upload-cb="[[courseImageUploadCb]]"
-				css-grid-view="[[cssGridView]]"
-				updated-sort-logic="[[updatedSortLogic]]">
-			</d2l-my-courses-content>
+				<d2l-tabs>
+					<template slot="tab" items="[[tabSearchActions]]" is="dom-repeat">
```
dbatiste: Shouldn't need the slot attribute here.

d2l-my-courses.html
```diff
@@ -30,26 +34,37 @@
...
+						<!-- item.name is an OrgUnitId, and querySelector does not work with components with ids that start with a number -->
+						<d2l-tab-panel slot="panel" id="panel-[[item.name]]" text="[[item.title]]" selected=[[item.selected]]>
```
dbatiste: Can remove the slot attribute from this guy too.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
dbatiste: Yep, that convention can definitely help folks who come later to know what is part of public API and what is not.  Aside from the convention, for reusable/shared components we use the iron demo pages to auto generate docs, and it will categorize public vs private based on this convention as well.  Sometimes you may also see double-underscore in a behavior to prevent conflict with possible members on the consuming component.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
ryantmer: Riiight, yep! Makes sense, I was getting the hierarchy backwards in my head.

Someday we'll be able to remove `updatedSortLogic` and this will all become so much simpler... but it is not this day 😢

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
dbatiste: Seems like we're sort of tricking the browser into thinking there was a resize event to cause it to do some work, presumably wired up to a resize handler.  If yes, is there a method that could be called more explicitly trigger that logic?  If that resize logic changed in some way, it might have an adverse affect on this code.

ryantmer: Yeah, this is an approach I've taken for a couple other things. In this case, I believe it's because switching to a new tab doesn't cause the course images to resize correctly. From what I recall, it's because the parent container doesn't have a calculated size when the image-resizing happens, so it ends up calculating that its size should be 0px :(

src/d2l-my-courses-content/d2l-my-courses-behavior.html
jlai-d2l: `promotedSearches` is an attributed passed to the `d2l-my-courses` component from the LE. Should it still be `_promotedSearches`? And if so, how do I reconcile that with the `promoted-searches` attribute being passed in?

ryantmer: Ah yep, ignore me - `promotedSearches` is a-okay as is :)

lang/en.html
```diff
@@ -9,6 +9,7 @@
 		en: {
 			'advancedSearch' : 'Advanced Search',
 			'viewAllCourses' : 'View All Courses',
+			'allTab' : 'All',
```
ryantmer: Missed this - this should be added (untranslated) to the all of the language files, then we point the translators at this repo and they'll pick up the changes (after this is merged, that is) :)

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,38 @@
...
+		},
+		_fetchTabSearchActions: function() {
+			return Promise.all([
+				this.fetchSirenEntity(this.enrollmentsUrl),
+				this.fetchSirenEntity(this.promotedSearches)
+			])
+				.then(function(values) {
+					var enrollmentsRootEntity = values[0];
+					var promotedSearchesEntity = values[1];
+					if (!enrollmentsRootEntity || !enrollmentsRootEntity.hasAction('search-my-enrollments')) {
```
ryantmer: Rather than using strings for these names/rels/etc., we've made a repo of constants for them, which are shared across many projects - [d2l-hm-constants-behavior](https://github.com/Brightspace/d2l-hm-constants-behavior). In this case, this should be `this.HypermediaActions.enrollments.searchMyEnrollments`. Should have caught this in the last review, sorry I didn't mention it earlier!

Did these constants need to be a Polymer behaviour? Nope, certainly not. Hindsight is 20/20 and we should have just have made it an object 😁

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,38 @@
...
+						this._tabSearchActions = [];
+					}
+					else {
+						var searchAction = enrollmentsRootEntity.getActionByName('search-my-enrollments');
+						this._tabSearchActions = [{'name': searchAction.name, 'title': this.localize('allTab'), 'selected': false, 'enrollmentsSearchAction': searchAction}];
```
ryantmer: nit: no need for the quotes here or below, they can just be 

```js
this._tabSearchActions = [{
  name: searchAction.name,
  title: etc.
]};
```

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -90,11 +91,11 @@
 		},
 		attached: function() {
 			this.performanceMark('d2l.my-courses.attached');
-			this._fetchRoot();
 
 			this._onEnrollmentPinnedMessage = this._onEnrollmentPinnedMessage.bind(this);
 			document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
 			document.body.addEventListener('set-course-image', this._onSetCourseImage.bind(this));
+			document.body.addEventListener('d2l-tab-panel-selected', this._onTabSelected.bind(this));
```
ryantmer: This didn't occur to me earlier, but I believe this should be a `listener` rather than here. By defining it within `listeners`, it'll automatically only listen for this event if it's coming from within this component's DOM (also sets up the listen/unlisten automatically), whereas using `document.body.addEventListener` means it'll fire if anything on the page fires it.

`d2l-course-pinned-change` needs to be here as it's actually coming from the host page - not sure what `set-course-image` is doing here, that could be a mistake actually...

src/d2l-utility-behavior.html
```diff
@@ -50,6 +50,9 @@
 			return window.D2L.Hypermedia.Siren.Parse(entity);
 		},
 		fetchSirenEntity: function(url) {
+			if (!url) {
```
ryantmer: Yeah, this is legit. w.r.t. the discussion about where it should reject - anything that's calling `fetchSirenEntity` probably has a `.then` anyway, which will itself explode. Sounds bad, but I think in this case it's good - if there's a use case for being able to handle it with a `catch`, then we can burn that bridge when we cross it.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -48,6 +56,38 @@
...
+					}
+					if (promotedSearchesEntity && promotedSearchesEntity.actions) {
+						var promotedSearchesArray = promotedSearchesEntity.actions.map(function(action) {
+							return { 'name': action.name, 'title': action.title, 'selected': false, 'enrollmentsSearchAction': action };
+						});
+						this._tabSearchActions = this._tabSearchActions.concat(promotedSearchesArray);
```
ryantmer: Oh, minor thing here as well - we end up assigning `_tabSearchActions` twice within `_fetchTabSearchActions`. From a perf standpoint, it's preferable to build the array of all actions we want first, then assign it, as otherwise we'd end up restamping the tabs twice. (I think restamping is the right word... either way, we render the tabs for the first value of `_tabSearchActions`, then immediately re-render it for the final value).

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -18,45 +63,51 @@ describe('d2l-my-courses', () => {
...
 	describe('Public API', () => {
 		it('should call d2l-my-courses-content-animated.courseImageUploadCompleted', done => {
 			component.updatedSortLogic = false;
-			setTimeout(() => {
-				var stub = sandbox.stub(component.$$('d2l-my-courses-content-animated'), 'courseImageUploadCompleted');
-				component.courseImageUploadCompleted();
-				expect(stub).to.have.been.called;
-				done();
-			});
+			Polymer.dom.flush();
```
ryantmer: Sweet, so this works! Good to know, we can probably get rid of a lot of `setTimeout` crap in the tests in that case.

On that note, `done` is no longer required here, as the test is now synchronous (can be left as-is though, no harm)

test/d2l-my-courses-content/d2l-my-courses-content-animated.js
```diff
@@ -429,7 +448,11 @@ describe('d2l-my-courses-content-animated', function() {
 				}
 			);
 
-			it('should focus on view all courses link when focus called initially', function() {
+			it.skip('should focus on view all courses link when focus called initially', function() {
+				widget.fetchSirenEntity.withArgs(sinon.match('/enrollments/users/169?search=')).returns(Promise.resolve(
```
jlai-d2l: At some point, I could really use another pair of eyes on this test. I'm unable to figure out why it fails, but right now, there's still a bunch of work left to do.

PR #535, opened 2018-06-05
US97633 Remove css-grid-view

ryantmer:
This property was being set by a feature flag, us90524-my-courses-css-grid-layout, which has been on in production for several months. By removing this flag, we remove a few code paths and simplify the codebase a bit.

There does still have to be some awareness of cssGridView, as the old tile view (`d2l-my-courses-content-behavior` and `d2l-all-courses-segregated-content`) still use the non-CSS-grid view and old course tile.


ryantmer:
Point 1: I suspect this will fail Sauce tests, since they bein' reaaaaal flakey right now. Going to look into that now.

Point 2: Ignoring whitespace in the diff makes this seem much, much less scary.

PR #537, opened 2018-06-08
Move alerts

ryantmer:
Previously, both `d2l-all-courses-segregated-content` and `d2l-all-courses-unified-content` had basically the same implementation of the same thing - being able to handle when a course image fails to be set correctly. While removing the duplicate code is nice, the actual reason for this change is to make adding the tabs to this view easier - it's just easier to have the alerts exist in one place - the over-arching `d2l-all-courses` component - rather than having it in two places, and needing to keep both up to date.

### DIFF
--- a/src/d2l-all-courses-segregated-content.html
+++ b/src/d2l-all-courses-segregated-content.html
@@ -38,12 +38,6 @@ This is only used if the `US90527-my-courses-updates` LD flag is OFF
 			[[localize('noPinnedCoursesInRole')]]
 		</span>
 
-		<template is="dom-repeat" items="[[_alerts]]">
-			<d2l-alert type="[[item.alertType]]">
-				[[item.alertMessage]]
-			</d2l-alert>
-		</template>
-
 		<d2l-course-tile-grid
 			id="all-courses-segregated-pinned-tile-grid"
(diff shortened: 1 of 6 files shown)


test/d2l-all-courses-segregated-content/d2l-all-courses-segregated-content.js
```diff
@@ -81,41 +81,6 @@ describe('d2l-all-courses-segregated-content', function() {
 		sandbox.restore();
 	});
 
-	describe('setting course image', function() {
```
ryantmer: These tests were basically just moved, to the `d2l-all-courses` tests.

src/d2l-all-courses-segregated-content.html
```diff
@@ -169,20 +163,9 @@ <h2>[[localize('unpinnedCourses')]]</h2>
 				return this._itemCount;
 			},
 			setCourseImage: function(details) {
-				this.removeSetCourseImageFailure();
```
ryantmer: This code block (and the similar code from `d2l-all-courses-unified-content`) was just moved to `d2l-all-courses` itself, with basically no modification.

Issue #507, opened 2018-02-24
Assumes organization-image entity is embedded


omsmith:
Dunno if this or https://github.com/Brightspace/course-image is "at fault", but if organization-image is an embedded link on the organization, then we try to fetch `undefined` via writing it out as the src/srcset on the `img`.

omsmith:
https://github.com/Brightspace/d2l-my-courses-ui/blob/dd91cd099290cffabcd782f4cc556f1df0282f5a/src/d2l-course-tile.html#L723-L725

omsmith:
https://github.com/Brightspace/d2l-my-courses-ui/blob/dd91cd099290cffabcd782f4cc556f1df0282f5a/src/d2l-course-tile.html#L384,L395

Not completely related (different code-path), but this assumes `getDefaultImageLink` and `getImageSrcSet` returned something useful.

omsmith:
https://github.com/Brightspace/course-image/blob/f264e99e8625c08a197c2e59c81012d68b922c58/d2l-course-image.html#L138-L140

And similarly, but somewhat more relatedly, course-image assumes it was passed in a full entity (could be a legit assumption, but it doesn't verify it) and the `getDefaultImageLink` and `getImageSrcSet` return something useful.

ryantmer:
Looks like [`d2l-organization-hm-behavior`](https://github.com/Brightspace/organization-hm-behavior/blob/master/d2l-organization-hm-behavior.html#L44-L61) is the underlying issue within `d2l-course-image`. Not sure if it should also work with image links, or leave fetching the entity up to the consumer...

ryantmer:
Hokay, resolved I believe, with https://github.com/Brightspace/organization-hm-behavior/pull/16, https://github.com/Brightspace/course-image/pull/18, and https://github.com/Brightspace/d2l-my-courses-ui/pull/509.

Note that these fixes just remove the assumption; they don't add support for linked entities. I figure that's probably it's own thing? In general we don't have a lot of places where we're supporting linked entities (`if (entity.href) { fetchTheEntity.then(stuff); }`), but we certainly could.

Also worth noting that `d2l-course-tile` is being replaced by `d2l-course-image-tile`, which only requires one of the two fixes mentioned above.

PR #538, opened 2018-06-08
Remove cache bust and extra `embedDepth`

ryantmer:
There were cache-busting headers on the request to fetch an organization, but this isn't actually necessary - first, because the request is only cached on the page (via `d2l-fetch`) anyway, so navigating away and then back will reload the organization response. In addition, it's only the image entity we wanted to cache-bust anyway, which leads to...

...not fetching the image right away. Rather than tacking an `embedDepth=1` on to the Link (which is dirty), we should be able to handle both cases: a hydrated image entity, and a linked image entity. It's actually slightly preferable to fetch the linked organization entity, since we don't even _need_ the image initially - we only load it when the course tile scrolls into view, which it may not ever do.

The performance impact of this is immediately noticeable - while it doesn't really speed up initial load much (only really by the amount of time embedding the course image entity took, which wasn't much), it does make opening the All Courses overlay significantly faster, as we're able to use the (likely already fetched) organization and organization image responses, rather than re-fetching them.

### DIFF
--- a/src/d2l-course-image-tile.html
+++ b/src/d2l-course-image-tile.html
@@ -479,9 +479,9 @@ This is used in `d2l-my-courses-content` (when the `us90524-my-courses-css-grid-
 				this._imageLoadingProgress = true;
 
 				var organizationHref = organization.getLinkByRel('self').href;
-				organizationHref += '?embedDepth=1';
 
-				return this._fetchOrganization(organizationHref)
+				return this.fetchSirenEntity(organizationHref)
+					.then(this._handleOrganizationResponse.bind(this))
 					.then(this._displaySetImageResult.bind(this, true, true))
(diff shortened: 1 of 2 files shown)


ryantmer:
This is a left-field PR, sorry about that - just something that I noticed as I was implementing the tabs in All Courses: every time its opened, it would re-fetch every organization response, which adds a significant overhead, and _plenty_ of jank. By only loading them once, we get a more single-page-app feel to the whole experience! Initial load is about the same, but moving into/around the All Courses overlay is much snappier. For example, if you apply a filter, then clear it, the clearing is almost immediate!

omsmith:
Fixes https://github.com/Brightspace/d2l-my-courses-ui/issues/507?

ryantmer:
Does indeed fix #507! I had a vague memory of that issue existing. Thanks :D

PR #539, opened 2018-06-12
Adding functionality to get/set last saved search.

jlai-d2l:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -566,6 +569,16 @@
 			} else {
 				this._removeAlert('onlyPastCourses');
 			}
+		},
+		_setLastSearchName: function(id) {
+			var formData = new FormData();
+			formData.append('mostRecentEnrollmentsSearchType', '0');
```
njostonehouse: FYI, this maps to `.None` which is correct in this case. It should basically be safe to pass in 0 here without risking some sort of bad parameter on the backend.

test/d2l-my-courses/d2l-my-courses.js
```diff
@@ -3,6 +3,7 @@ describe('d2l-my-courses', () => {
 		sandbox,
 		enrollmentsHref = '/enrollments/users/169',
 		promotedSearchHref = '/promoted-search-url',
+		lastSearchHref = 'homepages/components/1/user-settings/169',
```
njostonehouse: This pattern is likely to change, which I don't think matters for the test, but wanted to mention it.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -58,20 +59,25 @@
...
+				this.fetchSirenEntity(this.promotedSearches),
+				this.fetchSirenEntity(this.userSettingsUrl)
 			])
 				.then(function(values) {
 					var enrollmentsRootEntity = values[0];
 					var promotedSearchesEntity = values[1];
+					var userSettingsEntity = values[2];
 					var tabSearchActions = [];
+					var mostRecentSearchName = userSettingsEntity && userSettingsEntity.properties ?
+						userSettingsEntity.properties['MostRecentEnrollmentsSearchName'] : null;
```
njostonehouse: If there's a way to make this case-insensitive, it might be useful.

njostonehouse: Okay. Yuck. Leave it, then!

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -58,20 +59,25 @@
...
-				this.fetchSirenEntity(this.promotedSearches)
+				this.fetchSirenEntity(this.promotedSearches),
+				this.fetchSirenEntity(this.userSettingsUrl)
 			])
 				.then(function(values) {
 					var enrollmentsRootEntity = values[0];
 					var promotedSearchesEntity = values[1];
+					var userSettingsEntity = values[2];
 					var tabSearchActions = [];
+					var mostRecentSearchName = userSettingsEntity && userSettingsEntity.properties ?
```
ryantmer: Of no importance whatsoever, but this can just collapse down to

```js
var mostRecentSearchName = userSettingsEntity
    && userSettingsEntity.properties
    && userSettingsEntity.properties.MostRecentEnrollmentsSearchName
```

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
ryantmer: I'm kind of confused here - would setting the type to `None` not be equivalent to clicking the "Group by: None" option in the widget settings?

jlai-d2l: As the `mostRecentEnrollmentsSearchName` is an unique orgUnitId, we don't actually need to depend on `mostRecentEnrollmentsSearchType`. You are right in that it is equivalent to "Group By: None" Per Noel, there may be some changes to this property in the future, and so it's safer to just not use it. However, the action requires the parameter to be included, otherwise the request will fail.

ryantmer: Ah, right! Yeah that makes sense, I was thinking the `None` would take precedence and clear the value, effectively.

njostonehouse: It does, but until the UI cares about the Type value, it can effectively be anything. This isn't the value that's checked for which tab to use: The promoted filter type is set by the Admin, not the user.

njostonehouse: A future change will take the Type into account (I hope), but for now it doesn't matter.

PR #540, opened 2018-06-12
US97068 - All Courses tabs

ryantmer:
This adds tabs to the All Courses view. These tabs should match the ones being used in the main widget, as they are just passed in as attributes. No real surprises here - switching tabs fetches the courses associated with that tab and updates the `d2l-allcourses-unified-content` view.

Something I realized partway through this work: when a user has their enrollments already grouped by semester/department via the My Courses widget settings, we aren't able to then use the filter menu to further filter them, as it does a logical OR on all the `parentOrganizations` that are sent to the API. When this is the case, we can simply hide those two tabs. Similarly, if the user has tabs to group their enrollments by RoleAlias, we should hide the Role tab in the filter menu.

~Note that this is WIP-ish, as I haven't looked at tests yet, and this requires both #538 and #539 to be in first.~

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-behavior.html
@@ -76,16 +76,19 @@
 						this._enrollmentsSearchAction = searchAction;	// For d2l-my-courses-content-animated
 						tabSearchActions = [{name: searchAction.name, title: this.localize('allTab'), selected: false, enrollmentsSearchAction: searchAction}];
 					}
-					if (promotedSearchesEntity && promotedSearchesEntity.actions) {
-						var promotedSearchesArray = promotedSearchesEntity.actions.map(function(action) {
-							var selected = action.name === mostRecentSearchName;
-							return { name: action.name, title: action.title, selected: selected, enrollmentsSearchAction: action };
-						});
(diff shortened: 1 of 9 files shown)


jlai-d2l:
Batiste might have a philosophical argument with you about how you're using (not using) `tab-panels`, but given how coupled the search-and-filter row is, and how `on-tab-selected` modifies the `searchUrl` to trigger re-fetching of enrollments, this is probably the best way to handle it.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -83,6 +84,8 @@
 						tabSearchActions = tabSearchActions.concat(promotedSearchesArray);
 					}
 					this._tabSearchActions = tabSearchActions;
+					// TODO: Populate with actual value from response
```
njostonehouse: This should be merged now, so you can get a follow-up PR running soon.

ryantmer:
From some follow-up discussion, turns out this isn't a great idea - there are a bunch of semantics that I'm missing out on here by doing this. In particular, the tab content is no longer going to look like tab content to screen readers. Going to investigate further into doing it the way you did it, @jlai-d2l (i.e. the correct way), but in a way that doesn't even try to render the other tabs until they're clicked.

More commits coming!

PR #541, opened 2018-06-13
US97068 Save last search

ryantmer:
When a user clicks a tab, we save that as the most recent search. This behaviour exists in both the widget view and the all courses view - an attempt to combine the behaviours was made, but it ended up being a significantly larger change for very little benefit.

### DIFF
--- a/src/d2l-all-courses.html
+++ b/src/d2l-all-courses.html
@@ -13,15 +13,15 @@
 <link rel="import" href="../../d2l-simple-overlay/d2l-simple-overlay.html">
 <link rel="import" href="../../d2l-tabs/d2l-tabs.html">
 <link rel="import" href="d2l-alert-behavior.html">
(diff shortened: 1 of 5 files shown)


ryantmer:
@jlai-d2l pointed out on Slack - I've got this wired up so that All Courses opens at the same tab that you've currently got selected in the widget, but not the opposite (i.e. selecting the tab in the widget that you were on in All Courses, when the latter is closed). Going to look into that - if I can figure out a way to do it easily/quickly, I'll update this, but otherwise I'll merge this and create a DE for it.

njostonehouse:
@ryantmer I think that would be worth logging as a DE and following up on, yeah.

ryantmer:
Hokay, DE30403 created for that situation - in that case, I'll merge and deploy this.

PR #542, opened 2018-06-13
DE30396 Use the use-saved-searches attribute

ryantmer:
Previously, we were always rendering the tabbed view of both the widget and all courses when `updated-sort-logic` was true. However, there is also the feature flag for the tabbed view itself, `use-saved-searches`. This changes both views to also take this flag's value into account.

In effect, the change here is that when the flag is off, the widget and the overlay both need to kick off their respective `_fetchRoot` calls more manually.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -33,29 +33,48 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
 			</d2l-my-courses-content-animated>
 		</template>
 		<template is="dom-if" if="[[updatedSortLogic]]">
-			<d2l-tabs>
-				<template items="[[_tabSearchActions]]" is="dom-repeat">
-					<!-- item.name is an OrgUnitId, and querySelector does not work with components with ids that start with a number -->
-					<d2l-tab-panel id="panel-[[item.name]]" text="[[item.title]]" selected=[[item.selected]]>
-						<d2l-my-courses-content
-							advanced-search-url="[[advancedSearchUrl]]"
-							enrollments-url="[[enrollmentsUrl]]"
-							enrollments-search-action="[[item.enrollmentsSearchAction]]"
-							image-catalog-location="[[imageCatalogLocation]]"
-							show-course-code="[[showCourseCode]]"
-							show-semester="[[showSemester]]"
-							standard-department-name="[[standardDepartmentName]]"
-							standard-semester-name="[[standardSemesterName]]"
-							course-updates-config="[[courseUpdatesConfig]]"
-							course-image-upload-cb="[[courseImageUploadCb]]"
-							updated-sort-logic="[[updatedSortLogic]]"
-							user-settings-url="[[userSettingsUrl]]"
-							tab-search-actions="[[_tabSearchActions]]"
-							tab-search-type="[[_tabSearchType]]">
-						</d2l-my-courses-content>
-					</d2l-tab-panel>
-				</template>
-			</d2l-tabs>
+			<template is="dom-if" if="[[useSavedSearches]]">
+				<d2l-tabs>
+					<template items="[[_tabSearchActions]]" is="dom-repeat">
+						<!-- item.name is an OrgUnitId, and querySelector does not work with components with ids that start with a number -->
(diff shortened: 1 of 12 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -265,9 +264,6 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 					this._addAlert('call-to-action', 'noCourses', this.localize('noCoursesMessage'));
 				}
 			},
-			_enrollmentSearchActionChanged: function() {
```
ryantmer: This has moved to `d2l-my-courses-content-behavior`, as both `d2l-my-courses-content` and `d2l-my-courses-content-animated` need it now, in the case where `use-saved-searches` is false.

d2l-my-courses.html
```diff
@@ -33,29 +33,48 @@
...
-							course-image-upload-cb="[[courseImageUploadCb]]"
-							updated-sort-logic="[[updatedSortLogic]]"
-							user-settings-url="[[userSettingsUrl]]"
-							tab-search-actions="[[_tabSearchActions]]"
-							tab-search-type="[[_tabSearchType]]">
-						</d2l-my-courses-content>
-					</d2l-tab-panel>
-				</template>
-			</d2l-tabs>
+			<template is="dom-if" if="[[useSavedSearches]]">
```
ryantmer: Non-whitespace diff is useful here - in effect, this is just adding a case for updatedSortLogic on, useSavedSearches off.

src/d2l-all-courses-unified-content.html
```diff
@@ -35,7 +35,7 @@
 			[[localize('noCoursesInRole')]]
 		</span>
 
-		<template is="dom-if" if="[[_renderContents]]">
+		<template is="dom-if" if="[[renderContents]]">
```
ryantmer: Just needed to make this a public property, so that it can be set as an attribute by `d2l-all-courses`.

src/d2l-all-courses.html
```diff
@@ -336,6 +355,15 @@
 						this._unpinnedCoursesMap[this.getEntityIdentifier(enrollment)] = true;
 					}, this);
 				}
+
+				if (!this.useSavedSearches) {
+					this._showTabContent = true;
+					this._searchUrl = this.createActionUrl(this.enrollmentsSearchAction, {
```
ryantmer: This is basically the same as how it worked before adding any of the tab stuff - just need to set `_searchUrl` to an initial value to kick off the first search.

njostonehouse:
`error  Missing semicolon ` in `d2l-my-courses-behavior.html`?

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -45,7 +52,18 @@
...
 		attached: function() {
-			this._fetchTabSearchActions();
+			if (!this.useSavedSearches) {
+				this.fetchSirenEntity(this.enrollmentsUrl)
+					.then(function(enrollmentsRootEntity) {
+						if (!enrollmentsRootEntity.hasActionByName(this.HypermediaActions.enrollments.searchMyEnrollments)) {
+							return;
+						}
+
+						this._enrollmentsSearchAction = enrollmentsRootEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments)
```
njostonehouse: I think this is the missing semicolon.

ryantmer: d'oh, totally thought I had run `npm run lint` before pushing. Fixed!

ryantmer:
@jlai-d2l no kidding 😭 And we're effectively adding a third state here - fortunately, not a configurable one so it'll be gone in a very-finite amount of time (not sure if one release or two?)

njostonehouse:
@ryantmer @jlai-d2l It's possible that we'll want to keep something like this long-term in order to hide the tabs when there aren't any saved searches, or `.None` is chosen as the promoted filter type. 😬

ryantmer:
Oh, true - although it would be a bit different in that case, and less messy (hopefully)

PR #545, opened 2018-06-15
US97924 Improve All Courses tabbed view perf

ryantmer:
Previously, all tabs in the All Courses view were actually sharing the same _filteredEnrollments array. This meant we were only ever dealing with a single array, which has some advantages. The downside, however, is that each time the array is rendered, each tab is going to have to update itself to reflect the new state of the array. This resulted in noticeable jank when selecting a tab with a large number of enrollments.

Instead, only the `filtered-enrollments` of the newly-selected tab should be set/updated whenever the user switches tabs. The upside to this is that there is no more UI lag, even with large numbers of enrollments. The downside is that it is possible, if switching to and from tabs very quickly for the first time (i.e. before they're cached), to end up with the wrong tab being selected for the enrollments you're seeing. Fortunately, this can be fixed by just switching to another tab, then back (as at that point, everything has loaded and is cached, so things are speedy-speedy).

This also means we no longer need to delay rendering of the `d2l-all-courses-unified-content` component via the `renderContents` flag, since when it's initially created it'll always be empty anyway.

### DIFF
--- a/src/d2l-all-courses-unified-content.html
+++ b/src/d2l-all-courses-unified-content.html
@@ -35,21 +35,19 @@ This is only used if the `US90527-my-courses-updates` LD flag is ON
 			[[localize('noCoursesInRole')]]
 		</span>
 
-		<template is="dom-if" if="[[renderContents]]">
-			<div class="course-tile-grid">
-				<template is="dom-repeat" items="[[filteredEnrollments]]">
-					<div>
-						<d2l-course-image-tile
-							enrollment="[[item]]"
-							tile-sizes="[[_tileSizes]]"
-							show-course-code="[[showCourseCode]]"
-							show-semester="[[showSemester]]"
-							course-updates-config="[[courseUpdatesConfig]]">
-						</d2l-course-image-tile>
-					</div>
-				</template>
-			</div>
-		</template>
+		<div class="course-tile-grid">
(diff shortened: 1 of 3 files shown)


src/d2l-all-courses.html
```diff
@@ -242,10 +240,6 @@
 					type: Array,
 					value: function() { return []; }
 				},
-				_filteredEnrollments: {
```
ryantmer: Since we're no longer using the DOM binding, and instead setting `filteredEnrollments` directly on the `d2l-all-courses-unified-content` component, this can be removed.

src/d2l-all-courses-unified-content.html
```diff
@@ -35,21 +35,19 @@
 			[[localize('noCoursesInRole')]]
 		</span>
 
-		<template is="dom-if" if="[[renderContents]]">
```
ryantmer: Non-whitespace diff is useful here - just removed the `dom-if` around everything (along with `renderContents` itself), which was only added in #540 anyway.

ryantmer:
Yes indeed! Yeah, I'm basically doing exactly what you're doing in the widget now - I should have been, all along, but the sort-of-shared-state across the tabs (i.e. `_searchUrl`) lead me astray... but, fixed now, and therefore faster now!

PR #547, opened 2018-06-19
DE30426: Clicking View All Courses multiple times results in all enro…

jlai-d2l:
…llments being returned, rather than results based on tabs.

Changes:
- switched to `_semestersAndDepartmentTabsHidden`, as `_semesterTabHidden` and `_departmentTabHidden` are the same thing.
- the setting of `_semestersAndDepartmentTabsHidden` was done in `open()` of `d2l-filter-menu`, but is needed before. Added an observer, so that the right action and params could be used in `clearFilters()`

### DIFF
--- a/src/d2l-filter-menu/d2l-filter-menu.html
+++ b/src/d2l-filter-menu/d2l-filter-menu.html
@@ -80,7 +80,7 @@ Polymer-based web component for the filter menu.
 
 		<div id="contentView">
 			<div class="dropdown-content-tabs" role="tablist">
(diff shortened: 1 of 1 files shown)


src/d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -350,6 +346,12 @@
 				this.$.departmentsTabButton.setAttribute('aria-pressed', this._departmentsTabSelected);
 				this.$.rolesTabButton.setAttribute('aria-pressed', this._rolesTabSelected);
 			},
+			_tabSearchTypeChanged: function() {
```
ryantmer: Oh man, this is what I should have been doing anyway, really! Nice :)

PR #553, opened 2018-06-20
DE30477 Check returned action for sort

ryantmer:
The action we initially get that translates into a tab in the widget doesn't (necessarily) have the sort order set. However, the updated action we receive after fetching the enrollments for said tab-action should have the updated sort on it. This is what we use to determine whether to hide past courses, so by using the _initial_ action rather than the _updated_ action, the `sort` value was always empty, and we were always showing past courses.


ryantmer:
This will fail for a moment, as I'm referencing a version of `d2l-tile` that doesn't exist yet - I'll re-run CI and merge once https://github.com/BrightspaceUI/tile/pull/57 is merged and released.

PR #557, opened 2018-06-22
DE30485: Show past courses for Semester tabs.

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -528,10 +528,14 @@
 			var newEnrollments = [];
 
 			var searchAction = enrollmentsEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
-			if (
-				searchAction
+			if (searchAction
 				&& searchAction.hasFieldByName('sort')
 				&& searchAction.getFieldByName('sort').value.toLowerCase() === 'current'
+				&& !(searchAction.hasFieldByName('parentOrganizations')
+					&& searchAction.getFieldByName('parentOrganizations').value
```
jlai-d2l: Is there a case where `value` could be '<empty string>'? Is that a case we want this check to pass?

njostonehouse: I don't think we're likely to have that come up. However, to try and be backend-agnostic, I think we want to treat any non-null `parentOrganizations` string as a deliberate field parameter, even if the result isn't going to do a meaningful filter (as in the `''` case). When `parentOrganizations` _is_ `null`, it should represent the `All` tab, where we want to maintain the previous behaviour (to hide past courses).

PR #560, opened 2018-07-05
DE30513 Delay _addOnlyPastCoursesAlert

ryantmer:
This is a bit of a lame duck fix, but the issue we were seeing here was around timing - effectively, good browsers would not show the message when it was needed (_addOnlyPastCoursesAlert would run before the `past-course` attributes had been set). In slow browsers, it was running even before anything was added to the course tile grid, which meant it would always show. To avoid this, we can just delay the calling of the function - worst-case, this means the alert won't show up until a second after we see the content, but in practice it's far less than a second, and almost unnoticeable in any browser.

This also adds a missing `margin-bottom` that was mistakenly not copied over when some previous refactoring was done.


ryantmer:
Going to close, at least for now - after some discussion with Noel, there's almost certainly a better solution here, and we're not in panic mode with this DE or anything, so it's worth doing the right way rather than the quick way!

PR #563, opened 2018-07-10
Fix tests

ryantmer:
A small change to hm-organization-behavior now requires that course image entities have a class on them. This is part of the response spec anyway, so only manifested in the tests, where we're sending in technically-incomplete data.

This should fix master, which broke on this.


ryantmer:
Got past the tests that were failing, merging.

PR #565, opened 2018-07-11
DE30513 Improve "only past courses" alert logic

ryantmer:
Previously, we were calling into a function to recalculate whether or not we should be showing the "You don't have any future or past courses" alert. Instead, we can leverage computed properties to make this more resilient to the timing issues that were causing the alert to show when it shouldn't, and not show when it should. In addition, we can greatly simplify the logic, since this alert should never show when updatedSortLogic is false - i.e. it should never show in `d2l-my-courses-content-animated`.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -41,6 +41,10 @@
 				type: Boolean,
 				value: false
 			},
+			_courseTileOrganizationEventCount: {
+				type: Number,
+				value: 0
+			},
 			// Array of enrollments being displayed by the widget
 			_enrollments: {
 				type: Array,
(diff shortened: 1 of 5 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -41,6 +41,10 @@
 				type: Boolean,
 				value: false
 			},
+			_courseTileOrganizationEventCount: {
```
njostonehouse: What's this for?

ryantmer: Previously this was just a normal JS property on the WC, but in order for a property to be used in the various Polymer-y things (including `computed`), it needs to be defined here. 

Well, "needs". "Is supposed to be", it might actually work either way (but the docs say anything used in a Polymer-y way, should be defined in a Polymer-y way).

PR #566, opened 2018-07-11
DE30513 Clean up updatedSortLogic usage

ryantmer:
Noticed during DE30513 development, but not strictly required for it - it's only really the top-level `d2l-my-courses` component itself, along with `d2l-all-courses` itself, that needs to know what the `updated-sort-logic` value is. For example, `d2l-my-courses-content` is only ever loaded when it's true, and `d2l-my-courses-content-animated` when it's false. This means we can remove a few usage of this value, that do nothing but complicate things.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content.html
@@ -122,6 +122,13 @@ This is only used if the `US90527-my-courses-updates` LD flag is ON
 	<script>
 		Polymer({
 			is: 'd2l-my-courses-content',
+			properties: {
+				// Override for MyCoursesBehavior.updatedSortLogic
+				updatedSortLogic: {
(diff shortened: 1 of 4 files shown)


src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -122,6 +122,13 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 	<script>
 		Polymer({
 			is: 'd2l-my-courses-content',
+			properties: {
+				// Override for MyCoursesContentBehavior.updatedSortLogic
+				updatedSortLogic: {
```
ryantmer: This is really the crux of this (uninteresting) change - this is always going to be true for `d2l-my-courses-content`, and always false (the default value) for `d2l-my-courses-content-animated`.

PR #567, opened 2018-07-11
DE30375: Remove use-saved-searches attribute in favour of checking th…

njostonehouse:
…e tab collection's size.


src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -77,36 +74,40 @@
...
+					var promotedSearchesEntity = values[0];
+					var userSettingsEntity = values[1];
+
+					if (!promotedSearchesEntity || !promotedSearchesEntity.properties) {
+						return;
+					}
+
+					this._tabSearchType = promotedSearchesEntity.properties.UserEnrollmentsSearchType;
+
+					if(promotedSearchesEntity.actions.length < 2 || !promotedSearchesEntity.actions) {
```
ryantmer: Ah JS, where the order of `||` matters!

(The order of the conditions here is backwards 😄)

njostonehouse: Actually I think that condition shouldn't be here yet. Working ahead of myself.

ryantmer: Oh true - this is only needed for the "don't show tabs when only one tab" situation!

njostonehouse:
Tests failing all over the place. Need to reconsider how my refactors work.

PR #569, opened 2018-07-16
DE30375: Make sure that the parentElement of the selected tab exists …

njostonehouse:
…before accesing id.


njostonehouse:
I don't know, but it seems to happen a lot locally.

PR #576, opened 2018-07-24
US98268: Be less opinionated about boolean implementations of behavio…

njostonehouse:
…ur. Computed values of these booleans may not be as deterministic as these tests imply.


njostonehouse:
@ryantmer I don't think these verifications are providing us sufficient value for the work they'd require to keep up to date. Making sure they exist seems sufficient for now.

PR #578, opened 2018-07-25
Nstonehouse/de30375/attach saved searches

njostonehouse:


njostonehouse:
That's correct. The previous PR couldn't pass tests; now we've revisited the tests enough to move forward.

njostonehouse:
Except ```Error: 2 failed tests```. Spoke too soon I guess.

njostonehouse:
Why the tests failed: By moving the "get the All search action" out of the "fetch the tabs search action", it meant that the tests of the latter didn't have the "All" search set. It worked as expected (no All tab), but the tests didn't change with it.

PR #581, opened 2018-07-27
Nstonehouse/de30375/count tabs

njostonehouse:


njostonehouse:
Need to fix up some linting errors, and after looking into the "all" overlay, I should combine this update with an update there too. commits incoming.

njostonehouse:
I keep losing the All tab. Need to reconsider the timing of some things, I guess.

PR #582, opened 2018-07-27
Nstonehouse/us30375/count tabs

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -117,7 +113,18 @@
...
 						enrollmentsSearchAction: action
 					};
-				}));
+				});
+
+				if (!this._enrollmentsSearchAction) {
+					return;
+				}
+
+				this._tabSearchActions = [{
```
ryantmer: This is probably just Monday-morning brain, but I'm confused here - if there is an `_enrollmentsSearchAction`, we just replace all the other search actions? If so, should `_fetchTabSearchActions` also be a `catch` after trying to fetch `enrollmentsUrl` as well, since we can get by without that default search action now?

PR #583, opened 2018-07-27
When node is undefined, trying to access its properties causes JS err…

njostonehouse:
…ors.


njostonehouse:
Manual testing with the console window opening surprises me with JS errors from time to time :)

PR #584, opened 2018-07-30
Majones/us98454/covert d2l enrollment card to entity driven component

m6jones:
Takes away my-courses-widget knowledge of the enrollment card. It now gets the href for widget presentation settings and enrollment entity href and passes that to the enrollment card.

### DIFF
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -40,6 +40,7 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
 							<d2l-my-courses-content
 								advanced-search-url="[[advancedSearchUrl]]"
 								enrollments-url="[[enrollmentsUrl]]"
+								presentation-url="[[presentationUrl]]"
 								enrollments-search-action="[[item.enrollmentsSearchAction]]"
 								image-catalog-location="[[imageCatalogLocation]]"
 								show-course-code="[[showCourseCode]]"
@@ -59,6 +60,7 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
 			</template>
 			<template is="dom-if" if="[[!useSavedSearches]]">
 				<d2l-my-courses-content
+					presentation-url="[[presentationUrl]]"
 					advanced-search-url="[[advancedSearchUrl]]"
 					enrollments-url="[[enrollmentsUrl]]"
 					enrollments-search-action="[[_enrollmentsSearchAction]]"
--- a/d2l-my-courses.html
+++ b/d2l-my-courses.html
@@ -39,16 +39,13 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
 						<d2l-tab-panel id="panel-[[item.name]]" text="[[item.title]]" selected=[[item.selected]]>
 							<d2l-my-courses-content
 								advanced-search-url="[[advancedSearchUrl]]"
+								course-image-upload-cb="[[courseImageUploadCb]]"
 								enrollments-url="[[enrollmentsUrl]]"
-								presentation-url="[[presentationUrl]]"
 								enrollments-search-action="[[item.enrollmentsSearchAction]]"
 								image-catalog-location="[[imageCatalogLocation]]"
-								show-course-code="[[showCourseCode]]"
-								show-semester="[[showSemester]]"
+								presentation-url="[[presentationUrl]]"
 								standard-department-name="[[standardDepartmentName]]"
 								standard-semester-name="[[standardSemesterName]]"
-								course-updates-config="[[courseUpdatesConfig]]"
-								course-image-upload-cb="[[courseImageUploadCb]]"
 								use-saved-searches="[[useSavedSearches]]"
 								user-settings-url="[[userSettingsUrl]]"
 								tab-search-actions="[[_tabSearchActions]]"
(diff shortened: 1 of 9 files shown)


src/d2l-all-courses.html
```diff
@@ -511,7 +508,7 @@
 				var params = {
 					search: search,
 					autoPinCourses: false,
-					embedDepth: 1,
```
njostonehouse: What effect is this going to have? Does it mean we have to hydrate each Organization now (through the card(s))?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -609,7 +598,12 @@
 					method: 'PUT',
 					body: formData
 				}));
-		}
+		},
+		_getEntityIdentifier: function(entity) {
```
ryantmer: This already exists in the `d2l-utility-behavior` - this file could just use that behaviour so it doesn't need to get reimplemented.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -560,16 +559,11 @@
 			}
 
 			enrollmentEntities.forEach(function(enrollment) {
-				var enrollmentId = this.getEntityIdentifier(enrollment);
+				var enrollmentId = enrollment.href;
 				if (!this._existingEnrollmentsMap.hasOwnProperty(enrollmentId)) {
-					newEnrollments.push(enrollment);
```
ryantmer: I'm possibly just being overly cautious here, but I'd definitely make sure that this still works when UpdatedSortLogic is off - if it does, great! And if it does, I have no idea _how_ it does 😃

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -56,13 +61,6 @@
 				return;
 			}
 
-			this.fetchSirenEntity(this.enrollmentsUrl)
-				.then(function(enrollmentsRootEntity) {
-					if (enrollmentsRootEntity.hasActionByName(this.HypermediaActions.enrollments.searchMyEnrollments)) {
-						this._enrollmentsSearchAction = enrollmentsRootEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
```
ryantmer: This appears to still be required - if `promotedSearches` is not set (i.e. not viewing the widget in tabbed mode), nothing will load unless this Action is initially set.

d2l-my-courses.html
```diff
@@ -59,15 +58,15 @@
...
 					enrollments-search-action="[[_enrollmentsSearchAction]]"
 					image-catalog-location="[[imageCatalogLocation]]"
-					show-course-code="[[showCourseCode]]"
-					show-semester="[[showSemester]]"
+					presentation-url="[[presentationUrl]]"
 					standard-department-name="[[standardDepartmentName]]"
 					standard-semester-name="[[standardSemesterName]]"
-					course-updates-config="[[courseUpdatesConfig]]"
-					course-image-upload-cb="[[courseImageUploadCb]]">
+					use-saved-searches="[[useSavedSearches]]"
```
njostonehouse: I don't think we want this line. I thought I'd managed to remove support for the attribute.

src/d2l-my-courses-content/d2l-my-courses-behavior.html
```diff
@@ -54,15 +56,24 @@
 			return groups.length > 2 || (groups.length > 0 && !this._enrollmentsSearchAction);
 		},
 		attached: function() {
-			if (!this.enrollmentsUrl) {
+			if (!this.enrollmentsUrl || !this.userSettingsUrl) {
```
ryantmer: Is `userSettingsUrl` set if `updatedSortLogic` is off (on the LMS)? If not, then this would have to be split apart into two independent promises.

src/d2l-all-courses.html
```diff
@@ -178,7 +174,8 @@
 				/*
 				* Public Polymer properties
 				*/
-
+				// presentation url
```
ryantmer: nit, but this comment is just restating the name of the variable 😜 (Might be worth briefly explaining what "presentation" is, since it's necessarily obvious from the name)

src/d2l-all-courses.html
```diff
@@ -353,7 +350,7 @@
 
 				this._searchUrl = this.createActionUrl(this.enrollmentsSearchAction, {
 					autoPinCourses: false,
-					embedDepth: 1,
+					embedDepth: this.updatedSortLogic ? 0 : 1,
```
ryantmer: Not something that needs to be addressed at this point, but it would probably be a good idea to allow either 0 or 1 here, in the new stuff. By not specifying the value at all here, we let the server decide (whether it's 1, or 0, or even if the query param doesn't exist anymore), and can client-side figure out what to do. In particular, if the server returns un-embedded (linked) entities, then we're good, we just pass the `href`s down to the `d2l-enrollment-card`. If the server decides to start sending back hydrated entities by default, all we have to do is instead grab the `self` link from said hydrated entities, and pass those to the `d2l-enrollment-card`s instead.

Like I said, not something that needs to be addressed now, but is something worth thinking about in general - with HM APIs, the clients are intended to be fairly flexible!

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -29,6 +29,11 @@
 			},
 			tabSearchType: String,
 			userSettingsUrl: String,
+			presentationUrl: {
+				type: String,
+				reflectToAttribute: true,
```
ryantmer: I don't think this actually needs to be reflected to an attribute, does it?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -104,7 +113,8 @@
 			'course-tile-organization': '_onCourseTileOrganization',
 			'course-image-loaded': '_onCourseImageLoaded',
 			'initially-visible-course-tile': '_onInitiallyVisibleCourseTile',
-			'started-inactive': '_onStartedInactiveAlert'
+			'started-inactive': '_onStartedInactiveAlert',
+			'd2l-enrollment-card-fetched': '_onEnrollmentCardAdded'
```
ryantmer: nit, but listener names are generally just `_onEventName` (i.e. `_onD2lEnrollmentCardFetched`)

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -29,6 +29,11 @@
 			},
 			tabSearchType: String,
 			userSettingsUrl: String,
+			presentationUrl: {
+				type: String,
+				reflectToAttribute: true,
+				observer: '_resetPresentationUrlCache'
```
ryantmer: This is purely a convention thing, but generally observers are named `_variableChanged` or similar, i.e. `_presentationUrlChanged` (`*Changed` being for observers, `on*` being for listeners)

src/d2l-my-courses-content/d2l-my-courses-behavior.html
m6jones: It is set and required too.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
m6jones: I agree, when I did this I didn't have a full understanding of what reflectToAttribute did.

PR #585, opened 2018-07-31
DE30375: Hide tabs if there's only an All tab and a duplicate tab tha…

njostonehouse:
…t contains all the same elements.


src/d2l-all-courses.html
```diff
@@ -624,7 +624,7 @@
 				return !!link;
 			},
 			_computeShowGroupByTabs: function(groups) {
-				return groups.length > 0;
+				return groups.length > 2 || (groups.length > 0 && !this._enrollmentsSearchAction);
```
ryantmer: Ah, so if there is an `_enrollmentsSearchAction`, that means we're working with no extra tabs, and therefore the All tab and the other tab(s?) will have the same info?

njostonehouse: If there's _isn't_ an `_enrollmentsSearchAction`, it means all of the remaining tabs represent a "saved search." So we can't conclude they're mutually exclusive: We must show them all.

If there _is_ an `_enrollmentsSearchAction` it means one of the tabs is "All." If there is 1 tab, it's just the All tab, so we don't use tabs. If there are a total of 2 tabs, they are mutually exclusive, and we don't show tabs.

ryantmer: Right, makes sense! I get quite confused as to what `_enrollmentsSearchAction` is like... several times a day 😜

PR #588, opened 2018-08-08
Update d2l-enrollments

ryantmer:
Just fixes a margin, nothing too exciting.


ryantmer:
Not actually required - this change already got pulled into BSI, yay floating versions!

PR #589, opened 2018-08-13
In all courses view, updates, course code and semester name show up.

m6jones:
Removed separator when semesterName was empty
Both these changes are for when updatedSortLogic=false.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-animated.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-animated.html
@@ -267,7 +267,31 @@ This is only used if the `US90527-my-courses-updates` LD flag is OFF
 			/*
 			* Utility/helper functions
 			*/
+			// Override for MyCoursesContentBehavior._openAllCoursesView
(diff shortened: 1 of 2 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-animated.html
```diff
@@ -267,7 +267,31 @@ <h3 class="d2l-body-standard">[[_viewAllCoursesText]]</h3>
 			/*
 			* Utility/helper functions
 			*/
-
+			// Override for MyCoursesContentBehavior._openAllCoursesView
```
ryantmer: Nice - I was thinking initially the base-behaviour's `_openAllCoursesView` would need to be modified, but I much prefer the override approach. (Eventually, someday) removing this will be easier!

m6jones: Yeah, when I first noticed it was getting it from the base-behaviour - I sunk. However, I then realized wait, override!

PR #592, opened 2018-08-15
US99396 Add use-enrollment-cards LD flag

ryantmer:
This flag will cut between the two possible representations of courses - with the flag on, we use the new (but not yet complete) `d2l-enrollment-card`. When it's off (current default), we use the old `d2l-course-imate-tile`. Once we are ready to release the `d2l-enrollment-card` and for it to entirely replace the tile, we can remove this flag.

This is _kind of_ the opposite of https://github.com/Brightspace/d2l-my-courses-ui/pull/574, with a few more changes here and there to make things work. Ultimately though, it's just re-adding a reference to the `1.x-hybrid-components` repo, and making things `useEnrollmentCards`-LD-flag-aware. Also used [this diff](https://github.com/Brightspace/d2l-my-courses-ui/compare/a1e8ed865090df024589471eb8147768c9d21539...master) as a reference, which is the diff between master and when the `d2l-course-image-tile` was removed.


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -532,6 +545,10 @@
 			allCourses.updatedSortLogic = this.updatedSortLogic;
 			allCourses.cssGridView = this.cssGridView;
 			allCourses.presentationUrl = this.presentationUrl;
+			allCourses.showCourseCode = this.showCourseCode;
+			allCourses.showSemester = this.showSemester;
+			allCourses.courseUpdatesConfig = this.courseUpdatesConfig;
```
m6jones: So these don't have properties defined. Is that a problem? Still work and would be easier to cleanup?

ryantmer:
I'll take a look at the pinning! Thought I had checked that, but there are manymanymany different edge cases and such 😰

src/d2l-all-courses-unified-content.html
```diff
@@ -36,13 +37,28 @@
...
-				<div>
-					<d2l-enrollment-card
-						href="[[item]]",
-						presentation-href=[[presentationUrl]]>
-					</d2l-enrollment-card>
-				</div>
+			<template is="dom-if" if="[[useEnrollmentCards]]">
+				<template is="dom-repeat" items="[[filteredEnrollments]]">
+					<div>
+						<d2l-enrollment-card
```
njostonehouse: Shouldn't the `dom-if` be inside the `dom-repeat`?

m6jones: If it was inside the dom repeat then it would be called a lot more? Isn't that bad?

njostonehouse: Hm, I don't know. I would've assumed it would only get it once, but you may be right.

m6jones:
Course Code and Semester names were not showing up on my local build.

src/d2l-all-courses-unified-content.html
ryantmer: I'm actually not sure either - I _think_ what would happen the other way is you'd end up with a set of `n` enrollment cards (or tiles), and a set of `n` empty `template`s!

ryantmer:
To confirm, the pinning/course codes/semester not showing is when the flag is off (and you're seeing `d2l-course-image-tile`s), correct?

src/d2l-all-courses-unified-content.html
njostonehouse: That's true, it's probably best to keep the "hidden" templates to a minimum too. Cool, I'm sold.

m6jones:
Yes, I am not seeing it for the d2l-course-image-tile.

ryantmer:
Gotcha. I'll dig into it - fortunately course code/semester was just me forgetting to pass those attributes down... whoops.

ryantmer:
Updated! Fortunately, small fixes for both problems. 😌

m6jones:
All Courses view is dropping the Course Code and Semester. Otherwise the update is good!

m6jones:
Note that updates are not showing up on both views. (All Courses and the My Courses)

PR #596, opened 2018-08-21
Fixed bug where all my courses would not load.

m6jones:
- Fixed a rogue comma
- a map on an array returned the wrong value as `this` was not defined. Bond the function to this so it would be defined.

### DIFF
--- a/src/d2l-all-courses.html
+++ b/src/d2l-all-courses.html
@@ -663,8 +663,9 @@ If it is off and the attribute is not added, the `d2l-all-courses-segregated-con
 				var enrollmentEntities = enrollments.getSubEntitiesByClass(this.HypermediaClasses.enrollments.enrollment);
 
 				if (this.updatedSortLogic) {
-					var gridEntities = enrollmentEntities.map(value => {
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.html
```diff
@@ -663,8 +663,9 @@
 				var enrollmentEntities = enrollments.getSubEntitiesByClass(this.HypermediaClasses.enrollments.enrollment);
 
 				if (this.updatedSortLogic) {
+					var useEnrollmentCards = this.useEnrollmentCards;
 					var gridEntities = enrollmentEntities.map(function(value) {
-						return this.useEnrollmentCards ? value.href : window.D2L.Hypermedia.Siren.Parse(value);
+						return useEnrollmentCards ? value.href : window.D2L.Hypermedia.Siren.Parse(value);
 					});
```
ryantmer: Ah yeah - this should have a `.bind(this)` on it. Mind doing that instead? (Just a bit more consistent)

ryantmer:
Riiiight, I missed this since I was only looking at the course-tile case anyway! I was wondering how I could have not noticed this... thanks!

m6jones:
I missed this too. Not sure how either...

PR #598, opened 2018-08-22
DE31088: Include the current value of *TabHidden values when they're …

njostonehouse:
…being recalculated.


m6jones:
I am testing it on my instance with group by role selected. I cannot filter by semester or department only role.

njostonehouse:
@m6jones Yep, thank you. This seems to happen when you have insufficient (aliased) roles assigned. Will update this PR.

njostonehouse:
Regrouping and will try again.

PR #599, opened 2018-08-22
DE31088: Include the current value of *TabHidden values when they're …

njostonehouse:
…being recalculated.

### DIFF
--- a/src/d2l-filter-menu/d2l-filter-menu.html
+++ b/src/d2l-filter-menu/d2l-filter-menu.html
@@ -212,7 +212,10 @@ Polymer-based web component for the filter menu.
 					type: String,
 					computed: '_computeSearchPlaceholderText(filterStandardDepartmentName)'
 				},
-				_rolesTabHidden: Boolean,
+				_rolesTabHidden: {
+					type: Boolean,
+					value: false
+				},
 				_semestersTabHidden: {
 					type: Boolean,
 					value: false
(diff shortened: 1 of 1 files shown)


src/d2l-filter-menu/d2l-filter-menu.html
```diff
@@ -358,10 +361,13 @@
...
 			},
 			_tabSearchTypeChanged: function() {
+				this._hideInvalidSearchTabs();
+			},
+			_hideInvalidSearchTabs: function() {
 				// If My Courses is grouped by semesters/departments, don't show either of these tabs
 				var semesterOrDepartmentGrouping = this.tabSearchType === 'BySemester' || this.tabSearchType === 'ByDepartment';
-				this._semestersTabHidden = semesterOrDepartmentGrouping && !this._searchSemestersAction;
-				this._departmentsTabHidden = semesterOrDepartmentGrouping && !this._searchDepartmentsAction;
+				this._semestersTabHidden = semesterOrDepartmentGrouping || !this._searchSemestersAction;
```
njostonehouse: It's possible that this was the whole problem all along - I haven't checked to be totally sure.

ryantmer: You may be onto something! Yeah, the `&&` seems like it would _usually_ be correct, but it does seem like there would be a circumstance where that's not the case... oops.

PR #602, opened 2018-08-29
Remove old codeowners

ryantmer:
Spoke with these folks a while back, all agreed that they were okay with being removed here.

Other people can feel free to add themselves when/if they feel comfortable doing so!


jstefaniuk-d2l:
🎈

PR #605, opened 2018-09-07
Majones/hide completed courses from widget

m6jones:
This actually hides the completed course and still shows 12 enrollments.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -114,7 +114,8 @@
 			'course-image-loaded': '_onCourseImageLoaded',
 			'initially-visible-course-tile': '_onInitiallyVisibleCourseTile',
 			'started-inactive': '_onStartedInactiveAlert',
-			'd2l-enrollment-card-fetched': '_onD2lEnrollmentCardFetched'
+			'd2l-enrollment-card-fetched': '_onD2lEnrollmentCardFetched',
+			'd2l-enrollment-card-status': '_onD2lEnrollmentCardStatus'
 		},
 		attached: function() {
 			this.performanceMark('d2l.my-courses.attached');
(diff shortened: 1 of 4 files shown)


src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -49,9 +49,7 @@
 			.course-tile-grid:not([hide-past-courses]) > div:nth-of-type(n+13) > d2l-enrollment-card:not([pinned]),
 			.course-tile-grid:not([hide-past-courses]) > div:nth-of-type(n+13) > d2l-course-image-tile:not([pinned]),
 			.course-tile-grid[hide-past-courses] > div > d2l-enrollment-card[past-course]:not([pinned]),
-			.course-tile-grid > div > d2l-enrollment-card[completed]:not([pinned]),
 			.course-tile-grid[hide-past-courses] > div > d2l-course-image-tile[past-course]:not([pinned]),
-			/* If hide-past-courses, past courses are already hidden, so exclude them from the n+13 count */
```
m6jones: Removed comment as it is misleading on how this works.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -238,6 +239,25 @@
...
+			) {
+				return;
+			}
+
+			if (e.detail.status.completed) {
+				var index = this._enrollments.indexOf(e.detail.enrollmentUrl);
+				if (index !== -1) {
+					this.splice('_enrollments', index, 1);
+				}
+			} else if (e.detail.status.completed === false && (this._enrollments.indexOf(e.detail.enrollmentUrl))) {
```
ryantmer: If the `indexOf` returns zero, I think we'd bail out of this, which I don't believe is what we want? (i.e. I think you need to check specifically that the index > -1 rather than just falsy here)

m6jones: Oops, good catch. It suppose to be `this._enrollments.indexOf(e.detail.enrollmentUrl)) === -1`

PR #608, opened 2018-09-11
Solving the race condition with alert being added and removed.

m6jones:
I solved the race condition by holding onto the information using an object.
I saw that the when holding all the information in the array, the wrong element could get removed.
To prevent this the object allows you to only change the elements you want to change.

Is there a better way?

### DIFF
--- a/src/d2l-alert-behavior.html
+++ b/src/d2l-alert-behavior.html
@@ -14,34 +14,48 @@
 		properties: {
 			// Array containing alert objects for display
 			_alerts: {
-				type: Array,
-				value: function() { return []; }
+				type: Array
+			},
+			_alertContainer: {
+				type: Object,
+				value: function() { return {}; }
 			}
 		},
+		_alertToArray: function(alertContainer) {
+			this._alerts = Object.keys(alertContainer).map(function(key) {
+				if (!alertContainer[key]) {
+					return null;
+				}
+				return {
+					alertType: alertContainer[key].alertType,
+					alertName: key,
+					alertMessage: alertContainer[key].alertMessage
+				};
+			}.bind(this)).filter(function(element) {
+				return element;
+			});
+		},
 		_addAlert: function(type, name, message) {
-			if (this._alerts) {
-				this.push('_alerts', {
+			if (this._alertContainer) {
+				this._alertContainer[name] = {
 					alertType: type,
-					alertName: name,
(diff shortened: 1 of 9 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -379,7 +382,6 @@
 			}
 		},
 		_onStartedInactiveAlert: function(e) {
-			this._removeAlert('startedInactiveCourses');
```
njostonehouse: Was this an intentional change?

m6jones: Yes, since I am holding the alerts in an object. The name is the key so it get replaced instead of added. So removing is not required.

src/d2l-alert-behavior.html
```diff
@@ -14,34 +14,48 @@
...
-				return false;
-			}
-			return this._alerts.some(function(alert) {
-				return alert.alertName === name;
-			});
+			return !!this._alertContainer && !!this._alertContainer[name];
 		},
 		_removeAlert: function(name) {
-			if (this._alerts && this._alerts.length > 0 && this._hasAlert(name)) {
-				this.splice('_alerts', this._alerts.map(function(alert) { return alert.name; }).indexOf(name), 1);
```
njostonehouse: If we used `this._alerts.indexOf(this.get( ['_alerts', name] ))` to get the index instead of `this._alerts.map`, would it help? I wonder if using the prototype `map` method is circumventing Polymer's concurrency.

m6jones: It might be. The problem with using `this._alerts.indexOf(this.get( ['_alerts', name] ))` is `get` requires the path. However, I am going to look into seeing if I can figure out how to do it without `map`.

njostonehouse: Yeah, I wasn't sure what's supposed to go into the "key" parameter. The Polymer docs talk about an "array key," which I assume is the thing that normally goes into the `[...]`. Not sure if it requires indexing in a different way?

m6jones: yeah, it would look like this, `this._alerts.indexOf(this.get( ['_alerts', 1, 'alertName'] ))`. Going through the polymer api nothing popped out at me that would help.

PR #612, opened 2018-09-14
Hide completed and close courses in css

m6jones:
Use course count from org->enrollment map
Display no open courses alert based enrollments
Enrollment count.
Fixed double count.
Maxed out the course count to 99+
Fixed alerts being cleared
Fixed View All Courses link not showing when there is one hidden course

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -258,7 +258,7 @@
 
 			if (hide) {
 				var index = this._enrollments.indexOf(e.detail.enrollmentUrl);
-				if (index !== -1) {
+				if (index !== -1 && index >= this._lastPinnedIndex) {
 					this.splice('_enrollments', index, 1);
 				}
 			} else if (!hide && (this._enrollments.indexOf(e.detail.enrollmentUrl) === -1)) {
@@ -372,6 +372,9 @@
 				return this._refetchEnrollments();
 			}
 
+			var enrollmentCard = Polymer.dom(e).event && Polymer.dom(e).event.srcElement;
+			var shouldHide = enrollmentCard && (enrollmentCard.hasAttribute('completed') || enrollmentCard.hasAttribute('closed'));
+
 			var lastPinnedIndex = isPinned ? this._lastPinnedIndex++ : --this._lastPinnedIndex;
 			var removalIndex = this._enrollments.indexOf(changedEnrollmentId);
 
@@ -381,12 +384,17 @@
 			} else if (removalIndex === lastPinnedIndex) {
 				// Course is already in the correct position, just update it
 				// (avoids removal animation + insertion animation in exact same position)
+				if (shouldHide && !isPinned) {
+					this.splice('_enrollments', removalIndex, 1);
+				}
 			} else {
 				// Remove the enrollment from its current location
 				this.splice('_enrollments', removalIndex, 1);
 
 				// Re-add the (updated) enrollment at the end of the pinned list
-				this.splice('_enrollments', lastPinnedIndex, 0, changedEnrollmentId);
+				if (!shouldHide || isPinned) {
+					this.splice('_enrollments', lastPinnedIndex, 0, changedEnrollmentId);
+				}
 			}
 		},
 		_onStartedInactiveAlert: function(e) {
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -79,6 +79,10 @@
 				type: Object,
 				value: function() { return {}; }
 			},
+			_numberOfEnrollment: {
+				type: Number,
+				value: 0
+			},
 			_lastPinnedIndex: {
 				type: Number,
 				value: 0
@@ -101,7 +105,7 @@
 			// Text to render for "View All Courses" link (includes enrollment count approximation)
 			_viewAllCoursesText: {
 				type: String,
-				computed: '_getViewAllCoursesText(_hasMoreEnrollments, _enrollments.length)'
+				computed: '_getViewAllCoursesText(_hasMoreEnrollments, _numberOfEnrollment)'
 			}
 		},
 		listeners: {
@@ -189,8 +193,9 @@
 		_enrollmentsChanged: function(enrollments) {
 			this._hasEnrollments = enrollments.length > 0;
 
-			this._clearAlerts();
+			this._removeAlert('noCourses');
 			if (!this._hasEnrollments) {
+				this._clearAlerts();
 				this._addAlert('call-to-action', 'noCourses', this.localize('noCoursesMessage'));
 			}
 		},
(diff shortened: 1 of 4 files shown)


test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -846,14 +846,6 @@ describe('d2l-my-courses-content', () => {
 			});
 		});
 
-		it('should remove all existing alerts when enrollment alerts are updated', () => {
```
m6jones: Why did we want all alerts to removed when enrollments changed? When enrollments changing doesn't always update the visible enrollments which means the relevant alerts could be cleared. They were being cleared and so why I took it out and why I am taking out this test case.

Figure I'd explain why I am taking out a test case and open the discussion about it.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -79,6 +79,10 @@
 				type: Object,
 				value: function() { return {}; }
 			},
+			_numberOfEnrollment: {
```
m6jones: Since I am taking away enrollment cards from `_enrollments` in order to hide them. (closed and complete cards) Taking the length of `_enrollments` is no longer a valid way to get the number of enrollments.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -251,7 +258,7 @@
 
 			if (hide) {
 				var index = this._enrollments.indexOf(e.detail.enrollmentUrl);
-				if (index !== -1) {
+				if (index !== -1 && index >= this._lastPinnedIndex) {
```
njostonehouse: Does this depend on all pinned items appearing at the top of the list?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -189,8 +193,9 @@
 		_enrollmentsChanged: function(enrollments) {
 			this._hasEnrollments = enrollments.length > 0;
 
-			this._clearAlerts();
+			this._removeAlert('noCourses');
```
m6jones: Since the `_enrollments` change when something is hidden it no longer makes sense to clear all of the alerts.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -189,8 +193,9 @@
...
 			if (!this._hasEnrollments) {
+				this._clearAlerts();
```
m6jones: However, if there is no enrollments being shown then there shouldn't be any alerts so clear them now.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -237,7 +242,9 @@
 			}
 
 			var orgUnitId = this._getOrgUnitIdFromHref(e.detail.organizationUrl);
-
+			if (!this._orgUnitIdMap[orgUnitId]) {
+				this._numberOfEnrollment++;
```
m6jones: This event gets called to register an enrollment so it is the best place to keep track of how many enrollments are loaded. The `_orgUnitIdMap` makes sure we don't duplicate the number.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
m6jones: I don't want to remove/hide an enrollment if it is pinned.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -365,6 +372,9 @@
 				return this._refetchEnrollments();
 			}
 
+			var enrollmentCard = Polymer.dom(e).event && Polymer.dom(e).event.srcElement;
```
m6jones: I need this so I can check to see if this enrollment should be hidden. As in if it is completed or closed.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -554,10 +569,12 @@
 
 			// With individual fetching of courses as they get pinned, we can end
 			// up with "21+", "22+", etc., so round down to nearest 5 for >20 courses
+			var maxCount = 99;
```
m6jones: Business case wanted to max these out at 99.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
njostonehouse: nit: `_numberOfEnrollment` pluralized?

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -46,11 +46,13 @@
 				margin-bottom: 20px;
 				clear: both;
 			}
-			.course-tile-grid:not([hide-past-courses]) > div:nth-of-type(n+13) > d2l-enrollment-card:not([pinned]),
+			.course-tile-grid:not([hide-past-courses]) > d2l-enrollment-card:nth-of-type(n+13):not([pinned]),
 			.course-tile-grid:not([hide-past-courses]) > div:nth-of-type(n+13) > d2l-course-image-tile:not([pinned]),
-			.course-tile-grid[hide-past-courses] > div > d2l-enrollment-card[past-course]:not([pinned]),
+			.course-tile-grid[hide-past-courses] d2l-enrollment-card[past-course]:not([pinned]),
+			.course-tile-grid > d2l-enrollment-card[completed]:not([pinned]),
```
m6jones: So I removed the div container to fix issues with the shady dom and IE11. It would leave the enrollment cards in place so it be an empty box.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -554,10 +569,12 @@
...
 			var count = enrollmentsLength < 20
 				? enrollmentsLength
 				: String(enrollmentsLength - (enrollmentsLength % 5));
-			if (hasMoreEnrollments) count += '+';
+			if (count > maxCount) count = maxCount + '+';
+			if (hasMoreEnrollments && count !== maxCount + '+') count += '+';
```
njostonehouse: I'm not a fan of dropping `{`s in conditional statements.

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -73,19 +75,17 @@
 			</d2l-alert>
 
 			<template is="dom-repeat" items="[[_alertsView]]">
-				<d2l-alert type="[[item.alertType]]">
+				<d2l-alert type="[[item.alertType]]" hidden$="[[!_enrollments.length]]">
```
m6jones: If the visible amount of enrollment is 0 then don't show alerts.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
m6jones: Yes, which they do in the my widget view and that is where it is applied.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
m6jones: Ohh! bad habit from last job.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
njostonehouse: If `this._hasEnrollments` is `false`, does that necessarily mean the View All Courses page is empty too?

PR #613, opened 2018-09-17
Fixed long single word names from stretching the card.

m6jones:
-- IE11 will hide closed and completed courses correctly.

### DIFF
--- a/src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
+++ b/src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
@@ -62,7 +62,9 @@
 				return;
 			}
 
-			var courseTileDivs = Polymer.dom(this.root).querySelectorAll('.course-tile-grid > div');
+			var courseTileDivs = this.useEnrollmentCards
+				? Polymer.dom(this.root).querySelectorAll('.course-tile-grid > d2l-enrollment-card')
+				: Polymer.dom(this.root).querySelectorAll('.course-tile-grid > div');
 			ie11retryCount = ie11retryCount || 0;
 			if (
 				ie11retryCount < 20
@@ -74,11 +76,16 @@
 				return;
 			}
 
-			for (var i = 0; i < courseTileDivs.length; i++) {
+			for (var i = 0, position = 0; i < courseTileDivs.length; i++, position++) {
 				var div = courseTileDivs[i];
+				if ((div.hasAttribute('completed') || div.hasAttribute('closed')) && !div.hasAttribute('pinned')) {
+					position--;
+					continue;
+				}
+
 				// The (* 2 - 1) accounts for the grid-gap-esque columns
-				var column = (i % numColumns + 1) * 2 - 1;
-				var row = Math.floor(i / numColumns) + 1;
+				var column = (position % numColumns + 1) * 2 - 1;
+				var row = Math.floor(position / numColumns) + 1;
 				div.style['-ms-grid-column'] = column;
 				div.style['-ms-grid-row'] = row;
 			}
--- a/src/d2l-css-grid-view/d2l-css-grid-view-styles.html
+++ b/src/d2l-css-grid-view/d2l-css-grid-view-styles.html
@@ -27,7 +27,8 @@
 				-ms-grid-columns: 1fr 15px 1fr 15px 1fr 15px 1fr;
 			}
 
-			.course-tile-grid > div {
+			.course-tile-grid > div,
+			.course-tile-grid > d2l-enrollment-card {
 				min-width: 0;
 			}
 
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -79,7 +79,7 @@
 				type: Object,
 				value: function() { return {}; }
 			},
-			_numberOfEnrollment: {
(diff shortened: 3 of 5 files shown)


src/d2l-css-grid-view/d2l-css-grid-view-styles.html
```diff
@@ -27,7 +27,8 @@
 				-ms-grid-columns: 1fr 15px 1fr 15px 1fr 15px 1fr;
 			}
 
-			.course-tile-grid > div {
+			.course-tile-grid > div,
+			.course-tile-grid > d2l-enrollment-card {
 				min-width: 0;
```
m6jones: The cause of the stretched cards. Obviously, `min-width: 0` was missing. 🤦‍♀️

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -79,7 +79,7 @@
 				type: Object,
 				value: function() { return {}; }
 			},
-			_numberOfEnrollment: {
+			_numberOfEnrollments: {
```
m6jones: Added an 's'. Name makes more sense.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -266,6 +266,7 @@
 					this.push('_enrollments', e.detail.enrollmentUrl);
 				}
 			}
+			this._onResize();
```
m6jones: This is required because the rows and columns need to be recalculated for IE11.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -396,6 +397,7 @@
 					this.splice('_enrollments', lastPinnedIndex, 0, changedEnrollmentId);
 				}
 			}
+			this._onResize();
```
m6jones: Unpinning a closed card makes it hide :)

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -889,19 +889,19 @@ describe('d2l-my-courses-content', () => {
 
 		it('should show the number of enrollments when there are no new pages of enrollments with the View All Courses link', () => {
 			component._hasMoreEnrollments = false;
-			component._numberOfEnrollment = 6;
+			component._numberOfEnrollments = 6;
```
m6jones: just the 's'

src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
```diff
@@ -74,11 +76,16 @@
 				return;
 			}
 
-			for (var i = 0; i < courseTileDivs.length; i++) {
+			for (var i = 0, position = 0; i < courseTileDivs.length; i++, position++) {
```
njostonehouse: This is a little surprising? Were we calculating rows incorrectly?

m6jones: This is to keep track when I skip an element because it is `closed` or `completed`

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -82,10 +82,10 @@
 			<div class="course-tile-grid">
 				<template is="dom-if" if="[[useEnrollmentCards]]">
 					<template is="dom-repeat" items="[[_enrollments]]">
-							<d2l-enrollment-card
-								href="[[item]]"
-								presentation-href="[[presentationUrl]]">
-							</d2l-enrollment-card>
+						<d2l-enrollment-card
```
ryantmer: Huh, so another way to solve this issue - wrap the `d2l-enrollment-card` in a div.

There seems to be something strange about how IE11 deals with WCs inside a DOM-repeat - this is why the `d2l-course-image-tile` was wrapped in a div itself. I can't 100% remember the problem that arose without that div, unfortunately, but it was a necessary evil for the unnecessary evil that is IE11, and it looks like it would actually solve this DE as well, since it was "missing" the `min-width: 0` being applied to it.

ryantmer: Aha, here - https://github.com/Brightspace/d2l-my-courses-ui/pull/489 more details there!

m6jones: It is actually what caused it :( Me removing it.

So the div make it hard for IE11 to hide the `completed` and `closed` courses. So thats why I removed it in the first place. IE11 sooo picky.

ryantmer: Ah, gotcha! I was thinking I did remember it being there at some point. Shouldn't it just require the extra `> div` in the CSS selector though?

m6jones: Doing it that way doesn't free of the grid spot. So you get an empty grid item.

ryantmer: Huh - works with `d2l-course-image-tile` though? Not sure what the difference would be though, IE11's handling is very wonky :(

m6jones: I don't know how the `d2l-course-image-tile`  handled hiding the `closed` courses.

PR #614, opened 2018-09-17
No course alert shows when there are no courses.

m6jones:
The 'You don't have any courses to display...' alert is not being shown in the My Courses Widget when a user logs in and isn't enrolled in any courses.

Steps to reproduce:
1) Create a new user
2) Log in as that user
3) Observe the My Courses Widget

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -191,6 +191,7 @@
 		_enrollmentsSearchUrl: null,
 
 		_enrollmentsChanged: function(viewAbleLength, totalLength) {
+			totalLength = this.useEnrollmentCards ? totalLength : viewAbleLength;
 			this._hasEnrollments = totalLength > 0;
 
 			this._removeAlert('noCourses');
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -143,7 +143,7 @@
 			document.body.removeEventListener('d2l-tab-panel-selected', this._onTabSelected.bind(this));
 		},
 		observers: [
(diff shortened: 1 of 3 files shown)


test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -856,10 +859,11 @@ describe('d2l-my-courses-content', () => {
...
 		});
 
-		it('should correctly evaluate whether it has enrollments', () => {
-			return component._fetchRoot().then(() => {
+		it('should correctly evaluate whether it has enrollments', done => {
+			setTimeout(() => {
 				expect(component._hasEnrollments).to.be.true;
-			});
+				done();
+			}, 2000);
```
m6jones: This is super long because `_hasEnrollments` get set after a longer chain of events.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -143,7 +143,7 @@
 			document.body.removeEventListener('d2l-tab-panel-selected', this._onTabSelected.bind(this));
 		},
 		observers: [
-			'_enrollmentsChanged(_enrollments)',
+			'_enrollmentsChanged(_enrollments.length, _numberOfEnrollments)',
```
ryantmer: Shouldn't these be the same value? 😮

m6jones: So `_enrollments` hold the current enrollments that can be shown. `_numberOfEnrollments` holds the number of enrollments that we currently have loaded and that will show up in the all courses view.

ryantmer: Riiight, I see. Yeah, that is correct, they don't have the concept of complete/closed, they're just doing hiding based on "past" and such (that horrible enormous CSS selector, you know the one 😜)

PR #615, opened 2018-09-19
Setup all courses view the same as the widget.

m6jones:
While enrolled in multiple, open/active courses, going to View All Courses shows only one card.

Steps to reproduce:
Use IE11
Login as a user that is enrolled in more than one course
Click on View All Courses
Observe
Actual:
Only one course card is rendered

Expected:
All enrolled courses should be rendered

### DIFF
--- a/src/d2l-all-courses-unified-content.html
+++ b/src/d2l-all-courses-unified-content.html
@@ -39,12 +39,10 @@ This is only used if the `US90527-my-courses-updates` LD flag is ON
 		<div class="course-tile-grid">
 			<template is="dom-if" if="[[useEnrollmentCards]]">
 				<template is="dom-repeat" items="[[filteredEnrollments]]">
-					<div>
-						<d2l-enrollment-card
-							href="[[item]]"
-							presentation-href=[[presentationUrl]]>
-						</d2l-enrollment-card>
-					</div>
+					<d2l-enrollment-card
+						href="[[item]]"
+						presentation-href=[[presentationUrl]]>
+					</d2l-enrollment-card>
 				</template>
 			</template>
 			<template is="dom-if" if="[[!useEnrollmentCards]]">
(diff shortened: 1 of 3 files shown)


src/d2l-all-courses-unified-content.html
```diff
@@ -39,12 +39,10 @@
 		<div class="course-tile-grid">
 			<template is="dom-if" if="[[useEnrollmentCards]]">
 				<template is="dom-repeat" items="[[filteredEnrollments]]">
-					<div>
```
m6jones: Style files that use this are:
- `<style include="d2l-all-courses-styles"></style>`
- `<style include="d2l-css-grid-view-styles"></style>`
Removing this div has no impact on them.

src/d2l-css-grid-view/d2l-css-grid-view-behavior.html
```diff
@@ -89,6 +89,12 @@
 				div.style['-ms-grid-column'] = column;
 				div.style['-ms-grid-row'] = row;
 			}
+		},
+
+		__hideElement: function(div) {
+			return this.is !== 'd2l-all-courses-unified-content'
```
m6jones: `d2l-all-courses-unified-content` is used in `d2l-all-courses` view. So checking to see if the current component is this basically checks to see if we are in `d2l-all-courses`
It is the only way I could find that would allow me to not hide in the all courses view.

njostonehouse: Does this still work with `d2l-all-courses-segregated-content`?

njostonehouse: It's incredibly likely that we don't use `-segregated-content` behind our flags so it might not matter.

m6jones: It looks like we use it `<template is="dom-if" if="[[!updatedSortLogic]]">` It looks like when `updatedSortLogic` flag is not enabled, we never use `enrollment-card`. So it looks like we are good.

`d2l-all-courses-segregated-content` doesn't use `d2l-enrollment-card`. The similar view in the widget is `d2l-my-courses-content-animated` also doesn't use `d2l-enrollment-card`.
Soo all good.

PR #616, opened 2018-09-24
Updated so the Alert shows when you only have closed courses.

m6jones:
image


m6jones:
Thought of a case where this doesn't work as intended.

m6jones:
I tested this case and it works. I was worried when a closed course isn't hidden and the alert would still show up. It doesn't show up in this case. So perfect.

m6jones:
That is correct.

PR #618, opened 2018-09-27
DE31508 - Handle hiding completed / closed courses

m6jones:
- Handle the case when the user has > 50 courses and <= 50 courses properly.
- Only using CSS to hide past (closed) courses. The backend puts these at the end of the api calls.
- Using a more complicated method to hide completed courses as the backend doesn't sort these. Completed courses won't hide when there are > 50 courses.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -65,9 +65,9 @@
 				value: false
 			},
 			// True when user has >1 page of enrollments
-			_hasMoreEnrollments: {
-				type: Boolean,
-				value: false
+			_nextEnrollmentEntityUrl: {
+				type: String,
+				value: null
 			},
 			_hasOnlyPastCourses: {
 				type: Boolean,
@@ -105,7 +105,7 @@
 			// Text to render for "View All Courses" link (includes enrollment count approximation)
 			_viewAllCoursesText: {
 				type: String,
-				computed: '_getViewAllCoursesText(_hasMoreEnrollments, _numberOfEnrollments)'
+				computed: '_getViewAllCoursesText(_nextEnrollmentEntityUrl, _numberOfEnrollments)'
 			}
 		},
 		listeners: {
@@ -190,7 +190,6 @@
 		_initiallyVisibleCourseTileCount: 0,
 		_enrollmentsSearchUrl: null,
 		_widgetMaxCardVisible: 12,
-		_nextEnrollmentEntityUrl: false,
 		_hidePastCourses: false,
 
 		_enrollmentsChanged: function(viewAbleLength, totalLength) {
@@ -627,8 +626,8 @@
 		},
 		_populateEnrollments: function(enrollmentsEntity) {
 			var enrollmentEntities = enrollmentsEntity.getSubEntitiesByClass('enrollment');
-			this._hasMoreEnrollments = enrollmentsEntity.hasLinkByRel('next');
-			this._nextEnrollmentEntityUrl = this._hasMoreEnrollments && enrollmentsEntity.getLinkByRel('next').href;
+			var hasMoreEnrollments = enrollmentsEntity.hasLinkByRel('next');
+			this._nextEnrollmentEntityUrl = hasMoreEnrollments ? enrollmentsEntity.getLinkByRel('next').href : null;
 			var newEnrollments = [];
 
 			var searchAction = enrollmentsEntity.getActionByName(this.HypermediaActions.enrollments.searchMyEnrollments);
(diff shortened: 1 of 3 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -189,6 +189,9 @@
 		_courseImagesLoadedEventCount: 0,
 		_initiallyVisibleCourseTileCount: 0,
 		_enrollmentsSearchUrl: null,
+		_widgetMaxCardVisible: 12,
+		_nextEnrollmentEntityUrl: false,
```
m6jones: I wasn't sure if this should replace the `_hasMoreEnrollments` property. In a sense if this is not falsy then there are more enrollments.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -251,13 +254,17 @@
 			this._orgUnitIdMap[orgUnitId] = e.detail.enrollmentUrl;
 		},
 		_onD2lEnrollmentCardStatus: function(e) {
+			// When we need to hide closed, the backend puts it at the back of the list. So nothing to do.
 			if (!e.detail
 				|| !e.detail.status
 				|| !e.detail.enrollmentUrl
+				|| !this._hidePastCourses
+				|| e.detail.status.closed
```
m6jones: Enrollment card still fires this event when the status is closed. However, d2l-my-courses doesn't care about it since the backend does the work. 

I added a comment in the code to explain this since it did cause some extra work that wasn't required. Hopefully the comment stops that.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -251,13 +254,17 @@
...
 			) {
 				return;
 			}
-			var hide = (e.detail.status.completed || e.detail.status.closed);
+
+			var hide = e.detail.status.completed;
```
m6jones: This was added when I was hiding both completed and closed. When I went to take it out, I decided it still added readability so I decided to keep it.

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -46,14 +46,14 @@
...
-			.course-tile-grid[hide-past-courses] d2l-enrollment-card[past-course]:not([pinned]),
-			.course-tile-grid > d2l-enrollment-card[completed]:not([pinned]),
-			.course-tile-grid > d2l-enrollment-card[closed]:not([pinned]),
 			.course-tile-grid[hide-past-courses] > div > d2l-course-image-tile[past-course]:not([pinned]),
-			.course-tile-grid[hide-past-courses] > d2l-enrollment-card:nth-of-type(n+13):not([pinned]):not([past-course]):not([completed]),
-			.course-tile-grid[hide-past-courses] > div:nth-of-type(n+13) > d2l-course-image-tile:not([pinned]):not([past-course]):not([completed]) {
+			.course-tile-grid[hide-past-courses] > d2l-enrollment-card:nth-of-type(n+13):not([pinned]),
+			.course-tile-grid:not([hide-past-courses]) > d2l-enrollment-card:nth-of-type(n+13):not([pinned]),
+			.course-tile-grid[hide-past-courses] > d2l-enrollment-card[completed]:not([pinned]),
+			.course-tile-grid[hide-past-courses] > d2l-enrollment-card[closed]:not([pinned]), {
```
m6jones: So I reorganized the CSS here. I wanted to organize it by card and tile.

I also removed some CSS that doesn't do anything. I didn't touch tile versions.

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -865,7 +865,7 @@ describe('d2l-my-courses-content', () => {
 			setTimeout(() => {
 				expect(component._hasEnrollments).to.be.true;
 				done();
-			}, 2000);
+			}, 3000);
```
m6jones: Safari didn't like 2 seconds. So I added a second.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
njostonehouse: Thanks. The API to the Enrollments needs to be updated so that the "demotion"  of closed Enrollments is explicit rather than obfuscated.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
njostonehouse: I think it would be a good idea to coalesce those two properties, particularly if they're used interchangeably. The only situation I can imagine where they are _not_ interchangeable would be if we "have more Enrollments" but there is no valid "next" link. I'd also want to be cautious about when and how we reset this to something falsy, once we get to the end of the pages.

PR #627, opened 2018-10-26
US101469: Remove Enrollent Cards conditions. Always use Enrollment Ca…

njostonehouse:
…rds (if improved sort logic is enabled).


ryantmer:
In celebration, https://github.com/BrightspaceHypermediaComponents/1.x-hybrid-components is now archived!

PR #629, opened 2018-10-31
US101469: Try to use less sauce when PRing.

njostonehouse:
An experiment in fiddling with Travis and Sauce to try and make PR CI less likely to flake out. Not sure what will happen.


.travis.yml
```diff
@@ -10,13 +10,10 @@ before_script:
 script:
 - cp test/index.high-priority.html test/index.html
 - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
-    cp wct.conf.sauce-chrome.json wct.conf.json;
```
njostonehouse: The hope here is that the `wct.conf.local.json` gives us some confidence that Chrome is doing ok.

m6jones:
No, got the `elementbyId` error still. Restarted it.

njostonehouse:
Looks like it's still falling over.

PR #630, opened 2018-11-02
US101469: Stop testing Firefox on Sauce, in case automated test flake…

njostonehouse:
…s are restricted to that browser.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ script:
     cp wct.conf.sauce-chrome.json wct.conf.json;
     polymer test || travis_terminate 1;
 
-    cp wct.conf.sauce-firefox.json wct.conf.json;
-    polymer test || travis_terminate 1;
-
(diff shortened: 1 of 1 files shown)


njostonehouse:
This still seems flaky, although other bits are failing. Although Firefox seems to be especially bad at failing, the other browsers seem to get tripped up from time to time too (in new and exciting ways!)

m6jones:
I think the tests need to be redesigned and the wall of text might actually be a problem with `d2l-tabs` or the way we use it. Dragon Slayer that isn't taken over by My Courses.

njostonehouse:
I'm trying a few different things as experiments. Here's the next one: https://github.com/Brightspace/d2l-my-courses-ui/pull/631 (it didn't work).

svanherk:
Closing, the original issue with that Firefox-only flake was solved by moving to Polymer 3

PR #631, opened 2018-11-02
US101469: Wrap all tests in a describe.

njostonehouse:
Trying to reduce flakiness in sauce tests.


njostonehouse:
Still failed with the `document.getElementById(...).create` error.

PR #632, opened 2018-11-05
US101469: Move filter list tests into a describe block.

njostonehouse:


njostonehouse:
I don't think this makes a difference to test stability, but I also don't think it's harmful to have these tests wrapped in a `describe` like the others.

njostonehouse:
`Cannot set property 'selected' of null`
Apparently I messed something up.

njostonehouse:
Merging anyway - failures are related to CI flake that we're trying to address.

PR #638, opened 2018-11-19
DE31968: Only listen for tab updates on the active tab.

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -414,8 +414,13 @@
 		_onTabSelected: function(e) {
 			// Only handle if tab selected corresponds to this panel
 			if (!this.parentElement || e.target.id !== this.parentElement.id) {
+				document.body.removeEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
```
njostonehouse: By default, every tab gets the listener. This avoids regressions in pinning status when there are no tabs used (the only panel continues to get all the events).

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -414,8 +414,13 @@
...
 				return;
 			}
+
+			// Only listen to pin updates for the current tab
+			document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
```
njostonehouse: This is redundant when the current tab is initially selected, but doesn't seem harmful.

njostonehouse:
CI failed with Firefox in Sauce; a common occurrence we've been ignoring lately.

PR #639, opened 2018-11-22
DE30581-fix the initial sort order to have pinned courses first

edwardg8:


ryantmer:
I'm kind of out of the loop on this stuff now, but I was under the impression that we wanted the order in All Courses to just be alphabetical by default, without promoting pins. Is that no longer the case? (Totally fine if so, just wanted to call out that this was by design at some point.)

edwardg8:
I'm not too sure about the initial design, just picking up some defects in the backlog and following its description. We can wait for @njostonehouse to clarify things for us, hopefully he knows better.

njostonehouse:
This is interesting. I also expected that the "Default" sort order wouldn't have promoted pins on the overlay. There is definitely a mismatch here, since re-selecting Default seems to promote pins later, so we need some consistency. Like @ryantmer I was pretty sure we wanted to avoid promoting pins on the overlay for users with >50 Enrollments for technical (rather than design) reasons.

njostonehouse:
Ultimately I think this is probably ok, but let's do an audit of the performance/load implications on the backend of making this change. My gut feeling is that if you're promoting pins to the top on the Widget there wouldn't be any _technical_ drawback to also doing it on the overlay. But something is bugging me about this and I can't put my finger on it.

src/d2l-all-courses.html
```diff
@@ -360,7 +360,8 @@
 				this._searchUrl = this.createActionUrl(this.enrollmentsSearchAction, {
 					autoPinCourses: false,
 					embedDepth: this.updatedSortLogic ? 0 : 1,
-					sort: this._sortParameter || (this.updatedSortLogic ? 'Current' : '-PinDate,OrgUnitName,OrgUnitId')
+					sort: this._sortParameter || (this.updatedSortLogic ? 'Current' : '-PinDate,OrgUnitName,OrgUnitId'),
+					promotePins: true
```
njostonehouse: I think the reasoning in leaving this _out_ was that if a user goes to the View All Courses page, their pins weren't doing the job in finding the course they wanted in the Widget. From a design standpoint, alphabetical (or some other sort order) makes more sense to me.

edwardg8:
After talking to Jeff and Sandra, we decided to just close the defect. We'll reopen this defect if we hear our clients complaining about it, for now we don't know what is the best decision here.

PR #641, opened 2018-11-27
US101520: Refactor direct references to d2l-fetch into the utility fu…

njostonehouse:
…nction to prepare for using a d2l-fetch wrapper instead.

### DIFF
--- a/src/d2l-utility-behavior.html
+++ b/src/d2l-utility-behavior.html
@@ -49,13 +49,22 @@
 		parseEntity: function(entity) {
 			return window.D2L.Hypermedia.Siren.Parse(entity);
 		},
-		fetchSirenEntity: function(url) {
+		fetchSirenEntity: function(url, clearCache) {
 			if (!url) {
 				return;
 			}
+
+			var headers = {
+				Accept: 'application/vnd.siren+json'
+			};
+
+			if( clearCache ) {
+				headers['cache-control'] = 'no-cache';
+			}
+
 			return window.d2lfetch
 				.fetch(new Request(url, {
-					headers: { Accept: 'application/vnd.siren+json' }
+					headers: headers
 				}))
 				.then(this.responseToSirenEntity.bind(this));
 		},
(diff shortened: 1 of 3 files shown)


njostonehouse:
polymer-siren-behaviors https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-store.html will include additional fetching behaviour that extends on the existing d2l-fetch. My intention here is to keep the code that uses d2l-fetch in one spot so that the new fetching can be replaced in one spot.

src/d2l-utility-behavior.html
```diff
@@ -79,6 +88,20 @@
 				}
 			}
 		},
+		submitForm: function(formParameters) {
+			var formData = new FormData();
+			for ( var formKey in formParameters ) {
+				if( formParameters.hasOwnProperty(formKey) ) {
```
m6jones: nit (lint): space after if

m6jones: And remove spaces in the brackets.

njostonehouse: Argh! Right, thank you.

njostonehouse:
Darn it, I should've added tests. Stay tuned.

m6jones:
I'd assume the current tests would still cover these changes. You only did a refactor?

njostonehouse:
I refactored, but I made two new functions. Looking again, it probably isn't super valuable to add coverage there.

src/d2l-utility-behavior.html
```diff
@@ -79,6 +88,20 @@
...
+		submitForm: function(formParameters) {
+			var formData = new FormData();
+			for (var formKey in formParameters) {
+				if (formParameters.hasOwnProperty(formKey)) {
+					formData.append(formKey, formParameters[formKey]);
+				}
+			}
+
+			return window.d2lfetch
+				.fetch(new Request(this.userSettingsUrl, {
```
ryantmer: I think this URL should be passed in!

njostonehouse: oh gosh
yes

src/d2l-utility-behavior.html
```diff
@@ -49,13 +49,22 @@
 		parseEntity: function(entity) {
 			return window.D2L.Hypermedia.Siren.Parse(entity);
 		},
-		fetchSirenEntity: function(url) {
+		fetchSirenEntity: function(url, clearCache) {
```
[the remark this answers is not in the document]
njostonehouse: You're right, it seems to be missing. Not sure what happened there.

njostonehouse:
Dang it I forgot to lint again.

PR #643, opened 2018-12-18
US102322: Initial commit of a my courses widget demo page.

njostonehouse:


njostonehouse:
This is just a preliminary demo of My Courses, in tab mode. There are a bunch more test cases to cover still, but this one at least exposes a shadow DOM bug, so worth getting some foundation set.

demo/data/organization-01.json
```diff
@@ -0,0 +1,6 @@
+{
+  "properties": {
+    "name": "Course 01",
+    "isActive": true
```
njostonehouse: My aim with these .json files is to present a minimal view of dependent components (enrollments and organizations). In this case, the base case requires `isActive` to be `true` explicitly or we get a bunch of "inactive" course cards.

PR #644, opened 2018-12-18
DE32267: Catch a dispatched event with proper Polymer syntax (breaks …

njostonehouse:
…in Shadow DOM otherwise).


njostonehouse:
Tests need to be updated to accommodate this change.

PR #646, opened 2018-12-20
US102322: Create a demo and test data for alerts.

njostonehouse:


m6jones:
The title says test data. Did you mean demo data?

njostonehouse:
@m6jones Oh, yeah, good point. Dummy data for the demo, not for tests.

PR #651, opened 2019-01-09
DE32449: Initialize Last Pinned Index to -1 (no pinned items). Determ…

njostonehouse:
…ine insert and removal indices on pin event more clearly.

### DIFF
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -385,6 +385,14 @@
 			var removalIndex = this._enrollments.indexOf(changedEnrollmentId);
 			var insertIndex = this._lastPinnedIndex + 1;
 
+			if (!isPinned) {
+				this._lastPinnedIndex--;
+			}
+
+			if (isPinned) {
+				this._lastPinnedIndex++;
+			}
+
 			if (removalIndex === insertIndex && !shouldHide) {
 				this._onResize();
 				return;
@@ -392,10 +400,6 @@
 
 			if (removalIndex !== -1) {
 				this.splice('_enrollments', removalIndex, 1);
-
-				if (!isPinned) {
-					this._lastPinnedIndex--;
-				}
 
 				if (shouldHide) {
 					this._onResize();
@@ -408,10 +412,6 @@
 			}
 
 			this.splice('_enrollments', insertIndex, 0, changedEnrollmentId);
-
-			if (isPinned) {
-				this._lastPinnedIndex++;
-			}
 
 			this._onResize();
 		},
--- a/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
+++ b/src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
@@ -84,7 +84,7 @@
 			},
 			_lastPinnedIndex: {
 				type: Number,
-				value: 0
+				value: -1
 			},
 			// The organization which the user is selecting the image of
 			_setImageOrg: {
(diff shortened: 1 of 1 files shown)


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -84,7 +84,7 @@
 			},
 			_lastPinnedIndex: {
 				type: Number,
-				value: 0
+				value: -1
```
njostonehouse: By coincidence rather than by design, initializing this value as 0 would seem to indicate that we always assume there's at least one pin (the "last pinned index" is the beginning of the array). This should be -1 if there are no pins.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -380,29 +380,37 @@
 			}
 
 			var enrollmentCard = Polymer.dom(e).event && Polymer.dom(e).event.srcElement;
-			var shouldHide = enrollmentCard && (enrollmentCard.hasAttribute('completed') || enrollmentCard.hasAttribute('closed'));
+			var shouldHide = enrollmentCard && !isPinned && (enrollmentCard.hasAttribute('completed') || (enrollmentCard.hasAttribute('closed')));
```
njostonehouse: This still isn't quite right, but is more correct to include the "unpinned" event status.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -380,29 +380,37 @@
...
 
-			var lastPinnedIndex = isPinned ? this._lastPinnedIndex++ : --this._lastPinnedIndex;
 			var removalIndex = this._enrollments.indexOf(changedEnrollmentId);
+			var insertIndex = this._lastPinnedIndex + 1;
 
-			if (removalIndex === -1) {
-				// Course was not already in list, insert it as last pinned item
-				this.splice('_enrollments', lastPinnedIndex, 0, changedEnrollmentId);
```
njostonehouse: I think I missed this case, will update.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -380,29 +380,37 @@
...
-				if (!shouldHide || isPinned) {
-					this.splice('_enrollments', lastPinnedIndex, 0, changedEnrollmentId);
-				}
+			this.splice('_enrollments', removalIndex, 1);
+
+			if (!isPinned) {
+				this._lastPinnedIndex--;
+			}
+
+			if (shouldHide) {
```
njostonehouse: If this Enrollment is supposed to be hidden, don't add it to the list, just leave the function.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -380,29 +380,37 @@
...
+				this._onResize();
+				return;
+			}
+
+			if (removalIndex < insertIndex) {
+				insertIndex--;
```
njostonehouse: If the earlier removal would shift where this item is supposed to be inserted, shift the index back one to fill in the gap.

njostonehouse:
I think this change could use some test cases, but manual testing looks good.

edwardg8:
My local changes will have conflict with this PR. I'll get them resolved easily as long as this can be merged early.

njostonehouse:
Found a bug in manual testing. Fix coming soon.
Automated tests have proven to be a challenge. Will have to add them in later.

PR #652, opened 2019-01-09
Nstonehouse/de32377/hide past courses

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -201,22 +201,10 @@
...
-			pinnedEnrollment = this.$$('.course-tile-grid d2l-enrollment-card[pinned]');
-
-			return this._numberOfEnrollments !== 0
-				&& shouldHidePastCourses
-				&& !currentOrFutureEnrollment
-				&& !pinnedEnrollment;
+			return this.updatedSortLogic
+				&& this._hidePastCourses
+				&& this._numberOfEnrollments !== 0
+				&& this._enrollments.length === 0;
```
njostonehouse: Since the number of shown Enrollment (`_enrollments`) and the number of total enrollments (`_numberOfEnrollments`) and `_hidePastCourses` are determined elsewhere, this function can be greatly simplified.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -201,22 +201,10 @@
...
-				// d2l-my-courses-content-animated should never show this alert, as it doesn't use current/future courses
-				return false;
-			}
-
-			var shouldHidePastCourses = this.$$('.course-tile-grid').hasAttribute('hide-past-courses');
-
-			var currentOrFutureEnrollment;
-			var pinnedEnrollment;
-			currentOrFutureEnrollment = this.$$('.course-tile-grid d2l-enrollment-card:not([closed])');
-			pinnedEnrollment = this.$$('.course-tile-grid d2l-enrollment-card[pinned]');
```
njostonehouse: This css querying won't work in ShadowDOM anymore anyway.

src/d2l-my-courses-content/d2l-my-courses-content.html
```diff
@@ -40,7 +40,6 @@
 				align-items: center;
 			}
 			d2l-alert {
-				display: block;
```
njostonehouse: This was getting in the way of setting `display: none` to hide alerts.

m6jones: How were you hiding the alerts? I feel like if you are hiding the alerts they should just be removed from the alert list.

njostonehouse: It works in Shady DOM, for some reason.

njostonehouse: Oh, right. One sec, I can find an answer to your actual question.

njostonehouse: The "Only Past Courses" alert is Special: https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content/d2l-my-courses-content.html#L68

m6jones: Oh, another way to make the hidden attribute work is to do,
`d2l-alert([hidden]) { display: none; }` 
Unless the block is completely useless this might be a better way to go about it?

njostonehouse: This technique allows `d2l-alert` to do what it wants with the `hidden` attribute, instead of overriding the behaviour of its internals. Currently `d2l-alert` configures its (non-hidden) layout as `flex`, which seems reasonable not to try and override.

njostonehouse: Not that I saw. The demo looks good, and local testing didn't have a difference that was apparent to me.

PR #653, opened 2019-01-11
Egao/refresh pinned tab

edwardg8:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -461,6 +491,17 @@
...
 		},
+		_onCourseDataChange: function(newValue) {
+			if (newValue) {
+				var isAllTab = this.enrollmentsSearchAction.name === this.HypermediaActions.enrollments.searchMyEnrollments;
+				var isPinnedTab = this.enrollmentsSearchAction.name === this.HypermediaActions.enrollments.searchMyPinnedEnrollments;
+				if (isAllTab || isPinnedTab) {
+					this._isRefetchNeeded = true;
+				} else if (newValue.orgUnitId && this._orgUnitIdMap[newValue.orgUnitId]) {
+					this._isRefetchNeeded = true;
+				}
```
edwardg8: only all tab, pinned tab an the semester/department that contains the course need to refetch when pin/unpin action happened

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -346,11 +353,18 @@
 				return this._refetchEnrollments();
 			}
 
+			this.fire('d2l-course-data-change', {
+				orgUnitId: orgUnitId,
+				isPinned: isPinned
+			});
+			this._isRefetchNeeded = false;
```
edwardg8: we don't have to refetch current tab either

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -378,7 +392,11 @@
 				}
 			}
 
-			this.splice('_enrollments', insertIndex, 0, changedEnrollmentId);
+			if (isPinnedTab) {
```
njostonehouse: If `shouldHide` is `true` then we won't get here. The item will be removed, but `_numberOfEnrollments` won't be reduced.

edwardg8: oh, right, will change this

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -402,6 +420,18 @@
 					.then(function() {
 						window.dispatchEvent(new Event('resize'));
 					});
+			} else if (this._isRefetchNeeded) {
+				this._showContent = false;
+
+				var completeFetch = function() {
+					this._showContent = true;
+					this._isRefetchNeeded = false;
```
njostonehouse: Shouldn't this be set to `true` immediately to keep extra refreshes from happening?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -461,6 +491,17 @@
 		_onPresentationUrlChange: function(newValue) {
 			this.fetchSirenEntity(newValue, true);
 		},
+		_onCourseDataChange: function(newValue) {
+			if (newValue) {
+				var isAllTab = this.enrollmentsSearchAction.name === this.HypermediaActions.enrollments.searchMyEnrollments;
+				var isPinnedTab = this.enrollmentsSearchAction.name === this.HypermediaActions.enrollments.searchMyPinnedEnrollments;
+				if (isAllTab || isPinnedTab) {
```
njostonehouse: Could this be one condition, since `_isRefetchNeeded` is set to `true` either way?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -461,6 +491,17 @@
 		_onPresentationUrlChange: function(newValue) {
 			this.fetchSirenEntity(newValue, true);
 		},
+		_onCourseDataChange: function(newValue) {
+			if (newValue) {
+				var isAllTab = this.enrollmentsSearchAction.name === this.HypermediaActions.enrollments.searchMyEnrollments;
```
njostonehouse: Maybe these could be turned into computed properties, instead of needing to check them all over.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
edwardg8: good point, but I don't think it makes huge different, unless the user keeps clicking this tab, but i'll change this.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -461,6 +491,17 @@
 		_onPresentationUrlChange: function(newValue) {
 			this.fetchSirenEntity(newValue, true);
 		},
+		_onCourseDataChange: function(newValue) {
+			if (newValue) {
```
njostonehouse: Is this likely to be `falsy`?
It might be a little more legible if you just return if it is `falsy`.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -610,6 +651,9 @@
 				this._hidePastCourses = true;
 			}
 
+			if (this._isRefetchNeeded) {
+				this._resetEnrollments();
```
njostonehouse: This might be better done during the refetch, instead of before.

njostonehouse: Oh, this might not be where the refetch happens.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
edwardg8: I'm not sure I understand you here, what one condition?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
edwardg8: This is during the fetch, not before?

edwardg8:
Weird, I don't see my new pushed changes here.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -461,7 +499,23 @@
...
 			this.fetchSirenEntity(newValue, true);
 		},
+		_onCourseDataChange: function(newValue) {
+			if (!newValue) {
+				return;
+			}
 
+			if (this._isAllTab || this._isPinnedTab) {
+				this._isRefetchNeeded = true;
+			} else if (newValue.orgUnitId && this._orgUnitIdMap[newValue.orgUnitId]) {
```
njostonehouse: I just meant instead of `else if` you could add this as an `||` to the previous condition.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -100,6 +101,19 @@
...
 				computed: '_getViewAllCoursesText(_nextEnrollmentEntityUrl, _numberOfEnrollments)'
+			},
+			// Whether or not to refetch the courses data
+			_isRefetchNeeded: {
+				type: Boolean,
+				value: false
+			},
+			_isAllTab: {
+				type: Boolean,
+				computed: '_checkAllTab(enrollmentsSearchAction.name)'
```
njostonehouse: I think the convention here is a method name with "compute" like `_computeIsAllTab`, but it doesn't really matter.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
```diff
@@ -346,6 +361,12 @@
 				return this._refetchEnrollments();
 			}
 
+			this.fire('d2l-course-data-change', {
```
njostonehouse: There's a newer way of firing events that is preferred:
```this.dispatchEvent(new CustomEvent('d2l-course-data-change`, {
    bubbles: [true|false],
    composed: [true|false],
    detail: {
       <your stuff>
    }
});```
https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Example

File: src/d2l-my-courses-content/d2l-my-courses-behavior.html [hunk 12]
```diff
@@ -51,15 +51,20 @@
...
 			},
-			_tabSearchType: String
+			_tabSearchType: String,
+			_changedCourseData: Object
 		},
 		_computeShowGroupByTabs: function(groups) {
 			return groups.length >= 2 || (groups.length > 0 && !this._enrollmentsSearchAction);
 		},
+		listeners: {
+			'd2l-course-data-change': '_onCourseDataChange'
```
njostonehouse: A minor thing, but the Course is not changing when there is a pinned event. The _Enrollment_ is what gets pinned, and whose data is modified. I'd prefer this was `d2l-course-enrollment-change` or something like that.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.html
edwardg8: Oh dam, they are doing the same thing, haha, I original have some additional logic there lately become redundant, and didn't realize they are the same now, will change it.

njostonehouse:
Consider writing some tests. Sometimes they can be a bit awkward with dispatching events and so on, so I don't think they should block this change. But see if you can add a bit of coverage separately.

edwardg8:
I think i'm going to merge this and added tests separately @njostonehouse , so it's easier for you to manage to convert to polymer3.

PR #658, opened 2019-01-15
Nstonehouse/us103365/polymer 3 whitespace

njostonehouse:


src/d2l-course-tile-styles.html
```diff
@@ -2,7 +2,7 @@
 <link rel="import" href="../../d2l-typography/d2l-typography-shared-styles.html">
 
 <dom-module id="d2l-course-tile-styles">
-	<template>
+	<template strip-whitespace>
```
dbatiste: Not sure this one needed it since it's just styles.  Probably doesn't hurt though.

PR #662, opened 2019-01-16
Fix some linting

svanherk:


src/d2l-filter-menu/d2l-filter-menu-tab-roles.js
```diff
@@ -66,7 +66,6 @@ Polymer({
 		}
 	},
 	behaviors: [
-		window.D2L.Hypermedia.HMConstantsBehavior,
```
svanherk: I don't think we need this anymore, because now we import `Actions` and use it directly?

njostonehouse: Yes! I must've missed this one, thank you.

PR #663, opened 2019-01-16
Remove self from codeowners

ryantmer:
Realistically, I haven't been up to date on this in months - if anybody wants to take over as a code owner, feel free - at this point, I'm really only useful for perhaps historical insight, not on PRs.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-* @ryantmer


m6jones:
Should you be replaced with @njostonehouse and @edwardg8? I am happy to be on there as well.

ryantmer:
Whatever works :) Generally I feel people can add themselves as codeowners when/if they feel comfortable doing so.

PR #670, opened 2019-01-25
DE31831: Add a segregated view demo.

njostonehouse:


edwardg8:
I just added another example to cover the case when there are no courses at all, can revert this if it's not useful.

edwardg8:
Sorry, clicked wrong when submitted the comment. The new added example is useful to me, so I can check if the "View All Courses" link show up or not when there are no courses.

njostonehouse:
The `push` CI is failing with ```**Test run ended in failure: Error thrown outside of test function: Cannot read property '$$' of null**```, not clear why. Will update this branch if I can figure out a fix.

njostonehouse:
I think the `$$` error got introduced in this merge: https://github.com/Brightspace/d2l-my-courses-ui/pull/668

edwardg8:
I think I've seen that error after Polymer3 conversion.

PR #672, opened 2019-01-29
DE31831: Make sure animation end clears the class of the departing tile.

njostonehouse:
The `unpin` class is added when a tile moves between the "pinned" and "unpinned" lists. This class kept the new location of incoming tiles marked as though the tile wasn't there. This messed up animations by hiding the tiles that slid into the slot of a departing tile.

This problem is only in Polymer 3 for reasons I haven't yet deduced.


edwardg8:
yea this is so weird, glad that you figured out this fix.

njostonehouse:
I think it's useful to merge this change even though there is a test problem with the `search-widget`. I'd rather not hold up the P3 upgrade because our tests are junk.

PR #674, opened 2019-01-30
DE31831: Synchronize a test with done.

njostonehouse:
This was causing the `$$` error.


njostonehouse:
This solved the `$$` problem but seems to have exposed a new one. Will try to address the `search-widget` issue in a different PR.

PR #676, opened 2019-01-31
DE32800: Only fetch notifications data if the component is configured…

njostonehouse:
… to show it.


njostonehouse:
This change ports a similar logic from Updated Sort Logic ON (the `organizations` implementation) to Updated Sort Logic OFF. It should keep a widget that is not configured to show Notifications from fetching from the Notifications API.

PR #677, opened 2019-01-31
DE32768-make overlay page able to refetch once there is pinned and un…

edwardg8:
…ppined action


edwardg8:
Broke the tests, will fix them

edwardg8:
Just realized my last commit broke the test, fixing it

src/d2l-all-courses.js
```diff
@@ -569,6 +579,19 @@ Polymer({
 	* Utility/helper functions
 	*/
 
+	_updateSearchUrlWithNewBustCacheToken: function() {
+		var s = this._searchUrl;
+		var r = '&bustCache=';
+		var n = s.indexOf(r);
```
njostonehouse: Aren't there libraries for parsing query strings? You could end up with some situations where `bustCache` isn't necessarily the last item in the string, right?

src/d2l-all-courses.js
```diff
@@ -370,7 +373,7 @@ Polymer({
 		}
 	},
 	_onFilterChanged: function(e) {
-		this._searchUrl = e.detail.url;
+		this._searchUrl = e.detail.url + '&bustCache=' + this._bustCacheToken;
```
njostonehouse: Again, to the point about URL parsing/building, this might need a leading `?` instead of a `&`, etc.

edwardg8: the old refetch method uses `&` I think, but I'm going to create a constant for this '&bustCache=' string at least

edwardg8: Oh it's because it already started query string from its `params` so it's adding one more to it.

src/d2l-all-courses.js
edwardg8: i think we always added `bustcache` at the end, but you are right, it can possibly be in other positions too.

src/d2l-all-courses.js
```diff
@@ -569,6 +591,23 @@ Polymer({
...
+			index += this._bustCacheStr.length;
+			var prefix = url.substring(0, index);
+			var suffix = url.substring(index, url.length);
+			index = suffix.indexOf('&');
+			suffix = index === -1 ? '' : suffix.substring(index, suffix.length);
+			this._searchUrl = prefix + this._bustCacheToken + suffix;
+		} else {
+			this._searchUrl += this._cacheQueryString;
+		}
+	},
```
edwardg8: This gets so ugly but works for the cases where `bustCache` isn't at the end

src/d2l-all-courses.js
```diff
@@ -569,6 +591,23 @@ Polymer({
 	* Utility/helper functions
 	*/
 
+	_updateSearchUrlWithNewBustCacheToken: function() {
```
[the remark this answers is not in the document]
edwardg8: Looks like the code is much easier to read after using the library. I don't know if we want to pull in this library just for this though. Maybe other code will benefit from this later?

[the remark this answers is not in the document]
njostonehouse: I think it's reasonable. This change is messing about with the url, so using a purpose-built library for doing that sounds like the right approach. My Courses shouldn't be too dedicated to the specifics of URLs.

src/d2l-all-courses.js
```diff
@@ -370,7 +380,10 @@ Polymer({
 		}
 	},
 	_onFilterChanged: function(e) {
-		this._searchUrl = e.detail.url;
+		if (e.detail.url) {
+			var queryString = e.detail.url.indexOf('?') !== -1 ? this._cacheQueryString : ('?' + this._bustCacheStr + this._bustCacheToken);
```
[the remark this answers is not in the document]
edwardg8: I don't see how `bustcache` query string can be added right away before any other events happened that overwrites the url. So that means we may fetch dirty cached data until some other events happened for correction.

src/d2l-all-courses.js
```diff
@@ -569,6 +581,22 @@ Polymer({
...
+	_createActionUrlWithBustCache: function(action, parameters) {
+		return this._appendBustCacheQueryString(this.createActionUrl(action, parameters));
+	},
+	_appendBustCacheQueryString: function(url) {
+		if (!url) {
+			return null;
+		}
+
+		var queryStr = this._bustCacheStr + '=' + this._bustCacheToken;
+		return url + (url.indexOf('?') !== -1 ? '&' : '?') + queryStr;
```
njostonehouse: Can you use `new URL` here too?

src/d2l-all-courses.js
```diff
@@ -569,6 +581,22 @@ Polymer({
 	* Utility/helper functions
 	*/
 
+	_createSearchUrlWithNewBustCacheToken: function() {
+		var parsedUrl = new URL(this._searchUrl);
+		parsedUrl.searchParams.set(this._bustCacheStr, this._bustCacheToken);
+		return parsedUrl.toString();
+	},
+	_createActionUrlWithBustCache: function(action, parameters) {
+		return this._appendBustCacheQueryString(this.createActionUrl(action, parameters));
```
njostonehouse: You could converge on a single Url-creating method with a `bustCache` boolean parameter, if that feels better to you.

src/d2l-all-courses.js
```diff
@@ -490,9 +496,15 @@ Polymer({
...
 			params.roles =  this._enrollmentsSearchAction.getFieldByName('roles').value;
 		}
-		this._searchUrl = this.createActionUrl(tabAction.enrollmentsSearchAction, params);
+		this._searchUrl = this._createActionUrlWithBustCache(tabAction.enrollmentsSearchAction, params);
+	},
+	_onEnrollmentPinned: function(e) {
+		this._bustCacheToken = Math.random();
+		var actionName = this._selectedTabId.replace('all-courses-tab-', '');
+		if (!e.detail.isPinned &&  actionName === Actions.enrollments.searchMyPinnedEnrollments) {
+			this._searchUrl = this._createSearchUrlWithNewBustCacheToken();
```
njostonehouse: Isn't this just `_appendBustCacheQueryString( this._searchUrl )`?

src/d2l-all-courses.js
```diff
@@ -255,7 +255,12 @@ Polymer({
 			type: Boolean,
 			value: false
 		},
-		_isSearched: Boolean
+		_isSearched: Boolean,
+		_bustCacheToken: Number,
+		_bustCacheStr: {
```
njostonehouse: You could probably hard-code this into your `_appendBustCacheQueryString` method, to save any potential for it to change.

src/d2l-all-courses.js
edwardg8: Not quite, it's replacing the token with new generated instead of append.

src/d2l-all-courses.js
edwardg8: I'm still seeking ways to parse the url...as I tested it's not working for IE11.

src/d2l-all-courses.js
edwardg8: `_appendBustCacheQueryString` takes parameter of url, this url might not be `this._searchUrl`, it's also used in `_onFilterChanged` which in that case it's passing in `e.detail.url`

src/d2l-all-courses.js
```diff
@@ -569,6 +577,21 @@ Polymer({
...
 
+	_createSearchUrlWithNewBustCacheToken: function() {
+		var parsedUrl = new URL(this._searchUrl);
+		parsedUrl.searchParams.set('bustCache', this._bustCacheToken);
+		return parsedUrl.toString();
+	},
+	_createActionUrlWithBustCache: function(action, parameters) {
+		return this._appendBustCacheQueryString(this.createActionUrl(action, parameters));
+	},
+	_appendBustCacheQueryString: function(url) {
```
njostonehouse: If `url` already has a `bustCache` key, will this add another one?

edwardg8: yea, it will add another one, hmm do we have the case where the existing url may have `bustCache`...i'll check. Maybe let's search if bustCache exists or not, if not then append otherwise replace.

edwardg8:
I combined all those functions to `_appendOrUpdateBustCacheQueryString` function which basically add bustCache query string if it does not exist, or replace it with new one if it already existed. And now make everywhere use `_appendOrUpdateBustCacheQueryString`  instead.

PR #679, opened 2019-02-05
Nstonehouse/de32844/polymer 3

njostonehouse:


njostonehouse:
I didn't realize we'd up with no changes here. This reapplies the Polymer 3 upgrade and the patch to `hybrid` since last time. Apparently all changes since then are compatible with no conflicts.

PR #682, opened 2019-02-08
Fix Edge RTL styles

dlockhart:
Replacing `:host-context([dir="rtl"])` usages with :host(:dir(rtl))`. Otherwise in Edge the RTL styles will be incorrectly applied in LTR situations.

### DIFF
--- a/src/d2l-course-tile-styles.js
+++ b/src/d2l-course-tile-styles.js
@@ -153,7 +153,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-styles">
 				display: flex;
 			}
 
-			:host-context([dir="rtl"]) .menu-bar {
+			:host(:dir(rtl)) .menu-bar {
 				left: 10px;
 				right: auto;
 			}
@@ -165,10 +165,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-styles">
 				padding-left: 0;
 				padding-right: 0;
 				opacity: 0;
-			}
-
-			:host-context([dir="rtl"]):not([pinned]) .pin-indicator {
-				margin-left: 0;
 			}
 
 			.pin-indicator {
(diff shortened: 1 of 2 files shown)


src/d2l-course-tile-styles.js
```diff
@@ -167,10 +167,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-styles">
 				opacity: 0;
 			}
 
-			:host-context([dir="rtl"]):not([pinned]) .pin-indicator {
-				margin-left: 0;
-			}
```
dlockhart: This rule doesn't appear to be doing anything, since non-pinned pin indicators are hidden anyway.

src/d2l-course-tile-styles.js
```diff
@@ -454,7 +450,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-styles">
 				right: -2px;
 			}
 
-			[dir="rtl"] .alert-color-circle {
+			:host(:dir(rtl)) .alert-color-circle {
```
dlockhart: This wasn't logged in the defect but I noticed it was a problem:

screen shot 2019-02-08 at 4 11 35 pm

In the above screenshot it's in RTL and the orange circle should be on the left.

src/d2l-filter-menu/d2l-filter-list-item-styles.js
```diff
@@ -18,7 +18,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-filter-list-item-styles">
 				margin-right: 0.5rem;
 			}
 
-			:host-context([dir="rtl"]) > d2l-icon {
+			:host(:dir(rtl)) > d2l-icon {
```
dlockhart: This technically wasn't broken as having a `>` prevents the Edge bug, but I fixed it anyway just because `:host-context` shouldn't be used in Polymer 3 and this was the last usage in my-courses.

PR #683, opened 2019-02-11
DE32866: Animate tiles by a wrapper with an ID, instead of by dom-rep…

njostonehouse:
…eated element.

### DIFF
--- a/src/d2l-course-tile-grid-styles.js
+++ b/src/d2l-course-tile-grid-styles.js
@@ -31,9 +31,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-grid-styles">
 			--four-column-width: 24.25%;
 		}
 
-		:host([hide-past-courses]) d2l-course-tile[past-course]:not([pinned]),
-		:host([limit-to-12]:not([hide-past-courses])) d2l-course-tile:not([pinned]):nth-child(n+13),
-		:host([limit-to-12]) d2l-course-tile:not([pinned]):not([past-course]):nth-child(n+13) {
+		:host([hide-past-courses]) d2l-course-tile[past-course]:not([pinned]) {
 			display: none;
 		}
 
--- a/src/d2l-course-tile-grid-styles.js
+++ b/src/d2l-course-tile-grid-styles.js
@@ -48,43 +48,43 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-grid-styles">
 			/* To offset 10px left padding on tile containers, where tile image left edge must align with headers */
 			margin-left: -10px;
 		}
-		.course-tile-container.animate > d2l-course-tile {
+		.course-tile-container.animate d2l-course-tile {
 			margin-right: 0;
 			margin-left: 0;
 		}
 
-		.grid-child {
+		.course-tile-item-container {
 			width: 100%;
 		}
 
-		.columns-2 > .grid-child {
+		.columns-2 .course-tile-item-container {
 			width: var(--two-column-width);
 		}
 
-		.columns-3 > .grid-child {
+		.columns-3 .course-tile-item-container {
 			width: var(--three-column-width);
 		}
 
-		.columns-4 > .grid-child {
+		.columns-4 .course-tile-item-container {
 			width: var(--four-column-width);
 		}
 
-		.columns-2 > .grid-child,
-		.columns-3 > .grid-child,
-		.columns-4 > .grid-child {
+		.columns-2 .course-tile-item-container,
+		.columns-3 .course-tile-item-container,
+		.columns-4 .course-tile-item-container {
 			margin-left: 0.5%;
 			margin-right: 0.5%;
 		}
 
-		.columns-2 > .grid-child:nth-child(2n),
-		.columns-3 > .grid-child:nth-child(3n),
-		.columns-4 > .grid-child:nth-child(4n) {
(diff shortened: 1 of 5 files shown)


src/d2l-course-tile-grid-styles.js
```diff
@@ -31,9 +31,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-grid-styles">
 			--four-column-width: 24.25%;
 		}
 
-		:host([hide-past-courses]) d2l-course-tile[past-course]:not([pinned]),
-		:host([limit-to-12]:not([hide-past-courses])) d2l-course-tile:not([pinned]):nth-child(n+13),
```
njostonehouse: We no longer limit the grid to 12 tiles. The 12 limit is only used with cards (a different component entirely).

src/d2l-course-tile-grid-styles.js
```diff
@@ -48,43 +46,43 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-course-tile-grid-styles">
...
 			margin-right: 0;
 			margin-left: 0;
 		}
 
-		.grid-child {
+		.course-tile-item-container {
 			width: 100%;
 		}
 
-		.columns-2 > .grid-child {
```
njostonehouse: With shadow DOM, these `>`s are no longer sufficient to reach the child.

src/d2l-course-tile-sliding-grid-behavior.js
```diff
@@ -76,13 +76,13 @@ D2L.MyCourses.CourseTileSlidingGridBehavior = {
 		'__slide_onEnrollmentsChanged(enrollments.*)'
 	],
 
-	__slide_knownGridTileElements: null,
+	__slide_knownGridTileElementIds: [],
```
njostonehouse: The most significant change here is to wrap the component in a uniquely-identified `div` wrapper. Those IDs are then tracked instead of the elements themselves, since the elements' contents are unreliable.

src/d2l-course-tile-sliding-grid-behavior.js
```diff
@@ -218,7 +220,9 @@ D2L.MyCourses.CourseTileSlidingGridBehavior = {
 		});
 	},
 
-	__slide_translateTile: function slideTranslateTile(tile, change, delay) {
+	__slide_translateTile: function slideTranslateTile(tileId, change, delay) {
+		var tile = this._getGridContainerElement().querySelector('[id="' + tileId + '"]');
```
njostonehouse: Slide the container with the matching ID, to make sure the right image slides about.

src/d2l-course-tile-grid.js
```diff
@@ -236,8 +238,10 @@ Polymer({
 	_getGridContainerElement: function() {
 		return this.$$('.course-tile-container');
 	},
-	_getGridTileElements: function() {
-		return this._getGridContainerElement().querySelectorAll('d2l-course-tile');
+	_getGridTileElementIds: function() {
+		return Array.from(this._getGridContainerElement().querySelectorAll('.course-tile-item-container')).map(function(tile) {
+			return tile.id;
```
njostonehouse: Grab the id of the tile's container so that if the `dom-repeat` re-writes the contents of the element, we still animate the right tile.

njostonehouse:
Polymer 2 and 3's `dom-repeat` has a curious way of re-using DOM elements when the bound array changes. Without being able to force re-stamping of the elements, these animations that depended on they elements being unmutated go berzerk. This change compensates for it by storing a map of IDs that need animating.

njostonehouse:
Restamp mode WIP: https://github.com/Polymer/polymer/pull/4363

test/d2l-course-tile-grid/d2l-course-tile-grid.js
```diff
@@ -77,15 +77,6 @@ describe('d2l-course-tile-grid', () => {
 			});
 		});
 
-		it('should only show 12 tiles when limit-to-12 attribute is set', () => {
```
njostonehouse: I'm bothered that these tests were apparently passing before. It seems like I did something to break them, but they no longer need to work, so burning them up.

src/d2l-course-tile-sliding-grid-behavior.js
dlockhart:
```suggestion
		var tile = this._getGridContainerElement().querySelector(`[id="${tileId}"]`);
```

src/d2l-course-tile-sliding-grid-behavior.js
```diff
@@ -189,10 +192,9 @@ D2L.MyCourses.CourseTileSlidingGridBehavior = {
 		var targetHeight = 0;
 
 		if (this.__slide_rowChange < 0) {
-			var tiles = this.__slide_knownGridTileElements;
+			var tiles = this.__slide_knownGridTileElementIds;
 			if (tiles.length > 0) {
-				var lastTile = tiles[tiles.length - 1];
-				targetHeight = lastTile.offsetTop;
+				targetHeight = this._getGridContainerElement().querySelector('[id="' + tiles[tiles.length - 1] + '"]').offsetTop;
```
dlockhart:
```suggestion
				targetHeight = this._getGridContainerElement().querySelector(`[id="${tiles[tiles.length - 1]}"]`).offsetTop;
```

PR #684, opened 2019-02-12
Added function that reset org from image selector

ciss1995:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -455,7 +456,15 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 			}, 10);
 		}
 		this._setLastSearchName(this.enrollmentsSearchAction.name);
-
+		//send event when tab is changed
```
njostonehouse: You don't need to bother with this comment.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -455,7 +456,15 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+		var tabChanged = new CustomEvent('d2l-tab-changed', {
+			bubbles: true,
+			composed: true,
+			detail: {
+				_tabId: this.enrollmentsSearchAction.name
```
njostonehouse: I'd suggest removing the `_` here, since I think that usually indicates a private/protected variable, but it is intentionally exposed.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -129,7 +130,7 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 		'started-inactive': '_onStartedInactiveAlert',
 		'd2l-enrollment-card-fetched': '_onD2lEnrollmentCardFetched',
 		'd2l-enrollment-card-status': '_onD2lEnrollmentCardStatus',
-		'd2l-enrollment-new': '_onD2lEnrollmentNew'
+		'd2l-enrollment-new': '_onD2lEnrollmentNew',
```
njostonehouse: nit: there's a comma here

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,27 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 			isPinned: e.detail.isPinned
 		};
 	},
+	_tabSelectedChanged: function(e) {
+		this.currentTabId = 'panel-'.concat(e.detail._tabId);
```
njostonehouse: You probably want to bubble up the `panel-xxxx` id in the event, rather than tacking it on here.

njostonehouse: 👍

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,27 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 	getLastOrgUnitId: function() {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').getLastOrgUnitId()
+			? (this._showGroupByTabs
+				? this._fetchCurrentTabCoursesContent().getLastOrgUnitId()
+				: this.$$('d2l-my-courses-content').getLastOrgUnitId())
 			: this.$$('d2l-my-courses-content-animated').getLastOrgUnitId();
 	},
+	_fetchCurrentTabCoursesContent: function() {
+		return this.$$('#' + this.currentTabId).firstChild;
```
njostonehouse: Will
```
`#{this.currentTabId}`
```

src/d2l-my-courses-content/d2l-my-courses-behavior.js
njostonehouse: You might want to consider template literals here too, if it works.

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,27 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 	courseImageUploadCompleted: function(success) {
 		return this.updatedSortLogic
 			? this.$$('d2l-my-courses-content').courseImageUploadCompleted(success)
 			: this.$$('d2l-my-courses-content-animated').courseImageUploadCompleted(success);
 	},
 	getLastOrgUnitId: function() {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').getLastOrgUnitId()
+			? (this._showGroupByTabs
+				? this._fetchCurrentTabCoursesContent().getLastOrgUnitId()
```
njostonehouse: When `_showGroupByTabs` is `true`, do we know there _must_ be tabs? I believe tabs are hidden if there are fewer than 2, but `_showGroupByTabs` could still be `true`, no?

src/d2l-my-courses-content/d2l-my-courses-behavior.js
ciss1995: I will try that

src/d2l-my-courses-content/d2l-my-courses-behavior.js
ciss1995: Hum I do not know this. I will have to check.

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,26 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 		};
 	},
+	_tabSelectedChanged: function(e) {
+		this.currentTabId = `panel-${e.detail.tabId}`;
+	},
 	courseImageUploadCompleted: function(success) {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').courseImageUploadCompleted(success)
+			? (this._showGroupByTabs
+				? this._fetchCurrentTabCoursesContent().courseImageUploadCompleted(success)
```
njostonehouse: I think if `_fetchCurrentTabCoursesContent` returns `null` or `undefined` you're going to have a bad time. This would happen, for example, if `courseImageUploadCompleted` is triggered before (or without) `_tabSelectedChanged`.

ciss1995: Does that ever happen? _fetchCurrentTabCoursesContent should always return the current tab? and if there are no tabs, it would not be called.

njostonehouse: I don't see anything here that would restrict it from happening. I think what you probably want to do is try to make a single function that returns a valid component and calls `courseImageUploadCompleted` on it. If there's nothing valid that can be found (eg: the selector matches nothing), then you should avoid trying to call a function on it and just move along.

njostonehouse: You can probably combine the 3 different ways of getting a `courseImageUploadCompleted`-able component into a single function.

njostonehouse: If we were using `out` parameters or something I'd suggest a `bool TryGetMyCoursesContent`, but you can just as easily check to make sure the result isn't null, or even better that it actually has the function you expect.

ciss1995: Oh okay, I see. In that case, I probably should be doing the same for getLastOrgUnitId as well.

njostonehouse: I think that's a good idea :)

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,31 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
+			? this._courseImageUploadCompletedHelper(success)
 			: this.$$('d2l-my-courses-content-animated').courseImageUploadCompleted(success);
 	},
 	getLastOrgUnitId: function() {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').getLastOrgUnitId()
+			? this._getLastOrgUnitIdHelper()
 			: this.$$('d2l-my-courses-content-animated').getLastOrgUnitId();
 	},
+	_courseImageUploadCompletedHelper: function(success) {
```
njostonehouse: I'd prefer to avoid calling functions "helper"s. Each function does do a specific thing. In this case, it finds the content component.

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,31 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 	courseImageUploadCompleted: function(success) {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').courseImageUploadCompleted(success)
+			? this._courseImageUploadCompletedHelper(success)
 			: this.$$('d2l-my-courses-content-animated').courseImageUploadCompleted(success);
 	},
 	getLastOrgUnitId: function() {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').getLastOrgUnitId()
+			? this._getLastOrgUnitIdHelper()
```
njostonehouse: If you include the `d2l-my-courses-content-animated` case in the `_getLastOrgUnitIdHelper` function, you can just call `getLastOrgUnitId()` on whatever calls back (instead of having two places that call `getLastOrgUnitId`).

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,16 +95,31 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 			isPinned: e.detail.isPinned
 		};
 	},
+	_tabSelectedChanged: function(e) {
+		this.currentTabId = `panel-${e.detail.tabId}`;
+	},
 	courseImageUploadCompleted: function(success) {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').courseImageUploadCompleted(success)
+			? this._courseImageUploadCompletedHelper(success)
```
njostonehouse: Similar to below, consider moving the `d2l-my-courses-content-animated` case into the "Helper" function and just call `.courseImageUploadCompleted(success)` here only.

src/d2l-my-courses-content/d2l-my-courses-behavior.js
```diff
@@ -94,15 +95,21 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 	courseImageUploadCompleted: function(success) {
-		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-content').courseImageUploadCompleted(success)
-			: this.$$('d2l-my-courses-content-animated').courseImageUploadCompleted(success);
+		return this._fetchContentComponent().courseImageUploadCompleted(success);
 	},
 	getLastOrgUnitId: function() {
+		return this._fetchContentComponent().getLastOrgUnitId();
+	},
+	_fetchContentComponent: function() {
```
njostonehouse: ❤️

PR #685, opened 2019-02-13
DE32879: Remove the hover class from a tile when animation completes.…

njostonehouse:
… Fixes the z-location of animated tiles, which were holding onto hover unnecessarily.

### DIFF
--- a/src/d2l-course-tile.js
+++ b/src/d2l-course-tile.js
@@ -570,7 +570,8 @@ Polymer({
 		}
 
 		this.toggleClass('unpin', false, this);
-		this._setCourseTileHovered(false);
+		this._pinAnimationInProgress = false;
+		this._unhoverCourseTile();
 	},
 	_onOrganizationResponse: function(organization) {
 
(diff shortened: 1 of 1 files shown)


njostonehouse:
Although `unpin` was removed when animation was complete in a previous commit, we also need to remove `hover` to restore the correct z index to the animated tiles.

svanherk:
I may have messed up my BSI build, but I'm still seeing this as an issue.  Let me try to rebuild

svanherk:
I'm seeing the `hover` effect be removed from the new tile sliding into place, but if I hover back over it and move away, it stays and never turns off.

njostonehouse:
> I'm seeing the `hover` effect be removed from the new tile sliding into place, but if I hover back over it and move away, it stays and never turns off.

Well that's charming. I'll look into it.

src/d2l-course-tile.js
```diff
@@ -570,6 +570,8 @@ Polymer({
 		}
 
 		this.toggleClass('unpin', false, this);
+		this._pinAnimationInProgress = false;
```
njostonehouse: Before the tile-reuse problem came into play, this didn't need to be explicitly set to `false` when the animation finished. Now, it stays `true` and causes problems for things like removing the `.hover` styles.

PR #690, opened 2019-02-25
DE32846-add sync mechanism between overlay and the widget

edwardg8:


src/d2l-all-courses.js
```diff
@@ -466,6 +471,10 @@ Polymer({
 		this.$.filterMenu.clearFilters();
 		this._filterText = this.localize('filtering.filter');
 		this._resetSortDropdown();
+		if (this.hasEnrollmentsChanged) {
```
edwardg8: if pin/unpin happened in the widget, make the overlay refetch, note if the overlay is opened second time, it doesn't do refetch, this forces refetch to happen

src/d2l-all-courses.js
```diff
@@ -507,6 +516,25 @@ Polymer({
...
+		}
+
+		this.dispatchEvent(new CustomEvent('d2l-course-enrollment-change', {
+			bubbles: true,
+			composed: true,
+			detail: {
+				orgUnitId: orgUnitId,
+				isPinned: isPinned
+			}
+		}));
```
edwardg8: This blocks of code is copied from `_onEnrollmentPinnedMessage` which is used to decode the event to find `isPinned` and `orgUnitId`

src/d2l-all-courses.js
```diff
@@ -703,6 +731,14 @@ Polymer({
...
-	}
+	},
+	_getOrgUnitIdFromHref: function(organizationHref) {
+		var match = /[0-9]+$/.exec(organizationHref);
+
+		if (!match) {
+			return;
+		}
+		return match[0];
+	},
```
edwardg8: Again this function is copied from `_onEnrollmentPinnedMessage`

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -114,6 +114,10 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 		_isPinnedTab: {
 			type: Boolean,
 			computed: '_computeIsPinnedTab(enrollmentsSearchAction.name)'
+		},
+		_hasEnrollmentsChanged: {
+			type: Boolean,
+			value: false
 		}
```
edwardg8: this property is used to tell the overlay if enrollments are changed by the widget.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -488,6 +481,13 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 		this._removeAlert('setCourseImageFailure');
 		// update the startedInactive alert in case the user changed the pinned states in the overlay
 		this._onStartedInactiveAlert();
+
+		if (this._isRefetchNeeded) {
+			this._handleEnrollmentsRefetch();
+		}
+
+		document.body.addEventListener('d2l-course-pinned-change', this._onEnrollmentPinnedMessage, true);
```
edwardg8: Somehow the event listener gets removed after overlay closes, we add it back.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -488,6 +481,13 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+		this._hasEnrollmentsChanged = false;
```
edwardg8: The overlay has refetched enrollments after it's opened, so we set the property to false after overlay closes.

src/d2l-all-courses.js
```diff
@@ -507,6 +516,25 @@ Polymer({
...
+		} else if (e.detail.enrollment && e.detail.enrollment.hasLinkByRel(Rels.organization)) {
+			orgUnitId = this._getOrgUnitIdFromHref(e.detail.enrollment.getLinkByRel(Rels.organization).href);
+		}
+
+		this.dispatchEvent(new CustomEvent('d2l-course-enrollment-change', {
+			bubbles: true,
+			composed: true,
+			detail: {
+				orgUnitId: orgUnitId,
+				isPinned: isPinned
```
njostonehouse: You could avoid the separate `var` by just using `e.detail.isPinned` here, I think.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -702,6 +704,20 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 				.then(this._populateEnrollments.bind(this));
 		}
 	},
+	_handleEnrollmentsRefetch: function() {
```
njostonehouse: Name suggestion: `refetchEnrollments`?

edwardg8: `refetchEnrollments` is taken sadly

njostonehouse: Oh, I might be misunderstanding this function. How is it different from `this.refetchEnrollments`?

src/d2l-all-courses.js
edwardg8: yes, will remove it.

src/d2l-all-courses.js
```diff
@@ -507,6 +516,25 @@ Polymer({
 		if (!e.detail.isPinned &&  actionName === Actions.enrollments.searchMyPinnedEnrollments) {
 			this._searchUrl = this._appendOrUpdateBustCacheQueryString(this._searchUrl);
 		}
+
+		var isPinned = e.detail.isPinned;
+		var orgUnitId;
```
njostonehouse: This feels to me like a `findOrgUnitId` helper function could be useful.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
njostonehouse: Can you check to see if it's there already? Not sure what would happen if you re-added it and it wasn't removed.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -702,6 +704,20 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+		this._resetEnrollments();
+
+		var completeFetch = function() {
+			this._showContent = true;
+		}.bind(this);
+
+		this._refetchEnrollments().then(function() {
+			window.dispatchEvent(new Event('resize'));
+			setTimeout(completeFetch, 1000);
+		}.bind(this)).catch(completeFetch);
```
njostonehouse: Does this promise library implement `finally`?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
edwardg8: `refetchEnrollments` only does refetch. This new function does slightly more, it shows and hide contents, reset enrollments, and then call `_refetchEnrollments`. Before it was in `_onTabSelected` first if statement, then I have it moved to separate function because the same thing is needed again when overlay closes and if there is pinning action happened from overlay.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
edwardg8: When I manually checked it with console logs, the event listener is always removed. But i'll add some code check to make sure.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
edwardg8: I'm not too sure about JS and polymer, this was originally copied from `_fetchRoot`. However, when I check online, it seems like IE doesn't support `finally`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

src/d2l-all-courses.js
edwardg8: this is very tied to the event `d2l-course-pinned-change` though, I can't think of another places use it?

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
edwardg8: I just found this, "If multiple identical EventListeners are registered on the same EventTarget with the same parameters, the duplicate instances are discarded. They do not cause the EventListener to be called twice, and they do not need to be removed manually with the removeEventListener method."
So if the event listener already exist (with same signature), it will be discarded, there is no harm

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
edwardg8: I just found this, "If multiple identical EventListeners are registered on the same EventTarget with the same parameters, the duplicate instances are discarded. They do not cause the EventListener to be called twice, and they do not need to be removed manually with the removeEventListener method."
So if the event listener already exist (with same signature), it will be discarded, there is no harm

PR #692, opened 2019-02-26
US104330: Use the generated URL instead of the user-submitted one.

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-content-animated.js
```diff
@@ -168,8 +168,7 @@ Polymer({
 			}
 			setTimeout(this._onStartedInactiveAlert.bind(this), 50);
 		} else {
-			this.fetchSirenEntity(this.enrollmentsUrl)
-				.then(this._refetchEnrollments.bind(this));
```
njostonehouse: `_refetchEnrollments` fetches the enrollments anyway, no need to do it here first. https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content/d2l-my-courses-content-behavior.js#L721

PR #694, opened 2019-02-26
DE32768-fix issue where _selectedTabId is not defined when no tabs

edwardg8:
Thanks @njostonehouse for finding this issue and suggested the fix.  You are totally right that `_selectedTabId` should be only used when `_showGroupByTabs` is true, and it fixed the undefined issue when no tabs. I have it tested with both feature off and `UpdatedSortLogic` off.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -261,7 +261,8 @@ Polymer({
 			value: false
 		},
 		_isSearched: Boolean,
-		_bustCacheToken: Number
+		_bustCacheToken: Number,
+		_selectedTabId: String
 	},
 	behaviors: [
 		D2L.PolymerBehaviors.Hypermedia.OrganizationHMBehavior,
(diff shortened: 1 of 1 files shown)


src/d2l-all-courses.js
```diff
@@ -511,10 +512,12 @@ Polymer({
 		);
 	},
 	_onEnrollmentPinned: function(e) {
-		this._bustCacheToken = Math.random();
-		var actionName = this._selectedTabId.replace('all-courses-tab-', '');
-		if (!e.detail.isPinned &&  actionName === Actions.enrollments.searchMyPinnedEnrollments) {
-			this._searchUrl = this._appendOrUpdateBustCacheQueryString(this._searchUrl);
+		if (this._showGroupByTabs) {
```
njostonehouse: Just to reduce nesting you could do
```if( !this._showGroupByTabs ) {
return;
}```

edwardg8: I still need the blocks of code below to send an event to the widget to update the view when overlay closes though, can return immediately. (unless we don't want old my courses view benefit from the new sync mechanism between overlay and widget?)

njostonehouse: Nope, my mental brace balancing has failed me. Carry on.

PR #696, opened 2019-02-27
Nstonehouse/us104330/remove show properties

njostonehouse:


d2l-my-courses.js
```diff
@@ -47,13 +47,10 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-my-courses">
 							<d2l-my-courses-content
 								advanced-search-url="[[advancedSearchUrl]]"
 								course-image-upload-cb="[[courseImageUploadCb]]"
-								course-updates-config="[[courseUpdatesConfig]]"
```
njostonehouse: These properties are only used in `d2l-my-courses-content-animated`, so removing them from normal `-content`.

src/d2l-my-courses-content/d2l-my-courses-content-animated.js
```diff
@@ -236,6 +236,33 @@ Polymer({
 		}
 	},
 
+	// Override for MyCoursesContentBehavior._openAllCoursesView
+	_openAllCoursesView: function(e) {
```
njostonehouse: This is a full copy of the old version of `_openAllCoursesView` to maintain as much parity with the old version (for now).

edwardg8: I observed some issues with overriding a function or property before, maybe it was just for that particular case since I see overrides in many other places too, but worth some manual testing to see the override works.

njostonehouse: Yeah, the file does a lot of this "override"ing, so it's not new to `-animated`, but I'll make sure it continues to work.

PR #699, opened 2019-03-01
US104432: Create a DEMO for Sort Logic Off mode.

njostonehouse:


demo/data/my-enrollments-embedded
```diff
@@ -0,0 +1,80 @@
+{
+  "class": [
+    "enrollments",
+    "collection"
+  ],
+  "entities": [
```
njostonehouse: `embedDepth=1` is used, and required by Sort Logic Off. Took a while to sort that one out.

edwardg8: Curious what happens when it doesn't have `embedDepth=1`?

edwardg8: And I don't see how it got used/added in this PR?

edwardg8:
Nice so good to have this demo page!

demo/data/my-enrollments-embedded
njostonehouse: The demo imitates how the server will act. With the `-animated` component, all its requests have `embedDepth=1`, which means it expects the data to be organized differently. This demo has a new type of data that will work; the old data (`embedDepth=0`) wasn't compatible with the demo.

PR #702, opened 2019-03-05
US104432: Move listener for the all courses overlay into the all cour…

njostonehouse:
…ses overlay.


njostonehouse:
Failed a bunch of unit tests, will update.

PR #704, opened 2019-03-05
Nstonehouse/us104432/tile grid listener

njostonehouse:


src/d2l-course-tile-grid.js
```diff
@@ -93,23 +93,23 @@ Polymer({
 		'enrollment-pinned': '_onEnrollmentPinAction',
 		'enrollment-unpinned': '_onEnrollmentPinAction'
 	},
+	attached: function() {
```
njostonehouse: By adding a listener right on the `d2l-course-tile-grid`, it means the higher-order components don't have to pass the event down anymore. This is useful because `d2l-course-tile-grid` is only used in sort logic off, so those higher level components now behave the same despite the setting.

src/d2l-course-tile.js
```diff
@@ -378,18 +378,16 @@ Polymer({
 			.then(this._displaySetImageResult.bind(this, true, true))
 			.catch(this._displaySetImageResult.bind(this, false));
 	},
-	setCourseImage: function(details) {
-		details = details || {};
-		var tileContainer = this.$$('.tile-container');
-		var newImageHref = this.getDefaultImageLink(details.image),
-			newSrcset = this.getImageSrcset(details.image, 'tile');
+	setCourseImage: function(image, status) {
```
njostonehouse: Since this isn't an event listener function, take in some real variables.

njostonehouse:
Dang it, a lint error. I think I need a semicolon.

PR #705, opened 2019-03-06
US104432: Include some demo data around started inactive courses with…

njostonehouse:
… sort logic off.

Fixup the course tile component to handle limited organization data.

### DIFF
--- a/src/d2l-course-tile.js
+++ b/src/d2l-course-tile.js
@@ -574,8 +574,11 @@ Polymer({
 		}
 		this._canChangeCourseImage = this._getCanChangeCourseImage(organization);
 		this._pinnedChanged();
-		this._courseSettingsLabel = this.localize('courseSettings', 'course', this._organization.properties.name);
-		this._coursePinButtonLabel = this.localize('coursePinButton', 'course', this._organization.properties.name);
+
+		if( organization.properties ) {
+			this._courseSettingsLabel = this.localize('courseSettings', 'course', this._organization.properties.name);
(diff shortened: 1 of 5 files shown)


src/d2l-course-tile.js
```diff
@@ -615,9 +618,13 @@ Polymer({
 		this.$$('#courseUpdates').setAttribute('aria-hidden', updates === 0 ? 'true' : 'false');
 	},
 	_checkDateBounds: function(organization) {
+		if( !organization.properties ) {
```
njostonehouse: During the demo, this was `undefined`.

src/d2l-course-tile.js
```diff
@@ -615,9 +618,13 @@ Polymer({
...
+			return;
+		}
+
 		var nowDate = Date.now();
-		var endDate = Date.parse(organization.properties.endDate);
-		var startDate = Date.parse(organization.properties.startDate);
+		var endDate = Date.parse(organization.properties.endDate || Number.MAX_VALUE);
+		var startDate = Date.parse(organization.properties.startDate || Number.MIN_VALUE);
```
edwardg8: interesting, I wonder what happens when endDate or startDate is undefined.

njostonehouse: Yeah, I'm not sure. Actually, I wonder if `Date.parse` fails on `MIN_VALUE` than it does on `undefined`. I'm going to revert this part to at least maintain the previous functionality.

edwardg8: yea, i'm not sure if the value can be negative, safer with 0.

PR #709, opened 2019-03-07
Nstonehouse/breakup css grid

njostonehouse:


src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -220,6 +211,15 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 			&& this._numberOfEnrollments !== 0
 			&& this._enrollments.length === 0;
 	},
+	_getTileGrid: function() {
+		return this.$$('.course-tile-grid');
```
njostonehouse: Only `-animated` has `cssGridView` set to `false`. Override the tile grid selector there.

src/d2l-my-courses-content/d2l-my-courses-content-behavior.js
```diff
@@ -220,6 +211,15 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+	},
+	_refreshTileGridImages: function() {
```
njostonehouse: Similarly, `CssGridView == true` iterates through the tiles to refresh each image, whereas `== false` has its own technique (overriden in `-animated`).

src/d2l-my-courses-content/d2l-my-courses-content-animated.js
```diff
@@ -267,6 +267,14 @@ Polymer({
 		return false;
 	},
 
+	_getTileGrid: function() {
```
njostonehouse: Overrides for `CssGridView == true` case.

edwardg8:
Looks all good, there is one thing I hate (i know it was from the old code) is the property name `cssGridView`, `hasCssGridView` may be a better name, but we are ok to leave it I think

njostonehouse:
This is part of a larger effort to get rid of `cssGridView`. It removes it from the `-content` components at least. Then I plan to rename it to `card-grid`.

PR #714, opened 2019-03-08
US104432: Rename references of CssGridView to Card-Grid, including cs…

njostonehouse:
…s and html.

### DIFF
--- a/src/card-grid/d2l-card-grid-behavior.js
+++ b/src/card-grid/d2l-card-grid-behavior.js
@@ -5,34 +5,23 @@ window.D2L = window.D2L || {};
 window.D2L.MyCourses = window.D2L.MyCourses || {};
 
 /*
-* @polymerBehavior D2L.MyCourses.CssGridBehavior
+* @polymerBehavior D2L.MyCourses.CardGridBehavior
 */
-D2L.MyCourses.CssGridBehavior = {
-	properties: {
-		cssGridView: {
-			type: Boolean,
-			value: true
-		}
-	},
-
+D2L.MyCourses.CardGridBehavior = {
 	attached: function() {
 		afterNextRender(this, function() {
-			if (this.cssGridView) {
-				window.addEventListener('resize', this._onResize.bind(this));
-				// Sets initial number of columns
-				this._onResize();
-			}
+			window.addEventListener('resize', this._onResize.bind(this));
+			// Sets initial number of columns
+			this._onResize();
 		}.bind(this));
 	},
 
 	detached: function() {
-		if (this.cssGridView) {
-			window.removeEventListener('resize', this._onResize.bind(this));
-		}
+		window.removeEventListener('resize', this._onResize.bind(this));
 	},
 
 	_onResize: function(ie11retryCount) {
-		var courseTileGrid = this.$$('.course-tile-grid');
+		var courseTileGrid = this.$$('.course-card-grid');
 		if (!courseTileGrid) {
 			return;
 		}
@@ -62,7 +51,7 @@ D2L.MyCourses.CssGridBehavior = {
 			return;
 		}
 
-		var courseTileDivs = dom(this.root).querySelectorAll('.course-tile-grid > d2l-enrollment-card');
+		var courseTileDivs = dom(this.root).querySelectorAll('.course-card-grid d2l-enrollment-card');
 		ie11retryCount = ie11retryCount || 0;
 		if (
 			ie11retryCount < 20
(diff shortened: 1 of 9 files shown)


src/d2l-my-courses-content-behavior.js
```diff
@@ -704,6 +703,5 @@ D2L.MyCourses.MyCoursesContentBehavior = [
 	D2L.PolymerBehaviors.MyCourses.LocalizeBehavior,
 	D2L.MyCourses.AlertBehavior,
 	D2L.MyCourses.UtilityBehavior,
-	D2L.MyCourses.CssGridBehavior,
```
njostonehouse: This was a key I'd missed before. By moving this into `d2l-my-courses-content`, we avoid having the `CssGrid` code inside `-animated`, and can refactor any of the considerations `CssGridBehavior` used to have to exclude running its functions in `-animated`.

src/d2l-all-courses.js
```diff
@@ -767,6 +767,6 @@ Polymer({
 			return;
 		}
 		return match[0];
-	},
```
njostonehouse: Surprised lint didn't catch this.

src/card-grid/d2l-card-grid-styles.js
```diff
@@ -1,40 +1,40 @@
...
 				-ms-grid-columns: 1fr 15px 1fr 15px 1fr;
 			}
-			.course-tile-grid.columns-4 {
+			.course-card-grid.columns-4 {
 				grid-template-columns: repeat(4, 1fr);
 				-ms-grid-columns: 1fr 15px 1fr 15px 1fr 15px 1fr;
 			}
 
-			.course-tile-grid > div,
-			.course-tile-grid > d2l-enrollment-card {
```
njostonehouse: I'm not certain this had a noticeable change, but the `>` selector almost certainly wasn't working after the P3 upgrade.

src/card-grid/d2l-card-grid-behavior.js
```diff
@@ -89,9 +74,4 @@ D2L.MyCourses.CssGridBehavior = {
 		}
 	},
 
-	__hideElement: function(div) {
-		return this.is !== 'd2l-all-courses-unified-content'
```
njostonehouse: This is actually only used as part of `-unified-content` now, so this case isn't useful. Always returns `false`.

src/card-grid/d2l-card-grid-behavior.js
```diff
@@ -76,10 +65,6 @@ D2L.MyCourses.CssGridBehavior = {
 
 		for (var i = 0, position = 0; i < courseTileDivs.length; i++, position++) {
 			var div = courseTileDivs[i];
-			if (this.__hideElement(div)) {
```
njostonehouse: Since this is always `false` (see below), it can be dropped.

src/card-grid/d2l-card-grid-behavior.js
```diff
@@ -5,34 +5,23 @@ window.D2L = window.D2L || {};
...
-				this._onResize();
-			}
+			window.addEventListener('resize', this._onResize.bind(this));
+			// Sets initial number of columns
+			this._onResize();
 		}.bind(this));
 	},
 
 	detached: function() {
-		if (this.cssGridView) {
```
njostonehouse: Now that there's no use for the "disabled" version of these behaviours (in `-animated`), drop the conditions.

src/card-grid/d2l-card-grid-behavior.js
```diff
@@ -5,34 +5,23 @@ window.D2L = window.D2L || {};
 window.D2L.MyCourses = window.D2L.MyCourses || {};
 
 /*
-* @polymerBehavior D2L.MyCourses.CssGridBehavior
+* @polymerBehavior D2L.MyCourses.CardGridBehavior
 */
-D2L.MyCourses.CssGridBehavior = {
-	properties: {
-		cssGridView: {
```
njostonehouse: Good riddance.

src/card-grid/d2l-card-grid-behavior.js
```diff
@@ -62,7 +51,7 @@ D2L.MyCourses.CssGridBehavior = {
 			return;
 		}
 
-		var courseTileDivs = dom(this.root).querySelectorAll('.course-tile-grid > d2l-enrollment-card');
+		var courseTileDivs = dom(this.root).querySelectorAll('.course-card-grid d2l-enrollment-card');
```
ciss1995: Is the  `> selector` no longer needed here?

njostonehouse: The `>` implies that the child is a direct ancestor. Because of Polymer 3 (Shadow DOM), these selectors tend to fail.

njostonehouse: (The shadow root ends up between `.course-tile-grid` and `d2l-enrollment-card`)

edwardg8: Yep, removing them is better

ciss1995: Yeah I see! I didn't know that they were failing because of Polymer 3.

PR #715, opened 2019-03-14
Made Live Search find input element and make it fetch course cards

ciss1995:
Made Live Search find input element and make it fetch course cards instead of listbox

### DIFF
--- a/src/search-filter/d2l-search-widget-custom.js
+++ b/src/search-filter/d2l-search-widget-custom.js
@@ -14,7 +14,6 @@ import '@polymer/iron-input/iron-input.js';
 import '@polymer/iron-pages/iron-pages.js';
 import 'd2l-dropdown/d2l-dropdown.js';
 import 'd2l-dropdown/d2l-dropdown-content.js';
-import { Rels } from 'd2l-hypermedia-constants';
 import 'd2l-icons/d2l-icons.js';
 import 'd2l-search-widget/d2l-search-widget-behavior.js';
 import 'd2l-search-widget/d2l-search-widget-styles.js';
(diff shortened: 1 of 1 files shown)


src/search-filter/d2l-search-widget-custom.js
```diff
@@ -325,54 +301,9 @@ Polymer({
...
-			var result = this.fetchSirenEntity(url)
-				.then(function(organizationEntity) {
-					// Polymer's splice method causes empty results to show up for an
-					// unknown reason, so use a normal Array.prototype.splice
-					this._liveSearchResults.splice(i, 0, {
-						name: organizationEntity.properties.name
-					});
-				}.bind(this));
-			fetches.push(result);
+			this._setSearchUrl(this._liveSearchPageSize, 'searchUrl');
```
ciss1995: Fetch the cards directly instead of fetching the course name as suggestions in listbox

ciss1995: Performance cost seems similar. I am trying to understand and run the Tracing GUI to track the number of db calls.

src/search-filter/d2l-search-widget-custom.js
```diff
@@ -325,54 +301,9 @@ Polymer({
 
 	_liveSearch: function() {
 		if (this._searchInput.trim().length > 0) {
-			this._setSearchUrl(this._liveSearchPageSize, '_liveSearchUrl');
```
njostonehouse: This still feels like unexpected behaviour to me, but it's not like we've changed it for the worse, I don't think. At worst, I'd say we could stand to keep the number of results at 5 if we were nervous.

src/search-filter/d2l-search-widget-custom.js
```diff
@@ -325,54 +301,9 @@ Polymer({
...
-			if (req.abort) {
-				req.abort();
-			}
-		});
-		this.set('_organizationRequests', []);
-
-		return this.fetchSirenEntity(this._liveSearchUrl)
-			.then(this._onLiveSearchResponse.bind(this));
-	},
-	_onLiveSearchResponse: function(enrollmentsEntity) {
```
njostonehouse: Goodbye live search suggestions! Don't let the door hit you on the way out!

PR #716, opened 2019-04-01
Componentize all-courses

ciss1995:


ciss1995:
Breaking down this PR in 3: 
1/Data and demo update
3/Fetch Presentation and set attributes
2/Make All courses use entity store fetch

PR #718, opened 2019-04-02
Fetch presentation and set attributes

ciss1995:


m6jones:
I wanted to mention some where in my courses I had to make a call to the presentation to clear the cache. If you see the presentation not updating then you need to clear the presentation from the cache.

edwardg8:
I'm ok to merge this, I have a few things changed when componentizing `d2l-my-courses-content` (I branched off from this)

edwardg8:
> Also good with merging. I think on the next pass, we should take a good look at each component and try to clean up/centralize a lot of this logic. I'm sure a big part of the problem is the multiple code branches in my-courses though

Yea, because all these components are dependent on each other, easy running into conflicts, I think next pass won't be as bad since token and attributes stuff are in place.

PR #720, opened 2019-04-03
Use EntityStore Fetch

ciss1995:


edwardg8:
I'm surprised there are no tests

ciss1995:
> I'm surprised there are no tests

Yes, I am looking at the tests now.

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -139,7 +193,7 @@ describe('d2l-all-courses', function() {
 					roles: 0
 				}
 			});
-
+			console.log();
```
edwardg8: There is console.log here

ciss1995: Yup, thank you.

ciss1995:
Made no changes to the behavior of the tests, just reorganized it a little. Also, changing the way we fetch does not seem to break any test.
However, I can't seem to figure out what is causing this error and where it is coming from```Error thrown outside of test function: Cannot read property 'search-widget' of undefined
  Context.<anonymous> at d2l-all-courses.js:28```. I previously had a discussion with @njostonehouse about this error, which seems to have been around for a while. Should we attempt to fix it?

svanherk:
I think we definitely want these tests to be green when we walk away from them.  That doesn't need to be right now or even this pass, but I think the sooner the better?  That old firefox issue is gone with Polymer 3, so in theory we _can_ get these back to green.  I can try taking a look at the error

ciss1995:
> I think we definitely want these tests to be green when we walk away from them. That doesn't need to be right now or even this pass, but I think the sooner the better? That old firefox issue is gone with Polymer 3, so in theory we _can_ get these back to green. I can try taking a look at the error

Odd thing is that I can't see this error in the console of the tests. (All the tests seem to pass). Thanks for taking a look.

src/d2l-all-courses.js
```diff
@@ -450,9 +450,9 @@ Polymer({
 				this._enrollmentsSearchUrl = lastResponseEntity.getLinkByRel('next').href;
 				this.$.lazyLoadSpinner.scrollIntoView();
 
-				return this.fetchSirenEntity(this._enrollmentsSearchUrl)
+				return window.D2L.Siren.EntityStore.fetch(this._enrollmentsSearchUrl, this.token)
 					.then(function(enrollmentsEntity) {
-						this._updateFilteredEnrollments(enrollmentsEntity, true);
+						this._updateFilteredEnrollments(enrollmentsEntity.entity, true);
```
edwardg8: `enrollmentsEntity` can be null here though, safer with the check

PR #722, opened 2019-04-03
Us105714 entity store fetch my courses content animated

ciss1995:


src/tile-grid/d2l-my-courses-content-animated.js
```diff
@@ -309,6 +309,10 @@ Polymer({
 	},
 	// Override for MyCoursesContentBehavior._populateEnrollments
 	_populateEnrollments: function(enrollmentsEntity) {
+		if (enrollmentsEntity.entity) {
+			enrollmentsEntity = enrollmentsEntity.entity;
+		}
```
edwardg8: what happens if `enrollmentsEntity.entity is undefined, you may want to reverse this, so return when it's null

ciss1995: So we only need to care about enrollmentsEntity.entity when we recursively call `_populateEnrollments` from ```return window.D2L.Siren.EntityStore.fetch(url, this.token)
				.then(this._populateEnrollments.bind(this));```, which only happen `if (lastEnrollment.hasClass('pinned') && this._hasMoreEnrollments)`. I added this if statement because when call `_populateEnrollments`, enrollmentsEntity is an entity. However, when we call `_populateEnrollments` from the return statement, we need enrollmentsEntity.entity to get the entity.

edwardg8: Create a helper function/wrapper for the case where it uses enrollmentsEntity.entity then

ciss1995: Yeah ok good idea

src/tile-grid/d2l-my-courses-content-animated.js
```diff
@@ -307,6 +307,9 @@ Polymer({
 		}
 		return enrollmentsSearchUrl;
 	},
+	_populateNextPinnedEnrollments: function(enrollmentsEntity) {
+		return this._populateEnrollments(enrollmentsEntity);
```
edwardg8: Shouldn't this be `enrollmentsEntity.entity`?

ciss1995: Yes absolutely. Good catch! This was the whole point of this helper function.

edwardg8: I think you need null check here too

PR #725, opened 2019-04-05
Egao/componentize tab content

edwardg8:


edwardg8:
Closed previous PR because I messed up the branch.

src/d2l-my-courses-content-behavior.js
```diff
@@ -25,12 +27,7 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 			value: function() { return []; }
 		},
 		tabSearchType: String,
-		userSettingsUrl: String,
-		// URL to fetch widget settings
-		presentationUrl: {
```
ciss1995: Since we no longer have `presentationUrl` we can remove it from https://github.com/Brightspace/d2l-my-courses-ui/blob/d9747a2ce4ddd4450cc5e155134544cb46eacb24/src/d2l-my-courses-content-behavior.js#L667 It is also no longer used in `all-courses`. Also maybe this https://github.com/Brightspace/d2l-my-courses-ui/blob/d9747a2ce4ddd4450cc5e155134544cb46eacb24/src/d2l-my-courses-behavior.js#L38 could be called href as well??

edwardg8: I think we can just remove them

edwardg8: we should leave it as presentationUrl as href is more reserved for its own component.

ciss1995: Ok, no problem

edwardg8: Oh nice `presentationUrl` is already removed from all courses, just have it removed from course content now, thanks Cheick

PR #726, opened 2019-04-08
Fix My Courses Tests in Sauce

svanherk:
Let's try to get these builds back to green!  A number of these changes are cleanup that needs t be done after the move to Polymer 3.  The actual fix is just moving the language build files into the `src` folder.  When they are outside this folder, Sauce stalls trying to load them and unrelated error messages are thrown for each test.


m6jones:
WOOT my courses not going to be super flaky?

svanherk:
> WOOT my courses not going to be super flaky?

That's the hope!

gulpfile.js
```diff
@@ -29,7 +29,7 @@ function build() {
 	const options = {
 		client: true,
 		strict: true,
-		root: 'build/lang',
+		root: 'src/build/lang',
```
svanherk: This is the main change.  The other changes to this file and the template are to get the current `build` files and the `gulp` output back in sync after the Polymer 3 migration changes.

src/build/lang/ar.js
```diff
@@ -1,4 +1,3 @@
-import '@polymer/polymer/polymer-legacy.js';
```
svanherk: This actually isn't needed, so removed from the template.

test/index.all.html
```diff
@@ -8,7 +8,7 @@
 	</head>
 	<body>
 		<script>
-			var testVariants = [], tests = [
+			WCT.loadSuites([
```
svanherk: This should help cut the build time.  Now that we are on Polymer 3 and all browsers are using Shadow DOM if they can, Shady DOM if they don't have it, we're good to just test the version each browser actually uses.

njostonehouse:
oh my goodness thank you

PR #728, opened 2019-04-09
US105952 Fix tests for content animated

ciss1995:


src/tile-grid/d2l-my-courses-content-animated.js
```diff
@@ -307,11 +307,13 @@ Polymer({
 		}
 		return enrollmentsSearchUrl;
 	},
-	_populateNextPinnedEnrollments: function(enrollmentsEntity) {
-		return this._populateEnrollments(enrollmentsEntity.entity);
-	},
```
edwardg8: so cannot use the wrapper?

ciss1995: We don't need to. We needed it when `d2l-content-behavior` was not using the EntityStore.Fetch. Now that it does, we are getting {entity: entity} straight in `populateEnrollment`. 
This is why the demo page was broken.

PR #730, opened 2019-04-16
US105906 Show Learning Paths based on Tool Enablement

IhabAlkabary:


src/d2l-all-courses.js
```diff
@@ -411,6 +416,7 @@ Polymer({
 		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
 			this.createActionUrl(this.enrollmentsSearchAction, {
 				autoPinCourses: false,
+				orgUnitTypeId: this.showLearningPaths ? '3&orgUnitTypeId=7' : '3',
```
m6jones: This seems sloppy. createActionUrl probably encodes the parameter values which may break this? Does this work?

IhabAlkabary: It works for sure.
Better ideas appreciated. I like the CSV stuff, but for some reason it fell out of fashion.

IhabAlkabary: and why would it encode it _while_ it's building the query params in the code?

m6jones: I would suggestion modify createActionUrl to accept array parameters  `orgUnitTypeId: ['7', '3'] : '3'` This seems easier. But if it works then maybe it doesn't matter. That seems like a lot of work for not much gain.

 createActionUrl should be encoding the parameters to be compliant with shoving them into a URL so that `&` is encoded and everything would fail. So in the future someone might fix `createActionUrl` and this will break.

IhabAlkabary: I'm saying why would it encode it _while it's still_ building the query string. That would be a final step. Probably there are better ways to build the query string but not sure this is part of this PR.
Let's see what the code owners want here. Let's see if @njostonehouse has time.

m6jones: I am one of the code owners :O But Noel opinion would be great.

edwardg8:
We have some breaking changes in my courses UI, @ciss1995 is working on making the LE app passing token function down to my courses, and then we'll release my courses along with BSI and LE changes. Before that happens, if we release BSI of newer my courses, my courses will be broken I think. Though Cheick had some luck where no token is present to my courses, the app still works (not sure how reliable it is).

edwardg8:
We are close to have a solution to this, but not quite there yet.

src/d2l-all-courses.js
njostonehouse: I can see why using `3&orgUnitTypeId=7` makes sense, but it feels like it's taking advantage of implementation assumptions of `createActionUrl` that might not stay true. I'd prefer either to have `createActionUrl` accept an array of orgUnitTypeIds as a key's value, or that it can handle multiple entries with the same key.

d2l-my-courses.js
```diff
@@ -31,6 +31,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-my-courses">
 				enrollments-search-action="[[_enrollmentsSearchAction]]"
 				image-catalog-location="[[imageCatalogLocation]]"
 				show-course-code="[[showCourseCode]]"
+				show-learning-paths="[[showLearningPaths]]"
```
njostonehouse: Instead of passing this boolean all the way through, consider using an orgUnitType array of `[ 3 ]` or `[ 3, 7 ]` that gets used in the deeper components. Even this high-level component could benefit from that, instead of a boolean, then all the configuration can happen in `le`.

njostonehouse: This would help other consumers to be able to use whatever OUID is correct for their instance, if it's not 3, for example.

IhabAlkabary: I didn't think that OrgUnitTypes changes are that common, but sure, let's make it more generic

src/d2l-all-courses.js
IhabAlkabary: Sure, but do you know why we are moving away from CSVs in general? it is the join and split that is needed?

IhabAlkabary: Also, in this case, I'm seeing putting a special case if the field is OrgUnitTypeId, then I do the code that collapses the entries. Other option will be to check if the type is an array. Generic check for an array better?

Something like:
```
function isArray (value) {
return value && typeof value === 'object' && value.constructor === Array;
}
```

d2l-my-courses.js
njostonehouse: No, you're right, I think we assume Course Offerings are always 3. However, now that there's a case for using something other than `3`, it makes sense to me to move to something more generic.

src/d2l-all-courses.js
njostonehouse: My preference would be to handle 
```
this.createActionUrl(this.enrollmentsSearchAction, {
   orgUnitTypeId: 3,
   orgUnitTypeId: 7
}
```
It seems more intuitive to me that that would result in `&orgUnitTypeId=3&orgUnitTypeId=7`. Not sure if Javascript Objects accept something like that though.

njostonehouse: Otherwise, yes, I'd expect some special handling for arrays (as opposed to `OrgUnitTypeId` specifically)

IhabAlkabary: is this even possible in JS:
```
this.createActionUrl(this.enrollmentsSearchAction, {
   orgUnitTypeId: 3,
   orgUnitTypeId: 7
});
```
wouldn't 7 override 3

IhabAlkabary: I was thinking:

```
this.createActionUrl(this.enrollmentsSearchAction, {
   orgUnitTypeId: [3,7]
});
```

njostonehouse: Yep, either one seems good to me. The "duplicate keys" one seems more intuitive, but it might break JS syntax, so it'd be a non-starter if that's the case.

m6jones: I totally agree with Noel but it sadly breaks javascript

d2l-my-courses.js
IhabAlkabary: We can make this generalization as a future story since we agreed on passing a Boolean here and it's not an immediate need:
https://git.dev.d2l/projects/CORE/repos/le/pull-requests/14863/overview
I created a story for this here: 
https://rally1.rallydev.com/#/39602890252d/detail/userstory/301730748608

njostonehouse: I think we should make the time to do it now in order to avoid introducing the concept of learning paths to this component. Instead, let's decrease its reliance on OU Type knowledge, rather than increase it.

IhabAlkabary: I'm not saying put this in a backlog story, more like push to the next sprint since we have more urgent things to attend to. This again has changes to be done in the LE that we all agreed on the current approach before, I don't mind us changing our mind to make things more flexible and better for the future but we also need to assess urgency and we have other more urgent things to work on this sprint. Branching has already happened, so I would go for followup story.

m6jones: What Noel has mention makes sense, maybe we should just do it now. Shouldn't actually take that long to fix.

njostonehouse: When I approved the change in the LE, I didn't intend to implicitly approve the implementation, or even the API of `show-learning-paths` within this component, since I hadn't seen your proposed changes. Ideally, `show-learning-paths` would have been added here first, before the LE started to reference it, since as it stands, I don't believe the LE change does anything. One of the benefits of doing dependent changes first, is that you can get feedback on the implementation before relying on it and having to back out your assumptions that it would be approved.

src/d2l-utility-behavior.js
```diff
@@ -14,12 +14,28 @@ D2L.MyCourses.UtilityBehavior = {
 		parameters = parameters || {};
 		action.fields = action.fields || [];
 		var query = {};
+		var val;
 
 		action.fields.forEach(function(field) {
```
njostonehouse: Make sure to add some tests around the new behaviour.

d2l-my-courses.js
IhabAlkabary: I'm leaving this PR as is for now. I'll get back to it in the future when there is time.

src/d2l-utility-behavior.js
IhabAlkabary: sure, when I reopen it the PR

IhabAlkabary:
- [ ] Handle empty OrgUnitTypeIds

src/d2l-my-courses-content-behavior.js
```diff
@@ -166,6 +166,7 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 		document.body.addEventListener('set-course-image', this._onSetCourseImage.bind(this));
 		document.body.addEventListener('d2l-tab-panel-selected', this._onTabSelected.bind(this));
 		this.$['image-selector-threshold'].scrollTarget = this.$['basic-image-selector-overlay'].scrollRegion;
+		this.orgUnitTypeIds = JSON.parse(this.orgUnitTypeIds).value;
```
IhabAlkabary: Let me know if this is the best place to do the parsing, since I'm not a Polymer expert.
Shall I put it in ready instead?

njostonehouse: I think you could do worse than waiting to parse until you need the individual OUIDs. In `createActionUrl` for example.

IhabAlkabary: I'm going to try one last thing to simplify the code even more. But I have no clue why linting started failing like this.

njostonehouse: You might be getting hit with Travis upgrading to Node 12 and everything being terrible. Consider locking to Node 11. Eg: https://github.com/Brightspace/brightspace-integration/pull/1346/files

IhabAlkabary: yep, that fixed the linting.

IhabAlkabary: I think the PR is good to go but not sure why this test is failing, doesn't seem that the new code is related.

```
[BABEL] Note: The code generator has deoptimised the styling of undefined as it exceeds the max of 500KB.
chrome 74                Beginning tests via http://localhost:8081/components/d2l-my-courses/generated-index.html?cli_browser_id=0
chrome failed to maximize
chrome 74                ✖ test/d2l-my-courses-content/d2l-my-courses-content.html » d2l-my-courses-content » Card grid » should set the columns-"n" class on the correct card grid on resize
  expected 'course-card-grid' to include 'columns-'
    <unknown> at   <unknown> at d2l-my-courses-content.js:295:0
chrome 74                Tests failed: 1 failed tests
Test run ended in failure: 1 failed tests
chrome 74 (354/1/1)                   
Error: 1 failed tests
```

IhabAlkabary: https://github.com/Brightspace/d2l-my-courses-ui/pull/731
Based on the previous PR off master, looks like it's related to the new code somehow, or maybe how the test is written, I'll investigate.

IhabAlkabary: > I think you could do worse than waiting to parse until you need the individual OUIDs. In `createActionUrl` for example.

Not sure what you mean here, do you want to push the parsing code to `createActionUrl`? but that will make it need to have specific code for handling `orgUnitTypeIds`

njostonehouse: The result of JSON.parse should indicate to `createActionUrl` whether the value is an array or not, I think?

IhabAlkabary: yes, but we need to put that in a try catch which would be a code smell IMO

src/d2l-all-courses.js
```diff
@@ -411,6 +416,7 @@ Polymer({
 		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
 			this.createActionUrl(this.enrollmentsSearchAction, {
 				autoPinCourses: false,
+				orgUnitTypeId: this.orgUnitTypeIds,
```
AKobets: should var name be `orgUnitTypeIds` ?

IhabAlkabary: it shows like this: `orgUnitTypeId=3&orgUnitTypeId=7&...`

m6jones:
We shouldn't be using `try catch` to do logic.

IhabAlkabary:
Thanks Noel and Maya for the review.

PR #743, opened 2019-05-09
US106667 Restructure sortLogicOff code

ciss1995:
This is SortLogicOff form version 9. 
I will update it to version 8.


d2l-my-courses.js
```diff
@@ -15,10 +15,13 @@ import '@polymer/polymer/polymer-legacy.js';
 
 import 'd2l-tabs/d2l-tabs.js';
 import './src/card-grid/d2l-my-courses-content.js';
-import './src/tile-grid/d2l-my-courses-content-animated.js';
 import './src/d2l-my-courses-behavior.js';
 import './src/d2l-utility-behavior.js';
```
edwardg8: Remove the imports `./src` maybe?

ciss1995: I was not sure because we need all the `./src` in `d2l-my-courses` for `'./src/card-grid/d2l-my-courses-content.js';` and we also need all the `./legacy` for `'./legacy/tile-grid/d2l-my-courses-content-animated.js';`.

edwardg8: oh you are right, this the place where we combine both

legacy/d2l-my-courses-content-behavior.js
```diff
@@ -0,0 +1,797 @@
...
+				'value': id
+			}
+		]);
+	}
+};
+
+/*
+* @polymerBehavior D2L.MyCourses.MyCoursesContentBehavior
+*/
+D2L.MyCourses.MyCoursesContentBehavior = [
```
edwardg8: interesting, this isn't giving collision since the two behavior names are the same? Seems like only web component name matters, behaviors aren't.

m6jones: That does make sense? It probably uses the last one defined.

svanherk: I think we should change it to be safe/sure

edwardg8:
I don't see any code removal though, can we delete `SortLogicOff` code from `src` too?

m6jones:
> I don't see any code removal though, can we delete `SortLogicOff` code from `src` too?

Should we do that in a second PR?

ciss1995:
@edwardg8 @m6jones I actually deleted them. I am not sure why it isn't showing here.

legacy/d2l-my-courses-content-behavior.js
ciss1995: I changed most of the behavior. I must have forgotten this one.

d2l-my-courses.js
svanherk: So... do `d2l-utility-behavior` and `localize-behavior` even need to be imported?  They're not included in the `behaviors` array below, and I think they're imported in by the `d2l-my-courses-behavior`

edwardg8:
A few things we want to make sure before the merge:
1. All tests pass (checked)
2. Demo page works (make sure that moves to legacy folder too)
3. Try it with LMS to see if it works

d2l-my-courses.js
```diff
@@ -15,10 +15,13 @@ import '@polymer/polymer/polymer-legacy.js';
...
 import './src/localize-behavior.js';
+import './legacy/tile-grid/d2l-my-courses-content-animated.js';
+import './legacy/d2l-my-courses-behavior.js';
```
svanherk: We can't have this same behaviour loaded into the file, all the properties will just override each other.  I think we need to create some wrappers...

d2l-my-courses.js
```diff
@@ -87,6 +67,7 @@ document.head.appendChild($_documentContainer.content);
 Polymer({
 	is: 'd2l-my-courses',
 	behaviors: [
-		D2L.MyCourses.MyCoursesBehavior
+		D2L.MyCourses.MyCoursesBehavior,
```
svanherk: I don't think we want these behaviors, I think we just want to list the properties here that we are passing in.  We aren't even importing these behaviours

d2l-my-courses.js
```diff
@@ -14,70 +14,50 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
 import '@polymer/polymer/polymer-legacy.js';
 
 import 'd2l-tabs/d2l-tabs.js';
```
svanherk: No longer need tabs here, we need it in `d2l-my-courses-container`

d2l-my-courses.js
```diff
@@ -14,70 +14,50 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
...
+				standard-semester-name="[[standardSemesterName]]"
+				standard-department-name="[[standardDepartmentName]]"
+				show-course-code="[[showCourseCode]]"
+				course-updates-config="[[courseUpdatesConfig]]"
+				image-catalog-location="[[imageCatalogLocation]]"
+				promoted-searches="[[promotedSearches]]"
+				user-settings-url="[[userSettingsUrl]]"
+				show-semester="[[showSemester]]"
+				org-unit-type-ids="[[orgUnitTypeIds]]"
+				updated-sort-logic="[[updatedSortLogic]]"
```
svanherk: Do we need to pass this in?  I think we can just use it at this level, and after this point it doesn't matter?  Or do we need to know further down?

m6jones: We should do a search on the code to make sure. I know of some places that use it.

d2l-my-courses.js
```diff
@@ -14,70 +14,50 @@ If it is off and the attribute is not added, the `d2l-my-courses-content-animate
...
 const $_documentContainer = document.createElement('template');
 
 $_documentContainer.innerHTML = `<dom-module id="d2l-my-courses">
 	<template strip-whitespace="">
 		<template is="dom-if" if="[[!updatedSortLogic]]">
-			<d2l-my-courses-content-animated
-				href="[[presentationUrl]]"
-				token="[[token]]"
+			<d2l-my-courses-legacy
+				class="[[class]]"
```
svanherk: What is `class` needed for?  Doesn't look like either had that before

d2l-my-courses.js
```diff
@@ -86,7 +66,40 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-my-courses">
...
+		presentationUrl: String,
+		// Feature flag (switch) for using the updated sort logic and related fetaures
+		updatedSortLogic: {
+			type: Boolean,
+			value: false
+		}
+	},
 	behaviors: [
-		D2L.MyCourses.MyCoursesBehavior
+		D2L.PolymerBehaviors.Siren.EntityBehavior
```
svanherk: Do we need this here?  Looks like we didn't convert this outer component to use the entity-store, so maybe we don't include any behaviors and we just pass the token through (along with all the other properties you've defined above)?

ciss1995: Yes, that's better. It was only used for token.

d2l-my-courses.js
ciss1995: I am not sure but it was used in the demo as well. https://github.com/Brightspace/d2l-my-courses-ui/blob/master/demo/d2l-my-courses/d2l-my-courses-demo.html#L44

svanherk: It _seems_ like it's only needed at the top level (https://github.com/Brightspace/lms/blob/12c78892e6aa0043792cbaf980857ae045bba2f2/platformTools/homepage/definition/Web/Desktop/JavaScript/ClassInitializer.js), and doesn't need to propagate down to the container and legacy wrappers

ciss1995: Oh ok! I removed it.

d2l-my-courses.js
```diff
@@ -86,7 +64,38 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-my-courses">
...
+		// Types of notifications to include in update count in course tile -- passed to animation tile
+		courseUpdatesConfig: Object,
+		// Callback for upload in image-selector
+		courseImageUploadCb: Function,
+		// URL to fetch promoted searches for tabs
+		promotedSearches: String,
+		// URL to fetch a user's settings (e.g. default tab to select)
+		userSettingsUrl: String,
+		// URL to fetch widget settings
+		presentationUrl: String,
```
svanherk: Is this actually passed in by the widget?  Seems like it might be created from the `user-settings-url` and doesn't need to be a property or passed in at this level

svanherk: Looks like it's already not being passed in, so I think we can just remove this property here

legacy/build/lang/ar.js
```diff
@@ -0,0 +1,100 @@
+/* eslint-disable quotes, indent */
```
svanherk: I think this will break the automation of getting any new translations, but that's actually good (we're cool with these not getting anything new and just being a snapshot).  Should still work on the real files the container uses

PR #746, opened 2019-05-16
US106986 Update All Courses to use the new sdk

ciss1995:


src/d2l-all-courses.js
```diff
@@ -715,51 +747,58 @@ Polymer({
...
 
 		this.$.sortDropdownMenu.querySelector('d2l-menu-item-radio[value=' + sortName + ']').selected = true;
-	},
-	_updateFilteredEnrollments: function(enrollments, append) {
-		var enrollmentEntities = enrollments.getSubEntitiesByClass(Classes.enrollments.enrollment);
+	}
 
-		if (this.updatedSortLogic) {
+	_updateFilteredEnrollments(enrollments, append) {
+		if (!enrollments._entity) {
```
ciss1995: Added this if statement because on the initial load of all courses, the first enrollmentsEntity is our previous Entity (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-all-courses.js#L540). We will be able to remove this if once the component that fires this event uses the new sdk.

src/d2l-all-courses.js
```diff
@@ -437,36 +451,39 @@ Polymer({
...
-
-				return this.sirenEntityStoreFetch(this._enrollmentsSearchUrl, this.token)
-					.then(function(enrollmentsEntity) {
-						if (!enrollmentsEntity || !enrollmentsEntity.entity) {
-							return Promise.resolve();
-						}
-						this._updateFilteredEnrollments(enrollmentsEntity.entity, true);
-					}.bind(this));
+			var lastResponseEntity = this.lastEnrollmentsSearchResponse;
+			if (!lastResponseEntity._entity) {
```
ciss1995: Similar to the previous comment.

src/d2l-all-courses.js
```diff
@@ -30,360 +28,377 @@ import './search-filter/d2l-filter-menu.js';
...
+
+	static get ready() {
 		this._filterText = this.localize('filtering.filter');
 		this._bustCacheToken = Math.random();
-	},
-	attached: function() {
+	}
+
+	attached() {
+		this.addEventListener('d2l-simple-overlay-opening', this._onSimpleOverlayOpening);
```
ciss1995: https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-all-courses.js#L354

ciss1995:
Could also clean up All-courses and remove everything that's sortLogic off

ciss1995:
We should also clean up all the sortLogic off code from `src/d2l-all-courses`.

src/d2l-all-courses.js
edwardg8: I think Maya recommended to have them in `ready`

src/d2l-all-courses.js
```diff
@@ -621,15 +644,24 @@ Polymer({
...
 		this.$['all-courses-scroll-threshold'].clearTriggers();
-	},
-	_myEnrollmentsEntityChanged: function(entity) {
+	}
+
+	_onEnrollmentCollectionEntityChange(entity) {
+		if (!entity) {
+			return Promise.resolve();
+		}
+		return this._updateFilteredEnrollments(entity, true);
```
edwardg8: it looks to me `_updateFilteredEnrollments` second parameter is always true when called in other places, maybe we can remove it?

edwardg8: I see that `_onSearchResultsChanged` still pass in false, nvm

src/d2l-all-courses.js
```diff
@@ -30,360 +28,377 @@ import './search-filter/d2l-filter-menu.js';
 import './search-filter/d2l-search-widget-custom.js';
 import './d2l-utility-behavior.js';
 import './localize-behavior.js';
-import '../legacy/tile-grid/d2l-all-courses-segregated-content.js'; // TODO: remove this dependency, since updated-sort-logic will always be true
+import '../legacy/tile-grid/d2l-all-courses-segregated-content.js';
```
svanherk: Can we put the comment back?

src/d2l-all-courses.js
svanherk: We should also probably be removing them in `detached`

edwardg8: you may want to use `connectedCallback` instead of attached

PR #747, opened 2019-05-21
DE34274: Hide the advanced search link with a dom-if since hidden wil…

njostonehouse:
…l be overridden by class styles.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -101,9 +101,11 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-all-courses">
 							</d2l-dropdown>
 						</div>
 					</div>
-					<div class="search-and-filter-row advanced-search-link" hidden$="[[!_showAdvancedSearchLink]]">
-						<d2l-link href$="[[advancedSearchUrl]]">{{localize('advancedSearch')}}</d2l-link>
-					</div>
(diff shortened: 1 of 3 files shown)


src/d2l-all-courses.js
```diff
@@ -101,9 +101,11 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-all-courses">
 							</d2l-dropdown>
 						</div>
 					</div>
-					<div class="search-and-filter-row advanced-search-link" hidden$="[[!_showAdvancedSearchLink]]">
```
njostonehouse: `hidden` combined with `class` styles ends up showing the link regardless of the `_showAdvancedSearchLink` value. Changing it to `dom-if` hides the advanced search link when it isn't specified.

njostonehouse:
Well that's a relevant test failure. Hmm.

src/d2l-all-courses.js
m6jones: We could use `.search-and-filter-row[hidden] { display: none; }` to make hidden work again

njostonehouse: Aha! I like that a lot better. I'll give it a shot.

edwardg8: that's indeed much cleaner change

PR #748, opened 2019-05-21
Functions used by monolith need to be accessible a level up (d2l-my-c…

ciss1995:
…ourses)

### DIFF
--- a/d2l-my-courses.js
+++ b/d2l-my-courses.js
@@ -93,13 +93,13 @@ Polymer({
 
 	courseImageUploadCompleted: function(success) {
 		return this.updatedSortLogic
-			? document.querySelector('d2l-my-courses').$$('d2l-my-courses-container').courseImageUploadCompleted(success)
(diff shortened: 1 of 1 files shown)


d2l-my-courses.js
```diff
@@ -89,5 +89,17 @@ Polymer({
 			value: false
 		},
 		token: String
+	},
+
+	courseImageUploadCompleted: function(success) {
+		return this.updatedSortLogic
+			? document.querySelector('d2l-my-courses').$$('d2l-my-courses-container').courseImageUploadCompleted(success)
```
svanherk: Can you use `this` instead of `document.querySelector('d2l-my-courses')`?  I _think_ that should work...

edwardg8: so we are copying the code over instead of moving the code?

ciss1995: We can go 1 layer down (d2l-my-courses-content) but I don't think we can completely move all the code up.

PR #750, opened 2019-05-23
DE34415 - Fix Widget Refresh After Change in Settings

svanherk:
Now that we're going through the entity store, the presentation url is no longer getting re-fetched if `my-courses` is reattached after the user updates the widget settings.  This will make the url unique each time.  It will result in a duplicate call of the same API, but only the first load (after that, the main url will be cached and not refetched each time, only the new url with the `bustCache` attribute).  We use this technique elsewhere in my-courses (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content-behavior.js#L578).

Not sure what this means for the SDK... will a similar technique need to be used, or can we tell the SDK entity that we think it has a stale resource?

### DIFF
--- a/src/d2l-my-courses-behavior.js
+++ b/src/d2l-my-courses-behavior.js
@@ -59,7 +59,11 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 		},
 		_tabSearchType: String,
 		_changedCourseEnrollment: Object,
-		_updateUserSettingsAction: Object
+		_updateUserSettingsAction: Object,
+		_bustPresentationCache: {
+			type: Boolean,
+			value: true
+		}
 	},
 	_computeShowGroupByTabs: function(groups) {
 		return groups.length >= 2 || (groups.length > 0 && !this._enrollmentsSearchAction);
@@ -68,6 +72,9 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 		'd2l-course-enrollment-change': '_onCourseEnrollmentChange',
 		'd2l-tab-changed': '_tabSelectedChanged'
 	},
+	observers: [
+		'_onPresentationUrlChange(presentationUrl)'
+	],
 	attached: function() {
 		if (!this.enrollmentsUrl || !this.userSettingsUrl) {
 			return;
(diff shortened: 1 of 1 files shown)


edwardg8:
I used similar technique for pinned tab too. What confuses me here is why do we only bust cache the first time the url gets changed, shouldn't this be every time?

svanherk:
If the component gets `detached` and a new version `attached`, the property gets cleared and set back to `true`.  I'm seeing that happen every time I update the widget settings, so it only has to happen once per component version and then I stop it to avoid infinite looping.

ctwomey1:
Depending on where the caching happens, you can force a fetch using the bypass cache param

https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-store.js#L142-L144

svanherk:
> Depending on where the caching happens, you can force a fetch using the bypass cache param

Yeah, the problem is the `fetch` is all done by the `entity-behavior`, I don't currently have any hooks to say "wait, don't load with this href yet" or "wait, don't use the cache" (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L62).

ctwomey1:
But we could have that hook easily enough and it'd be super nice.  @jopnick would be singing your praises for days.

Like if we added the same property you added in this pr to that behavior, and changed https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L62 to bypass cache based on the value of that variable.  Well i think that would solve the issue right?

svanherk:
It should... I'll play around and confirm.  I was trying to avoid changing that behavior in cert, but if I do it that way it should be safe and backwards compatible and not used by others, I'll just check to make sure I pick a property name not being used by any other behaviours or components

ctwomey1:
Ahh i'm sorry i totally missed that this was a cert fix.  We can totally merge this for what's in cert if it works well enough, and fix it proper like going forward

svanherk:
> Ahh i'm sorry i totally missed that this was a cert fix. We can totally merge this for what's in cert if it works well enough, and fix it proper like going forward

Yeah no problem, there is still a week left but I just found another my-courses cert issue in Edge (unrelated to this fix) so I think I'll lean towards getting this in to unlock testing and fix it the "right" way in master to give more time for `entity-behavior` issues to be found

svanherk:
Yeah going to go with the my-courses-only change for cert, and the better way in master

PR #751, opened 2019-05-27
US107060 Update my-courses-behavior to use new sdk

ciss1995:
Progress PR! 

Will need to fix tests.

### DIFF
--- a/src/d2l-my-courses-behavior.js
+++ b/src/d2l-my-courses-behavior.js
@@ -1,9 +1,10 @@
 import '@polymer/polymer/polymer-legacy.js';
-import { Rels } from 'd2l-hypermedia-constants';
-import { Actions } from 'd2l-hypermedia-constants';
 import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
 import './d2l-utility-behavior.js';
 import './localize-behavior.js';
+import { UserSettingsEntity } from 'siren-sdk/src/userSettings/UserSettingsEntity';
+import { PromotedSearchEntity } from 'siren-sdk/src/promotedSearch/PromotedSearchEntity.js';
+import { EnrollmentCollectionEntity } from 'siren-sdk/src/enrollments/EnrollmentCollectionEntity.js';
 window.D2L = window.D2L || {};
 window.D2L.MyCourses = window.D2L.MyCourses || {};
 
@@ -41,11 +42,6 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 		userSettingsUrl: String,
 		// URL to fetch widget settings
 		presentationUrl: String,
-		// Feature flag (switch) for using the updated sort logic and related fetaures
-		updatedSortLogic: {
-			type: Boolean,
-			value: false
-		},
 		currentTabId: String,
 		_enrollmentsSearchAction: Object,
 		_pinnedTabAction: Object,
@@ -59,7 +55,10 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 		},
 		_tabSearchType: String,
 		_changedCourseEnrollment: Object,
-		_updateUserSettingsAction: Object
+		_updateUserSettingsAction: Object,
+		_enrollmentCollectionEntity: Object,
+		_userSettingsEntity: Object,
+		_promotedSearch: Object,
 	},
 	_computeShowGroupByTabs: function(groups) {
 		return groups.length >= 2 || (groups.length > 0 && !this._enrollmentsSearchAction);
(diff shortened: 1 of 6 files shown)


src/d2l-my-courses-behavior.js
```diff
@@ -68,34 +67,67 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
 		'd2l-course-enrollment-change': '_onCourseEnrollmentChange',
 		'd2l-tab-changed': '_tabSelectedChanged'
 	},
+	observers: [
```
ciss1995: Another idea:
1/Add an observer that will be called when `EnrollmentCollectionEntity` and `UserSettingsEntity` will be set and execute the code after https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-behavior.js#L80. 
2/Also add an observer for https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-behavior.js#L122 which will be called when the entities have been fetched.

ciss1995: We could also observe the respective URLs and get the 3 entities at first.

src/d2l-my-courses-behavior.js
```diff
@@ -68,34 +67,67 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
+				this._pinnedTabAction = enrollmentsRootEntity.searchMyPinnedEnrollmentsAction();
+			}
+
+			if (userSettingsEntity.userSettingsHref()) {
+				this.presentationUrl = userSettingsEntity.userSettingsHref();
+			}
+
+			this._updateUserSettingsAction = userSettingsEntity.userSettingsAction();
+			this._fetchTabSearchActions();
+		}.bind(this), 600);
```
ciss1995: I am also not sure about the timing. In the demo page, when I used 400 or 500 it was working on and off.

ciss1995:
This seems to work with the demo page! Although I am not sure if it is the best implementation.

ciss1995:
Also, `My-Courses` seems to not be getting the new `siren-sdk` https://travis-ci.org/Brightspace/d2l-my-courses-ui/builds/537946035?utm_source=github_status&utm_medium=notification

src/d2l-my-courses-behavior.js
```diff
@@ -68,34 +67,67 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
+			this._userSettingsEntity = entity;
+		}
+
+		if (entity instanceof PromotedSearchEntity) {
+			this._promotedSearchEntity = entity;
+		}
+	},
+	_setEnrollmentCollectionEntity: function(url) {
+		this._setEntityType(EnrollmentCollectionEntity);
+		this.href = url;
```
m6jones: This will overwrite `this.href` from the other methods below. 

I see what you are doing it might be better to not use the mixin and use the `entityFactory` directly.

src/d2l-my-courses-behavior.js
```diff
@@ -142,67 +169,65 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
-			this.sirenEntityStoreFetch(this.promotedSearches, this.token),
-			this.sirenEntityStoreFetch(this.userSettingsUrl, this.token)
-		]).then(function(values) {
-			var promotedSearchesEntity = values[0] && values[0].entity;
-			var userSettingsEntity = values[1] && values[1].entity;
+		return new Promise(function() {
+			this._setPromotedSearchEntity(this.promotedSearches);
 
-			this._tabSearchActions = [];
+			setTimeout(function() {
```
m6jones: What are we waiting for here and do you think we can some how know when it is done?

src/d2l-my-courses-behavior.js
ciss1995: I am waiting for `_enrollmentCollectionEntity` and `_userSettingsEntity` to be set

src/d2l-my-courses-behavior.js
ciss1995: Yup so each of them overwrite the other to create a new entity. 
Oh ok. I have never tried with `entityFactory`.

src/d2l-my-courses-behavior.js
ciss1995: We are waiting for promotedSearchEntity to be set, similarly to the `setTimeout` above.

src/d2l-my-courses-behavior.js
m6jones: Is there a way you can add a promise resolve in the callbacks? So it will notify you when they are both good to go?

src/d2l-my-courses-behavior.js
m6jones: The mixin is for elements that have one entity. Skip it if you have multiple and go to the source. (Just remember clean up)

src/d2l-my-courses-behavior.js
ciss1995: I can try to add one.

src/d2l-my-courses-behavior.js
ciss1995: Ok I will try it out.

src/d2l-my-courses-behavior.js
svanherk: Yeah, using a `setTimeout` in prod would worry me.  We cant test this with every connection speed and browser, and it's sure to fail some combinations

ciss1995: Yeah, it was very inconsistent when I was testing it as well. Also why I wasn't too confident about this approach.

ciss1995:
Waiting on enrollment to release so my-courses can pick up the new sdk.

src/d2l-my-courses-behavior.js
```diff
@@ -73,105 +73,55 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
 			}
 
-			var lastEnrollmentsSearchName = userSettingsEntity
-						&& userSettingsEntity.properties
-						&& userSettingsEntity.properties.MostRecentEnrollmentsSearchName;
+			var lastEnrollmentsSearchName = userSettingsEntity.mostRecentEnrollmentsSearchName();
 
-			if (promotedSearchesEntity.actions.length > 1) {
-				this._tabSearchActions = promotedSearchesEntity.actions.map(function(action) {
+			if (promotedSearchesEntity.actions().length > 1) {
```
edwardg8: maybe we can move the logic to the Entity, just make it return the mapped actions?

edwardg8: cross it out, leave it like this.

src/d2l-my-courses-behavior.js
```diff
@@ -73,29 +73,106 @@ D2L.MyCourses.MyCoursesBehaviorImpl = {
...
+
+		if (userSettingsEntity.userSettingsHref()) {
+			this.presentationUrl = userSettingsEntity.userSettingsHref();
+		}
+
+		this._updateUserSettingsAction = userSettingsEntity.userSettingsAction();
+
+		this._fetchTabSearchActions();
+	},
+	_onPromotedSeachEntityChange: function() {
```
svanherk: Minor note I just noticed- there's a typo here, missing the `r` in `Search`

PR #757, opened 2019-06-10
US107059 Remove Event Listeners

ciss1995:


src/d2l-my-courses-content-behavior.js
```diff
@@ -810,6 +810,39 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 				'value': id
 			}
 		]);
+	},
+	_dateTextAndStatus: function(isCompletionDate, date) {
```
ciss1995: Duplicate of https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/components/date-text-status-mixin.js#L13. 
I added this in `d2l-my-courses-content` to remove `redefine computed property` warnings. 
We could also remove https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content-behavior.js#L822 and import `DateTextAndStatus` from `d2l-enrollments` to remove the warnings.

src/card-grid/d2l-my-courses-content.js
```diff
@@ -24,11 +24,24 @@ import 'd2l-typography/d2l-typography-shared-styles.js';
...
-	<template strip-whitespace="">
+/**
+ * @customElement
+ * @polymer
+ */
+class MyCoursesContent extends mixinBehaviors([
+	D2L.MyCourses.MyCoursesContentBehavior,
+	D2L.MyCourses.CardGridBehavior
+], EntityMixin(PolymerElement)) {
+	constructor() {
```
m6jones: EntityMixen without saying what entity we are using.

src/d2l-my-courses-content-behavior.js
```diff
@@ -261,25 +263,23 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
-		) {
+	_onEnrollmentCardFetched: function(url, enrollmentCollectionEntity) {
+		if (!url || !enrollmentCollectionEntity) {
 			return;
 		}
-
-		var orgUnitId = this._getOrgUnitIdFromHref(e.detail.organizationUrl);
-
-		this._orgUnitIdMap[orgUnitId] = e.detail.enrollmentUrl;
+		enrollmentCollectionEntity.onEnrollmentEntityChange(url, (enrollmentEntity) => {
```
m6jones: I forgot what problem are we solving by passing in the `url` here?

src/d2l-my-courses-content-behavior.js
```diff
@@ -261,25 +263,23 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+			var orgUnitId = this._getOrgUnitIdFromHref(enrollmentEntity.organizationHref());
+			this._orgUnitIdMap[orgUnitId] = url;
+		});
 	},
-	_onD2lEnrollmentCardStatus: function(e) {
-		if (!e.detail || !e.detail.status || !e.detail.enrollmentUrl || e.detail.status.completed) {
+	_onEnrollmentCardStatus: function(enrollmentCardStatusDetails) {
```
m6jones: Note, that closed courses don't need to run through this method. So in the future we can remove this method completely :D

src/card-grid/d2l-my-courses-content.js
ciss1995: Since we are using `entityFactory` to get the entities, I did not think that we needed to set a general `entityType`.

src/d2l-my-courses-content-behavior.js
ciss1995: I believe that we want to get the EnrollmentEntity for that url.
https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/enrollments/EnrollmentCollectionEntity.js#L49

src/card-grid/d2l-my-courses-content.js
m6jones: That is true. It also means we don't need to extend `EntityMixen`

src/d2l-my-courses-content-behavior.js
ciss1995: If this method is currently not needed we can remove it. It was triggered by an event listener, from an event fired in `d2l-enrollments`. That's why I kept it.

src/card-grid/d2l-my-courses-content.js
ciss1995: Yes totally! I will remove `EntityMixen`

src/d2l-my-courses-content-behavior.js
```diff
@@ -261,25 +263,23 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+		if (!url || !enrollmentCollectionEntity) {
 			return;
 		}
-
-		var orgUnitId = this._getOrgUnitIdFromHref(e.detail.organizationUrl);
-
-		this._orgUnitIdMap[orgUnitId] = e.detail.enrollmentUrl;
+		enrollmentCollectionEntity.onEnrollmentEntityChange(url, (enrollmentEntity) => {
+			var orgUnitId = this._getOrgUnitIdFromHref(enrollmentEntity.organizationHref());
+			this._orgUnitIdMap[orgUnitId] = url;
```
edwardg8: just wonder if `enrollmentEntity` can have this `orgUnitId` property?

src/d2l-my-courses-content-behavior.js
edwardg8: Instead of having this computed property, why don't we just call it when `_enrollmentCardStatusDetails ` changes like here https://github.com/Brightspace/d2l-my-courses-ui/blob/b148b4b4f8e0ce3aed7e33e2cd62fabb3323cfd3/src/d2l-my-courses-content-behavior.js#L302

src/d2l-my-courses-content-behavior.js
m6jones: I like that.

src/d2l-my-courses-content-behavior.js
```diff
@@ -291,6 +291,31 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 
 		this._onResize();
 	},
+	_onEnrollmentCardStatusChange: function(url, enrollmentCollectionEntity) {
```
edwardg8: I would recommend to change the function names to `loadxxx` instead of  `onChange` to distinguish from event listeners.
I see `_onEnrollmentCardStatusChange` and `_onEnrollmentCardFetched`

src/d2l-my-courses-content-behavior.js
```diff
@@ -362,33 +387,11 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
 		var isPinned = e.detail.isPinned;
-		var orgUnitId;
-		if (e.detail.orgUnitId) {
-			orgUnitId = e.detail.orgUnitId;
-		} else if (e.detail.organization) {
-			orgUnitId = this._getOrgUnitIdFromHref(this.getEntityIdentifier(this.parseEntity(e.detail.organization)));
-		} else if (e.detail.enrollment && e.detail.enrollment.hasLinkByRel(Rels.organization)) {
-			orgUnitId = this._getOrgUnitIdFromHref(e.detail.enrollment.getLinkByRel(Rels.organization).href);
-		}
+		var orgUnitId = this._getOrgUnitIdFromHref(e.detail.enrollment.organizationHref());
```
edwardg8: If we we can get `orgUnitId` from entity, we can get rid of `_getOrgUnitIdFromHref` in this behavior.

src/d2l-my-courses-content-behavior.js
```diff
@@ -291,6 +291,31 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
 	},
+	_onEnrollmentCardStatusChange: function(url, enrollmentCollectionEntity) {
+		if (!url || !enrollmentCollectionEntity) {
+			return;
+		}
+
+		enrollmentCollectionEntity.onEnrollmentEntityChange(url, (enrollmentEntity) => {
+			enrollmentEntity.onUserActivityUsageChange((userActivityUsage) => {
+				const dateTextAndStatus = this._dateTextAndStatus(userActivityUsage.isCompletionDate(), userActivityUsage.date());
+				this._enrollmentCardStatusDetails = {
```
edwardg8: I don't see why we need this property here `_enrollmentCardStatusDetails`? we can just call `_onEnrollmentCardStatus` here?

src/d2l-my-courses-content-behavior.js
edwardg8: You can just add the mixin without removing the localizer, and that's the purpose of the mixin, Without `localizer`, we wouldn't be able to localize like here https://github.com/Brightspace/d2l-my-courses-ui/blob/b148b4b4f8e0ce3aed7e33e2cd62fabb3323cfd3/src/d2l-my-courses-content-behavior.js#L524 though.

src/d2l-my-courses-content-behavior.js
ciss1995: Yeah I guess we could

src/d2l-my-courses-content-behavior.js
ciss1995: Just curious, why the 3 `x`? Do you mean changing the name to e.g: `_onEnrollmentCardFetchedLoadxxx`?

edwardg8: i just mean like `loadEnrollmentCardInfo` for example, but depends on what the function does, change it to reflect that accordingly. Because the only thing the function does now is `load` info, they are no longer event listeners.

src/d2l-my-courses-content-behavior.js
ciss1995: I don't think that we can't get rid of `_getOrgUnitIdFromHref` because it is being used in other parts of the code.

edwardg8: that's sad

src/d2l-my-courses-content-behavior.js
ciss1995: If I add the mixin without removing the localizer, I am getting the warnings that I mentioned above. I just copied parts of the function that I needed in `d2l-my-courses`.

src/d2l-my-courses-content-behavior.js
```diff
@@ -257,29 +257,23 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
-	/*
-	* Listeners
-	*/
-	_onD2lEnrollmentCardFetched: function(e) {
-		if (dom(e).rootTarget === this
-			|| !e.detail.organizationUrl
-			|| !e.detail.enrollmentUrl
-		) {
+	_loadEnrollmentCard: function(url, enrollmentCollectionEntity) {
+		if (!url || !enrollmentCollectionEntity) {
```
edwardg8: From here it looks like it's taking `enrollmentsEntity` not `collection`.  https://github.com/Brightspace/d2l-my-courses-ui/blob/730fbfc1f9a276829d6bc419396de3c50c9ad81c/src/d2l-my-courses-content-behavior.js#L767
Maybe we can just move this to `_populateEnrollments` or rename it to `_insertToOrgUnitIdMap` since it's not loading anything.

m6jones: Good idea

src/d2l-my-courses-content-behavior.js
```diff
@@ -257,29 +257,23 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
-
-		this._orgUnitIdMap[orgUnitId] = e.detail.enrollmentUrl;
+		enrollmentCollectionEntity.onEnrollmentEntityChange(url, (enrollmentEntity) => {
+			var orgUnitId = this._getOrgUnitIdFromHref(enrollmentEntity.organizationHref());
+			this._orgUnitIdMap[orgUnitId] = url;
+		});
 	},
-	_onD2lEnrollmentCardStatus: function(e) {
-		if (!e.detail || !e.detail.status || !e.detail.enrollmentUrl || e.detail.status.completed) {
+	_loadEnrollmentCardStatus: function(enrollmentCardStatusDetails) {
```
edwardg8: sorry, I feel `setEnrollmentCardStatus` is more appropriate here.

src/d2l-my-courses-content-behavior.js
```diff
@@ -291,6 +285,37 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 
 		this._onResize();
 	},
+	_fetchEnrollmentCardStatus: function(url, enrollmentCollectionEntity) {
```
edwardg8: Same here, the caller is passing in `enrollmentsEntity` instead of `enrollmentCollectionEntity`, make sure they are consistent to avoid confusion

src/d2l-my-courses-container.js
```diff
@@ -33,7 +33,7 @@ class MyCoursesContainer extends mixinBehaviors([
 						<!-- item.name is an OrgUnitId, and querySelector does not work with components with ids that start with a number -->
 						<d2l-tab-panel id="panel-[[item.name]]" text="[[item.title]]" selected="[[item.selected]]">
 							<d2l-my-courses-content
-								href="[[presentationUrl]]"
```
svanherk: Do we need this change below as well? https://github.com/Brightspace/d2l-my-courses-ui/pull/757/files#diff-79c7c139ccd6d5d799b4998ccaba34a9R57

src/d2l-my-courses-content-behavior.js
```diff
@@ -824,5 +860,6 @@ D2L.MyCourses.MyCoursesContentBehavior = [
 	D2L.MyCourses.UtilityBehavior,
 	D2L.PolymerBehaviors.Siren.EntityBehavior,
 	D2L.PolymerBehaviors.Siren.SirenActionBehavior,
-	D2L.MyCourses.MyCoursesContentBehaviorImpl
+	D2L.MyCourses.MyCoursesContentBehaviorImpl,
```
svanherk: Dangling comma here

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -931,22 +868,105 @@ describe('d2l-my-courses-content', () => {
...
+			});
+		});
+
+		it('Should call _loadEnrollmentCardStatus', () => {
+			var spy = sandbox.spy(component, '_loadEnrollmentCardStatus');
+			component._fetchEnrollmentCardStatus(1, _enrollmentCollectionEntity);
+			expect(spy).to.have.been.calledTwice;
+		});
+
+		it('Should get _org proprely', () => {
```
svanherk: Minor: properly

src/d2l-my-courses-content-behavior.js
```diff
@@ -292,7 +292,7 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
 
 		enrollmentCollectionEntity.onEnrollmentEntityChange(url, (enrollmentEntity) => {
 			enrollmentEntity.onUserActivityUsageChange((userActivityUsage) => {
-				const dateTextAndStatus = this._dateTextAndStatus(userActivityUsage.isCompletionDate(), userActivityUsage.date());
+				const dateTextAndStatus = this.enrollmentStatus(userActivityUsage.isCompletionDate(), userActivityUsage.date());
```
edwardg8: this shouldn't be `dateTextAndStatus` any more?

PR #763, opened 2019-07-31
DE35341 - Simplify the promise chain

svanherk:
Simplify the promise chain down to the functions that actual need it, removing the expectation that `EntityFactory` returns a promise.


edwardg8:
Sorry I just made the same change, and waited for the build to test out, and distracted by the talk, should have pushed it. Now it breaks all the tests for populateEnrollments.

svanherk:
Yup, working through those now

PR #765, opened 2019-08-09
DE35094-this fixed double load problems

edwardg8:


edwardg8:
This should fix the double load problem, better approach is definitely fixing it on server side. However this will "break" the UI because we no longer show spinning icon while waiting response from fetching tabs, will revisit this problem next week.

edwardg8:
Added load spinner, this should give exact the same look and feel as before

edwardg8:
The Grid view is very weird...

PR #766, opened 2019-08-29
Update to use requestIdleCallback shim from core.

dbatiste:


svanherk:
@dbatiste How do you feel about closing this?  It's in the `legacy` codeline which we haven't been allowing any updates too.  It'll eventually be deleted once the migration effort is done, but is probably months away

svanherk:
But if you want to get it in, it _seems_ like a fairly risk-free change.

dbatiste:
Holy smokes.  This is old.  Yeah this should be low risk, but it's not that important to get in either.  I'm fine with closing this - definitely don't want anyone copying the original though.

PR #767, opened 2019-09-24
DE36168-remove inactive courses alert message

edwardg8:


svanherk:
Do we need to deal with this? https://github.com/Brightspace/d2l-my-courses-ui/blob/master/legacy/tile-grid/d2l-my-courses-content-animated.js#L169

test/d2l-filter-menu/d2l-filter-list-item-role.js
```diff
@@ -1,19 +0,0 @@
-describe('d2l-filter-list-item-role', function() {
-	var listItem;
-
-	beforeEach(function() {
-		listItem = fixture('d2l-filter-list-item-role-fixture');
-	});
-
-	it('should show the unchecked icon when the item is not selected', function() {
```
edwardg8: These are exactly same tests as in `d2l-filter-list-item`

test/legacy/d2l-filter-menu-legacy/d2l-filter-list-item-role-legacy.js
```diff
@@ -1,19 +0,0 @@
-describe('d2l-filter-list-item-role-legacy', function() {
-	var listItem;
-
-	beforeEach(function() {
-		listItem = fixture('d2l-filter-list-item-role-legacy-fixture');
-	});
-
-	it('should show the unchecked icon when the item is not selected', function() {
```
edwardg8: These are exactly same tests as in `d2l-filter-list-item-legacy`

test/legacy/d2l-filter-menu-legacy/d2l-filter-list-item-role-legacy.html
```diff
@@ -1,20 +0,0 @@
-<!doctype html>
```
svanherk: I'm thinking we do want to keep these tests, just because once I update the filter I'll only be doing that in SortLogicOn, not SortLogicOff (which is what these are testing)

test/index.all.html
```diff
@@ -15,7 +15,6 @@
 				'./d2l-filter-menu/d2l-filter-menu-tab-roles.html',
 				'./d2l-filter-menu/d2l-filter-menu-tab.html',
 				'./d2l-filter-menu/d2l-filter-list-item.html',
-				'./d2l-filter-menu/d2l-filter-list-item-role.html',
```
svanherk: I think the build is still mad because it wants you to delete from this from https://github.com/Brightspace/d2l-my-courses-ui/blob/master/test/index.high-priority.html#L18 too?

test/legacy/d2l-filter-menu-legacy/d2l-filter-list-item-role-legacy.html
edwardg8: It's actually duplicated in both `UpdatedSortLogic` On and Off, the exact same tests were written 4 times.

svanherk:
These latest test failures (`Unable to get property 'call' of undefined or null reference` in IE11 and Edge) are because of this: Polymer/tools#727
The fix is merged in but currently not released: https://github.com/Polymer/tools/blob/master/packages/build/CHANGELOG.md#change-log
Once this happens, we can re-enable Edge testing in the repos we've had to disable it in, including core and anything using d2l-icons.

test/legacy/d2l-filter-menu-legacy/d2l-filter-list-item-role-legacy.html
svanherk: Yeah but isn't this testing the file in `legacy`, and the other one is testing the version in `src` (which will be changing soon?

svanherk: Ah nvm, I get what's you're saying.  You left the  `d2l-filter-list-item` and `d2l-filter-list-item-legacy` tests, and just removed the role ones

edwardg8: yea, they are the exactly same for no reasons.

PR #769, opened 2020-01-16
Remove unused code from the UpdatedSortLogic ON workflow

svanherk:
While doing the my-courses caching spike I cleaned up some of this unused code, since UpdatedSortLogic will always be true for these components.  This should wait to merge with other my-courses work, but putting up this draft PR now because my computer is having issues and I don't want to lose it!


svanherk:
Moved to https://github.com/Brightspace/d2l-my-courses-ui/pull/771

PR #771, opened 2020-01-31
US113258 - Remove UpdatedSortLogic Off Code from the non-legacy path

svanherk:
This is the first of two PRs to get all two-way binding cleaned up.  I started this cleanup before losing my laptop and there will be lots of merge conflicts after removing two-way binding.  Since we need to do a regression test anyways, Mark is good with merging this at the same time.

This PR should cause no functional changes - since we split the code to `legacy` and `src` (latest), these components can always expect that `UpdatedSortLogic` will be true.  Here I'm removing code that only runs when it is `false`, and attributes only needed by that code path as well.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -2,8 +2,8 @@
 `d2l-all-courses`
 Polymer-based web component for all courses.
 
-TODO: This component is only rendered if the `d2l.Tools.MyCoursesWidget.UpdatedSortLogic` config variable is on, meaning `updated-sort-logic` is true.
-We can do a lot of cleanup here around `d2l-all-courses-segregated-content` (no longer needed).  Only the `d2l-all-courses-unified-content` component is rendered.
+This is only used if the `d2l.Tools.MyCoursesWidget.UpdatedSortLogic` config variable is on
+(meaning the `updated-sort-logic` attribute was added to the `d2l-my-courses` component).
 
 */
 import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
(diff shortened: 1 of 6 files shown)


test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -81,32 +55,6 @@ describe('d2l-all-courses', function() {
 		});
 	});
 
-	describe('getCourseTileItemCount', function() {
```
svanherk: This is testing `UpdatedSortLogic` OFF logic, and still exist in the legacy tests

src/d2l-all-courses.js
```diff
@@ -398,17 +348,6 @@ class AllCourses extends mixinBehaviors([
 	load() {
 		this.$['all-courses-scroll-threshold'].scrollTarget = this.$['all-courses'].scrollRegion;
 		this.$['all-courses-scroll-threshold'].clearTriggers();
-		if (!this.updatedSortLogic) {
-			this._filteredPinnedEnrollments.forEach(function(item) {
```
edwardg8: don't we want to keep the code since it's related to updatedSortLogic = off?

edwardg8: OH I see, you are removing the logics of 'updatedSortLogic = off' in new my courses, not the legacy. Make sense

svanherk: It still exists in `d2l-all-courses-legacy` (https://github.com/Brightspace/d2l-my-courses-ui/blob/b4cca7d3b6de6e256844e3c8772686cdb506a1ad/legacy/d2l-all-courses-legacy.js#L388), which is the component that gets loaded if `UpdatedSortLogic` is set to OFF.  Since we split the code, this component always had `UpdatedSortLogic` ON

edwardg8: Reading the title of the PR, I thought you are cleaning up in legacy code for some reason, my bad.

svanherk: You're right, it's not very clear.  I'll update!

src/d2l-all-courses.js
```diff
@@ -746,16 +666,8 @@ class AllCourses extends mixinBehaviors([
 		return prefix + this._bustCacheToken + suffix;
 	}
 
-	_clearFilteredCourses() {
-		if (!this.updatedSortLogic) {
-			this._pinnedCoursesMap = {};
-			this._unpinnedCoursesMap = {};
```
edwardg8: I can't remember what `_pinnedCoursesMap ` and `_unpinnedCoursesMap ` for, and why they are no longer needed?

svanherk: I believe this was specifically to fix the animation in shadow DOM with `updatedSortLogic`  OFF,  These variables were only ever used in `!this.updatedSortLogic` blocks

PR #777, opened 2020-03-24
Fix some tests that were broken after some Lit migrations

svanherk:
I'm starting a big refactor in this area, but this is just getting the tests back to green for now.  Future changes will refactor the tests themselves as needed

### DIFF
--- a/test/legacy/d2l-course-tile/d2l-course-tile.js
+++ b/test/legacy/d2l-course-tile/d2l-course-tile.js
@@ -121,7 +121,7 @@ describe('<d2l-course-tile>', function() {
 		semesterOrganizationEntity = window.D2L.Hypermedia.Siren.Parse(semesterOrganization);
 	});
 
-	beforeEach(function() {
+	beforeEach(function(done) {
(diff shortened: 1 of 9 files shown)


test/legacy/d2l-course-tile/d2l-course-tile.js
```diff
@@ -133,6 +133,10 @@ describe('<d2l-course-tile>', function() {
 				ok: true,
 				json: function() { return Promise.resolve(organization); }
 			}));
+
+		requestAnimationFrame(function() {
+			done();
+		});
```
edwardg8: weird that we need some sort of random delay here?

edwardg8:
Looks awesome

test/legacy/d2l-course-tile/d2l-course-tile.js
svanherk: Yeah, the `d2l-dropdown` moving to `lit` seems to have caused some weird timing changes around `sandbox`-ing

PR #779, opened 2020-03-25
Update eslint config version

svanherk:
- Updates the eslint config version
- Fixes the issues that came from that (mostly just re-ordering)
- Move the legacy tests to their own folder to make it easier to delete later, and so that we can have different eslint configs for the different code branches once part of this repo starts using the lit config


src/d2l-all-courses.js
```diff
@@ -314,11 +183,137 @@ class AllCourses extends mixinBehaviors([
 		];
 	}
 
-	static get is() { return 'd2l-all-courses'; }
+	static get template() {
+		return html`
+			<style include="d2l-all-courses-styles"></style>
 
-	static get ready() {
```
svanherk: This file is just re-ordering, and deleting this function here.  It isn't in the right format for a Polymer lifecycle method, so it never actually runs and both of these variables are set in the function that runs whenever the simple-overlay opens.

PR #780, opened 2020-03-26
Restructure the All Courses section

svanherk:
Combining things that don't need to be separated now that we have split the code paths.

- Renamed `d2l-all-courses-unified-content` to just `d2l-all-courses-content`
- Put All Courses overlay styles back into the component


legacy/d2l-all-courses-legacy.js
```diff
@@ -3,7 +3,7 @@
 Polymer-based web component for all courses.
 
 TODO: This component is only rendered if the `d2l.Tools.MyCoursesWidget.UpdatedSortLogic` config variable is off, meaning `updated-sort-logic` is false.
```
svanherk: Will be doing this next, we definitely don't want `legacy` having to refer back to `src` at all, so the changes below are temporary

PR #781, opened 2020-03-27
US114859 - Remove src component from legacy

svanherk:
I originally had _a lot_ more cleanup here, but then backtracked since it was a bit of an unnecessary risk and we can hopefully just delete it all soon.  Instead I just deleted the `d2l-all-courses-content`, `d2l-tabs` and it's event listeners, and some unused properties. Everything else around the `updatedSortLogic` property, `groupBy` logic, and remaining tabs logic could be deleted but it made quite a messy PR to review.


legacy/d2l-my-courses-content-behavior-legacy.js
```diff
@@ -635,38 +596,6 @@ D2L.MyCourses.MyCoursesContentBehaviorLegacyImpl = {
 
 		return enrollmentsLength > 0 ? viewAllCourses + ' (' + count + ')' : viewAllCourses;
 	},
-	_openAllCoursesView: function(e) {
```
AllanKerr: Is there a reason you chose to remove this one specifically? [It looks like `d2l-my-courses-content-animated` overrides a bunch of methods that could all be removed.](https://github.com/Brightspace/d2l-my-courses-ui/blob/2e610709303a18f6764cfe59cd1d23b66f83b436/legacy/tile-grid/d2l-my-courses-content-animated.js#L149) I'm fine with only removing one, removing none, or removing them all. I'm just a bit curious about why this one in particular.

svanherk: I chose this one because it was accessing all the properties I'm removing in this file and in `d2l-all-courses-legacy.js`.  I could just leave `d2l-my-courses-content-behavior-legacy.js` as is but it seemed nice to cleanup _at least_ this while I'm there.

You're right though, there's SO much that can be cleaned up and I started doing it and then backtracked because we aren't really supposed to be touching this coed path at all if we can help it

legacy/d2l-all-courses-legacy.js
```diff
@@ -109,67 +105,16 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-all-courses-legacy">
 						[[item.alertMessage]]
 					</d2l-alert>
 				</template>
-
-				<template is="dom-if" if="[[updatedSortLogic]]">
```
edwardg8: So good to see these go away!!

PR #782, opened 2020-03-31
US114859 - Remove d2l-my-courses-behavior

svanherk:
Adding the functionality of `d2l-my-courses-behavior` back into `d2l-my-courses-container`.  Now that the code is split, there's no need to have this as a behavior.  Everything should be the same except where noted.


src/d2l-my-courses-behavior.js
```diff
@@ -1,227 +0,0 @@
...
-		_enrollmentCollectionEntity: Object,
-		_userSettingsEntity: Object,
-		_promotedSearch: Object,
-		// Hides loading spinner and shows tabs when true
-		_showContent: {
-			type: Boolean,
-			value: false
-		}
-	},
-	listeners: {
```
svanherk: Moved these to the `constructor`

src/d2l-my-courses-behavior.js
```diff
@@ -1,227 +0,0 @@
...
-		return this._setPromotedSearchEntity(this.promotedSearches);
-	},
-};
-
-/*
-* @polymerBehavior D2L.MyCourses.MyCoursesBehavior
-*/
-D2L.MyCourses.MyCoursesBehavior = [
-	D2L.PolymerBehaviors.MyCourses.LocalizeBehavior,
-	D2L.PolymerBehaviors.Siren.EntityBehavior,
```
svanherk: This wasn't actually being used anymore, removed

test/d2l-my-courses-container/d2l-my-courses-container.js
```diff
@@ -123,11 +123,6 @@ describe('d2l-my-courses', () => {
 		sandbox.restore();
 	});
 
-	it('should properly implement the d2l-my-courses-behavior', () => {
```
svanherk: Deleted since this is no longer coming from the behaviour, doesn't really make sense to test it

test/d2l-my-courses-container/d2l-my-courses-container.js
```diff
@@ -170,26 +165,6 @@ describe('d2l-my-courses', () => {
 		expect(component._tabSearchActions[2].selected).to.be.true;
 	});
 
-	describe('Listener setup', () => {
```
svanherk: Deleted since the two tests that follow this one are testing that an event raised trigger the listener and update data accordingly.  This test was _just_ testing the hookup.

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -189,10 +189,6 @@ describe('d2l-my-courses-content', () => {
 		sandbox.restore();
 	});
 
-	it('should properly implement d2l-my-courses-behavior', () => {
```
svanherk: Moved to the other test, `d2l-my-courses-content` doesn't import `d2l-my-courses-behavior`

src/d2l-my-courses-container.js
```diff
@@ -82,6 +133,160 @@ class MyCoursesContainer extends mixinBehaviors([
 
 	constructor() {
 		super();
+		this.addEventListener('d2l-course-enrollment-change', this._onCourseEnrollmentChange);
+		this.addEventListener('d2l-tab-changed', this._tabSelectedChanged);
```
AllanKerr: Don't you need to bind `this` for these event listeners?

svanherk: From what I can tell, seems this is only necessary for child elements? https://polymer-library.polymer-project.org/3.0/docs/devguide/events#listener-on-the-custom-element

AllanKerr: Looks like that was my mistake, it should automatically bind whatever `addEventListener` is called on 😅

PR #784, opened 2020-04-03
US114859 - Remove d2l-my-courses-content-behavior

svanherk:
Similar to the `d2l-my-courses-behavior`, `d2l-my-courses-content-behavior` is only used in one place since we split the code.  Combining back into one file here and doing some event listener, imports and properties cleanup.  I left some todo comment for future PRs


src/d2l-my-courses-container.js
```diff
@@ -98,7 +98,6 @@ class MyCoursesContainer extends mixinBehaviors([
 							<d2l-my-courses-content
 								presentation-url="[[_presentationUrl]]"
 								token="[[token]]"
-								disable-entity-cache
```
svanherk, line 101: This only makes sense if `d2l-my-courses-content` is using the old Polymer `entity-behaviour` (https://github.com/Brightspace/polymer-siren-behaviors/blob/master/store/entity-behavior.js#L43).  It hasn't been since we moved to the siren-sdk, which means things are broken.  I made a defect here and will fix in an upcoming PR: https://rally1.rallydev.com/#/15545167705ud/detail/defect/379581808432?fdp=true

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -322,31 +322,6 @@ describe('d2l-my-courses-content', () => {
 
 	});
 
-	describe('Listener setup', () => {
```
svanherk, line 325: We are no longer using the `listeners` block so we can't stub out the handlers.  I've added tests to test the functionality of the `clear-image-scroll-threshold` and `course-tile-organization` events, and then the other four already had tests.

src/d2l-my-courses-content.js
```diff
@@ -1,174 +1,342 @@
-import '@polymer/polymer/polymer-legacy.js';
```
svanherk, line 1: This file is a lot easier to look at if you hide whitespace

src/d2l-my-courses-content.js
```diff
@@ -1,174 +1,342 @@
...
-		},
-		_hasEnrollmentsChanged: {
-			type: Boolean,
-			value: false
-		},
-		_rootTabSelected: {
-			type: Boolean,
-			value: false
-		},
-		_enrollmentDate: {
```
svanherk, line 151: This wasn't used for anything

src/d2l-my-courses-content.js
```diff
@@ -179,51 +347,38 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
-	getLastOrgUnitId: function() {
+	}
+	getLastOrgUnitId() {
 		if (!this._setImageOrg.links) {
 			return;
 		}
 		return this._getOrgUnitIdFromHref(this.getEntityIdentifier(this._setImageOrg));
-	},
-
-	_allCoursesCreated: false,
```
svanherk, line 219: Moved these up into the properties

src/d2l-my-courses-content.js
```diff
@@ -583,40 +742,40 @@ D2L.MyCourses.MyCoursesContentBehaviorImpl = {
...
+	}
+	_fetchEnrollments() {
 		this.performanceMark('d2l.my-courses.root-enrollments.response');
 		this.performanceMeasure(
 			'd2l.my-courses.root-enrollments',
 			'd2l.my-courses.root-enrollments.request',
 			'd2l.my-courses.root-enrollments.response'
 		);
 
-		this._enrollmentsSearchUrl = this._createFetchEnrollmentsUrl();
```
svanherk, line 614: This didn't seem to be ever access anywhere than here and in the function below, where's it's immediately used.  Changed to just a local variable instead

PR #786, opened 2020-04-06
US114859 - Remove old d2l-polymer-behaviors

svanherk:
- Use the method from `core` instead
- Fix some Firefox test flakiness (unrelated)

### DIFF
--- a/test/d2l-my-courses-content/d2l-my-courses-content.js
+++ b/test/d2l-my-courses-content/d2l-my-courses-content.js
@@ -1,4 +1,5 @@
 import { EnrollmentCollectionEntity } from 'siren-sdk/src/enrollments/EnrollmentCollectionEntity.js';
+import { flush } from '@polymer/polymer/lib/utils/render-status.js';
 
 describe('d2l-my-courses-content', () => {
 	var sandbox,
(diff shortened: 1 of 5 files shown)


test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -603,25 +616,26 @@ describe('d2l-my-courses-content', () => {
...
-			setTimeout(() => {
+			component.addEventListener('course-image-loaded', () => {
 				expect(stub).to.have.been.calledWith(
 					'd2l.my-courses',
 					'd2l.my-courses.attached',
 					'd2l.my-courses.visible-images-complete'
 				);
 				done();
 			});
+			component.addEventListener('initially-visible-course-tile', () => {
```
AllanKerr, line 631: Shouldn't we still be removing this event listener to prevent it from interfering with other test cases?

svanherk: So I thought the wct test-fixture handled that for us, and we don't seem to remove our event listeners anywhere else in the tests... I'm not sure why it was done here originally, unless it was being activated too many times in this one test?

AllanKerr: Yeah, it does seem to be done in a bunch of other places and it sounds like the fixture should get automatically cleaned up. It looks like it passes so I'm good with it as is.

PR #790, opened 2020-04-13
US114859 - Move search widget to class based

svanherk:
More modulizer cleanup, moving the search widget to es6 class based format

### DIFF
--- a/src/d2l-my-courses-content.js
+++ b/src/d2l-my-courses-content.js
@@ -327,14 +327,15 @@ class MyCoursesContent extends mixinBehaviors([
 		document.body.addEventListener('set-course-image', this._onSetCourseImage);
 		document.body.addEventListener('d2l-tab-panel-selected', this._onTabSelected);
 
+		this.addEventListener('course-tile-organization', this._onCourseTileOrganization);
+		this.addEventListener('course-image-loaded', this._onCourseImageLoaded);
+		this.addEventListener('initially-visible-course-tile', this._onInitiallyVisibleCourseTile);
+		this.addEventListener('d2l-enrollment-new', this._onD2lEnrollmentNew);
+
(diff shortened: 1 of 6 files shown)


src/d2l-my-courses-content.js
```diff
@@ -327,14 +327,15 @@ class MyCoursesContent extends mixinBehaviors([
 		document.body.addEventListener('set-course-image', this._onSetCourseImage);
 		document.body.addEventListener('d2l-tab-panel-selected', this._onTabSelected);
 
+		this.addEventListener('course-tile-organization', this._onCourseTileOrganization);
+		this.addEventListener('course-image-loaded', this._onCourseImageLoaded);
+		this.addEventListener('initially-visible-course-tile', this._onInitiallyVisibleCourseTile);
+		this.addEventListener('d2l-enrollment-new', this._onD2lEnrollmentNew);
```
AllanKerr, line 333: What was the reasoning for moving these 4 out of `afterNextRender` but leaving the other 3 inside?

svanherk: I chose to move out the ones that happen sometime during the loading process, and leave the ones that only get fired if the user does something (presses a button, etc).  None of these were being delayed before when in the `listeners` array so I thought it better to move these back to be safe.  I'm kinda tempted to move all of them back because it simplifies the tests but I think the three left inside _should_ be fine

AllanKerr: I was thinking they could likely all be moved out since they're all added directly to `this` element. I'm fine either way so feel free to do what you feel introduces the least risk.

svanherk: I do think moving them all out is the least risky, so yeah I'll do that

PR #791, opened 2020-04-14
US114859 - Use lit eslint config in src

svanherk:
Use the lit eslint config in the `src` folder.

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -1,0 +1,7 @@
+{
+  "extends": "brightspace/lit-config",
+  "rules": {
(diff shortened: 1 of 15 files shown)


src/.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "brightspace/lit-config",
+  "rules": {
+    "no-prototype-builtins": 0,
+    "lit/no-legacy-template-syntax": 0
```
svanherk, line 5: This catches a bunch of false positives while we're still using Polymer, so disable for now

PR #792, opened 2020-04-15
US114859 - Remove old dom method and broken code

svanherk:
- Cleanup usage of the old Polymer 1 dom method
- Cleanup some broken code that was using this

### DIFF
--- a/src/d2l-my-courses-content.js
+++ b/src/d2l-my-courses-content.js
@@ -22,8 +22,6 @@ import { entityFactory, updateEntity } from 'siren-sdk/src/es6/EntityFactory.js'
 import { html, PolymerElement } from '@polymer/polymer/polymer-element.js';
 import { Actions } from 'd2l-hypermedia-constants';
 
-import { dom } from '@polymer/polymer/lib/legacy/polymer.dom.js'; //todo: Remove
-
 import { EnrollmentCollectionEntity } from 'siren-sdk/src/enrollments/EnrollmentCollectionEntity.js';
 import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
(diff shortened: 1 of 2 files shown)


src/d2l-my-courses-content.js
```diff
@@ -571,14 +569,6 @@ class MyCoursesContent extends mixinBehaviors([
 		}));
 		this._isRefetchNeeded = false;
 
-		// Pretty sure this is broken... it returns the my-courses component in Shadow DOM, and even when I set it to the actual enrollment-card component,
-		// enrollmentCard.hasAttribute('completed') and enrollmentCard.hasAttribute('closed') always seem to return false even when the attribute is there
-		// I think because things have already started "moving" and dom-repeat is infuriating
-		// Going to cleanup next PR
-		const enrollmentCard = dom(e).event && dom(e).event.srcElement;
-
-		const shouldHide = enrollmentCard && !isPinned && (enrollmentCard.hasAttribute('completed') || (enrollmentCard.hasAttribute('closed')));
```
svanherk, line 580: From my testing, this is broken because `dom(e).event.srcElement` is giving us the `my-courses` component with shadow dom on, not the enrollment card.  Even when we get the enrollment card in shady dom, `enrollmentCard.hasAttribute('completed')` and `enrollmentCard.hasAttribute('closed')` always return false even when the attributes are there... I suspect because of `dom-repeat` craziness.  After speaking to Maya and Ed, they confirmed we don't hide completed cards anymore, another clue this is broken.

The code that actually handles the hiding is here: https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content.js#L420-L438

PR #793, opened 2020-04-15
DE38482 - Fix widget settings refresh

svanherk:
This fixes https://rally1.rallydev.com/#/15545167705ud/detail/defect/379581808432?fdp=true, which we fixed once before by adding `disable-entity-cache` to the `d2l-my-courses-content` component (https://github.com/Brightspace/d2l-my-courses-ui/pull/752/files).  Since we know longer use the entity behaviour, we can't use that.  Instead, we manually bust the cache if the presentation url is reassigned, which should only happen if the widget is reloaded after the settings have changed.  This is actually better than the previous solution, which refetches the presentation url for every tab.  This sends the same url into each tab.

We bust the cache like this in other places as well, until we have some type of HM change push functionality.

### DIFF
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -162,7 +162,8 @@ class MyCoursesContainer extends mixinBehaviors([
 
 		if (userSettingsEntity.userSettingsHref()) {
 			// We need to bust the cache so the entity store will refetch the presentation details after widget settings are updated
-			this._presentationUrl = `${userSettingsEntity.userSettingsHref()}?bustCache=${Math.random()}`;
+			const url = userSettingsEntity.userSettingsHref();
+			this._presentationUrl = `${url}${(url.indexOf('?') !== -1 ? '&' : '?')}bustCache=${Math.random()}`;
 		}
 
 		this._updateUserSettingsAction = userSettingsEntity.userSettingsAction();
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -161,7 +161,8 @@ class MyCoursesContainer extends mixinBehaviors([
 		}
 
 		if (userSettingsEntity.userSettingsHref()) {
-			this._presentationUrl = userSettingsEntity.userSettingsHref();
+			// We need to bust the cache so the entity store will refetch the presentation details after widget settings are updated
+			this._presentationUrl = `${userSettingsEntity.userSettingsHref()}?bustCache=${Math.random()}`;
 		}
 
 		this._updateUserSettingsAction = userSettingsEntity.userSettingsAction();


src/d2l-my-courses-container.js
```diff
@@ -161,7 +161,8 @@ class MyCoursesContainer extends mixinBehaviors([
 		}
 
 		if (userSettingsEntity.userSettingsHref()) {
-			this._presentationUrl = userSettingsEntity.userSettingsHref();
+			// We need to bust the cache so the entity store will refetch the presentation details after widget settings are updated
+			this._presentationUrl = `${userSettingsEntity.userSettingsHref()}?bustCache=${Math.random()}`;
```
AllanKerr, line 165: Is it always guaranteed that this `href` always has no existing query params?

svanherk: From my experience testing, yes, but you're right that it could change in the LMS/HM route and we're not handling it very gracefully.  I'll add some logic to handle that just in case

src/d2l-my-courses-container.js
```diff
@@ -161,7 +161,9 @@ class MyCoursesContainer extends mixinBehaviors([
 		}
 
 		if (userSettingsEntity.userSettingsHref()) {
-			this._presentationUrl = userSettingsEntity.userSettingsHref();
+			// We need to bust the cache so the entity store will refetch the presentation details after widget settings are updated
+			const url = userSettingsEntity.userSettingsHref();
+			this._presentationUrl = `${url}${(url.indexOf('?') !== -1 ? '&' : '?')}bustCache=${Math.random()}`;
```
m6jones, line 166: if the usersettingEntity is what you want to bust. You could try `userSettingsEntity.update()` That should also send the headers to refetch. It sets the bypasscache in the entity store fetch method. 

This might not work. Since I forget if this functionality was fully implemented or not.

svanherk: I don't _think_ it's the `userSettingsEntity` that we want to update, it is actually getting updates properly and returning us the new `userSettingsHref` (which is the same as last time, it will just return a different result when called).  We then pass that in to the `d2l-my-courses-content`, and it's _that_ component that uses it.  I was going to try to use the `update` function on the `PresentationEntity` (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content.js#L680) but by the time the presentation url updates the old component has been lost and it's actually a completely new `d2l-my-courses-content` component.
Not sure if I've explained that very well...

m6jones: Makes sense to me. Yup that is an issue so nevermind on my comment.

svanherk: Thanks for taking a look!  The `siren-sdk` has grown a lot since I last was in there and I'm not sure all the tools available to use

PR #795, opened 2020-04-16
US114864 - Make sure langterms are alphabetical

svanherk:
All other files are already in alphabetical order.  This change is to prep for the next PR where I'm removing unused langterms, which will be a lot easier to review if everything is in the same order.


d2l-my-courses.serge.json
```diff
@@ -4,11 +4,15 @@
   "output_file_path": "lang/%LANG%.json",
   "output_lang_rewrite": [
     "ar-sa ar",
+    "da-dk da-dk",
```
AllanKerr, line 7: Were these added because we're now translating into more languages?

svanherk: Yeah I think so, there seemed to be a one-time push from serge of these new files about a year ago but they were never added here.  But I actually don't know if having them missing here actually prevented serge from trying to give us new translations for them, since we haven't added any terms since then

PR #796, opened 2020-04-16
US114864 - Remove unused langterms

svanherk:
Removing a bunch of unused langterms that are:
- Only used in legacy, which has its own non-changing translation files (https://github.com/Brightspace/d2l-my-courses-ui/tree/master/legacy/build/lang)
- Found in other repos as we moved components there (for example, `d2l-enrollment-card`: https://github.com/BrightspaceHypermediaComponents/enrollments/blob/master/components/lang/en.json#L114)


AllanKerr:
> * Only used in legacy, which has its own non-changing translation files (https://github.com/Brightspace/d2l-my-courses-ui/tree/master/legacy/build/lang)

Based on this I assume we're good with never adding or updating any lang terms in the legacy code path since won't be any files we can generate the legacy `build/lang` from after this? Unless I missed something since there are a lot of lang files 😅

svanherk:
> Based on this I assume we're good with never adding or updating any lang terms in the legacy code path since there won't be any files we can generate the legacy build/lang from after this? Unless I missed something since there are a lot of lang files 😅

Yup, this is the assumption/decision I have made.  Sandra was good with the old code path never getting any updates, ever

PR #797, opened 2020-04-17
US114864 - Move to new serge parsing format and mixin

svanherk:
- Move to the new JS serge parser so we don't need to build langterms anymore
- Remove old template and gulp build stuff
- Remove json files
- Move from the my-courses localize behavior to a mixin


svanherk:
It's probably easier to review each commit of this PR with whitespace changes hidden than the whole thing at once.  Git is not very smart once you move files 😒

PR #801, opened 2020-04-22
US115271 - Move behavior and styles back into grid component

svanherk:
I can't delete the behavior and styles until `d2l-my-courses-content` also uses this component, which is the next PR.

### DIFF
--- a/src/d2l-my-courses-card-grid.js
+++ b/src/d2l-my-courses-card-grid.js
@@ -1,20 +1,20 @@
 /*
-`d2l-all-courses-content`
+`d2l-my-courses-card-grid`
 Polymer-based web component for the all courses content.
 */
 
 import 'd2l-enrollments/components/d2l-enrollment-card/d2l-enrollment-card.js';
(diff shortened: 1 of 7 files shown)


src/d2l-my-courses-card-grid.js
```diff
@@ -88,10 +119,75 @@ class MyCoursesCardGrid extends mixinBehaviors([
...
+			return;
+		}
+
+		let containerWidth = this.offsetWidth;
+
+		for (let parent = this.parentNode; containerWidth <= 0 && parent; parent = parent.parentNode) {
+			containerWidth = parent.offsetWidth;
+		}
+
+		if (isNaN(containerWidth)) {
```
svanherk, line 145: The styles and behavior code is direct copy into here except for this `if` statement.  I added this because I'm seeing `columns-NaN` being added to the course grid's class, which isn't useful.  If `containerWidth` isn't a number, no point in trying to do this work yet anyways.

PR #802, opened 2020-04-24
Cert 20.20.5 - DE38830 - Fix image upload

svanherk:
Replace all instances of `$$` in src with `this.shadowRoot.querySelector`.

### DIFF
--- a/d2l-my-courses.js
+++ b/d2l-my-courses.js
@@ -89,13 +89,13 @@ class MyCourses extends PolymerElement {
 
 	courseImageUploadCompleted(success) {
 		return this.updatedSortLogic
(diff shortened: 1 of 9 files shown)


d2l-my-courses.js
```diff
@@ -89,13 +89,13 @@ class MyCourses extends PolymerElement {
 
 	courseImageUploadCompleted(success) {
 		return this.updatedSortLogic
-			? this.$$('d2l-my-courses-container').courseImageUploadCompleted(success)
```
svanherk, line 92: Technically it was only the usage in this file that was failing because of legacy Polymer imports in other files, but I'm going to fix all of them so it doesn't happen again.

PR #805, opened 2020-04-27
US115271 - More shared component prep

svanherk:
Some more cleanup to try to make the next PR a bit cleaner

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -349,7 +349,6 @@ class AllCourses extends mixinBehaviors([
 									<div hidden$="[[!_showTabContent]]">
 										<d2l-my-courses-card-grid
 											token="[[token]]"
-											org-unit-type-ids="[[orgUnitTypeIds]]"
 											show-organization-code="[[showOrganizationCode]]"
 											show-semester-name="[[showSemesterName]]"
 											show-dropbox-unread-feedback="[[showDropboxUnreadFeedback]]"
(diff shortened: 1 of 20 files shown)


src/d2l-all-courses.js
```diff
@@ -349,7 +349,6 @@ class AllCourses extends mixinBehaviors([
 									<div hidden$="[[!_showTabContent]]">
 										<d2l-my-courses-card-grid
 											token="[[token]]"
-											org-unit-type-ids="[[orgUnitTypeIds]]"
```
svanherk, line 352: These weren't doing anything, the component does nothing with these

src/d2l-my-courses-card-grid.js
```diff
@@ -116,21 +119,26 @@ class MyCoursesCardGrid extends PolymerElement {
...
 		});
 	}
 
 	disconnectedCallback() {
 		super.disconnectedCallback();
-		window.removeEventListener('resize', this._onResize.bind(this));
+		window.removeEventListener('resize', this.onResize);
 	}
 
-	_onResize(ie11retryCount) {
```
svanherk, line 133: The `d2l-my-courses-content` component is going to need to tell the grid when to recalculate, so making this function public

src/d2l-all-courses.js
AllanKerr: Do you know what they were being used for in the past? I tried to find the code that referenced them that was dropped but had no luck because `d2l-my-courses-card-grid` lost its history on rename.

svanherk: Yeah it's a mess of a history to track... looks like ti was added here: https://github.com/Brightspace/d2l-my-courses-ui/commit/b229135240e974d8fd75abbf0325dad451dfe281#diff-7d9e8e373cdb89087fc53450617b7257R131
But `d2l-all-courses-unified-content` at that time wasn't updated or using it... https://github.com/Brightspace/d2l-my-courses-ui/blob/b229135240e974d8fd75abbf0325dad451dfe281/src/card-grid/d2l-all-courses-unified-content.js

So I think it's never been used, it gets used in the actual hypermedia calls instead.  This was before the code split so extra confusing

PR #806, opened 2020-04-28
US115271 - DE38877 - Use grid component in d2l-my-courses-content

svanherk:
Finally switching d2l-my-courses-content to use the new grid component here.  I tried to break up the commits to make it easier to follow, and I'll add comments throughout.


legacy/d2l-my-courses-content-behavior-legacy.js
```diff
@@ -4,7 +4,6 @@ import { Actions } from 'd2l-hypermedia-constants';
 import 'd2l-polymer-siren-behaviors/store/entity-behavior.js';
 import 'd2l-polymer-siren-behaviors/store/siren-action-behavior.js';
 import './d2l-all-courses-legacy.js';
-import '../src/card-grid/d2l-card-grid-behavior.js';
```
svanherk, line 7: This behaviour wasn't actually used below and should not be included here (this looks to be the last reference to `src` in `legacy`)

src/d2l-all-courses.js
```diff
@@ -584,7 +594,6 @@ class AllCourses extends mixinBehaviors([
 		this._isSearched = !!e.detail.searchValue;
 		this._updateFilteredEnrollments(e.detail.searchResponse, false);
 		this._myEnrollmentsEntity = e.detail.searchResponse;
-		this.fire('recalculate-columns');
```
svanherk, line 587: This event is only caught by the old legacy tile grid, and does nothing here

src/d2l-my-courses-card-grid.js
```diff
@@ -15,8 +15,21 @@ class MyCoursesCardGrid extends PolymerElement {
...
+			// since we remove closed courses from the filteredEnrollments array on load)
+			hidePastCourses: {
+				type: Boolean,
+				value: false,
+				reflectToAttribute: true
+			},
 			// Token JWT Token for brightspace | a function that returns a JWT token for brightspace
 			token: String,
+			// Limit the number of cards shown to 12 (unless more than 12 are pinned)
+			widgetView: {
```
svanherk, line 28: Named this generally for now because anything else specific to the widget view (versus the all courses view) can be controlled by this too

src/d2l-my-courses-content.js
```diff
@@ -353,26 +351,25 @@ class MyCoursesContent extends mixinBehaviors([
...
+	// If it's a catalog image this is handled by the enrollment card
 	courseImageUploadCompleted(success) {
 		if (success) {
 			this.$['basic-image-selector-overlay'].close();
-			this._refreshTileGridImages();
-		}
-		this.focus();
-	}
-	focus() {
-		if (this._getTileGrid().focus(this._setImageOrg)) {
```
svanherk, line 365: This doesn't make any sense, this was getting a div and trying to call a special focus function on it.  Looks like this was leftover from the old legacy grid component and has been broken for a while.  Instead of only calling this focus function after a user-uploaded image, I have added a fix that handles focus every time the image selector is closed down below.

src/d2l-my-courses-content.js
```diff
@@ -518,6 +506,9 @@ class MyCoursesContent extends mixinBehaviors([
 		if (this._courseTileOrganizationEventCount === this._initiallyVisibleCourseTileCount) {
 			// Only show content once the last visible organization has loaded, to reduce jank
 			this._showContent = true;
```
svanherk, line 508: Whenever we set `this._showContent` to `true`, we should resize the card grid because its width is no longer zero.  There seems to have been enough of a delay before with the behavior loading that things kinda "worked out", but now we want to be more explicit.

src/d2l-my-courses-content.js
```diff
@@ -891,9 +906,8 @@ class MyCoursesContent extends mixinBehaviors([
 				&& this.tabSearchType.toLowerCase() === 'bysemester'
 			)
 		) {
-			// This is needed for hiding a closed course that was just unpinned
-			this._getTileGrid().setAttribute('hide-past-courses', '');
```
svanherk, line 895: This is now hooked up as a property, so just need to set `this._hidePastCourses` below

src/d2l-my-courses-content.js
```diff
@@ -917,8 +931,6 @@ class MyCoursesContent extends mixinBehaviors([
 			this._showContent = true;
 		}
 
-		this.fire('recalculate-columns');
```
svanherk, line 920: Like in the other file, this hasn't done anything for a while

src/d2l-my-courses-content.js
```diff
@@ -641,15 +632,35 @@ class MyCoursesContent extends mixinBehaviors([
 			};
 		});
 	}
-	_onSimpleOverlayClosed() {
+	_onSimpleOverlayClosed(e) {
```
svanherk, line 635: This is the focus fix.  When we get a simple overlay closed event, we now check if it was the image selector that was closed, or the all courses overlay.  If it was the all courses overlay, we do the same work as before.  If it's the image selector, we try to return focus to the appropriate card, wort case falling back to the all course link like we used to

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -273,51 +273,19 @@ describe('d2l-my-courses-content', () => {
 			setTimeout(done, 300);
 		});
 
-		it('should set the columns-"n" class on the correct card grid on resize', done => {
```
svanherk, line 276: This test was updated and moved to `d2l-my-courses-card-grid.js`

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -273,51 +273,19 @@ describe('d2l-my-courses-content', () => {
...
+			const stub = sandbox.stub(courseGrid, 'refreshCardGridImages');
 
 			component.courseImageUploadCompleted(true);
 
-			expect(stub1).to.have.been.called;
-			//expect(stub2).to.have.been.called;
+			expect(stub).to.have.been.called;
 		});
 
-		it('should call focus on the correct card grid when focus is called', () => {
```
svanherk, line 308: This test no longer makes sense, so it has been deleted.  I added some focus test to `d2l-my-courses-card-grid.js`.

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -328,10 +296,6 @@ describe('d2l-my-courses-content', () => {
 			expect(component.courseImageUploadCompleted).to.be.a('function');
 		});
 
-		it('should implement focus', () => {
```
svanherk, line 331: I _believe_ this test would never fail, no matter the component

test/d2l-my-courses-content/d2l-my-courses-content.js
```diff
@@ -705,13 +671,6 @@ describe('d2l-my-courses-content', () => {
 			});
 		});
 
-		it('should rescale the course tile grid on search response', () => {
```
svanherk, line 708: Deleted, since this is testing the unused events

src/d2l-my-courses-content.js
AllanKerr: Is there a reason we don't also want this in inside `_enrollmentRefetchResponse` and `_handleEnrollmentsRefetch`? Also, I assume there are no issues with a flash on load due to the loading spinner getting removed an animation frame before resize is called?

src/d2l-my-courses-content.js
AllanKerr: Is it worth attaching separate event listeners to the two overlays so that they can both be handled in their own handlers rather than having to check the ids?

test/d2l-my-courses-card-grid/d2l-my-courses-card-grid.js
```diff
@@ -24,5 +24,94 @@ describe('d2l-my-courses-card-grid', function() {
...
+
+			const response = widget.focusCardDropdown('org2');
+
+			expect(response).to.be.true;
+			expect(stub1).to.have.been.calledOnce;
+			expect(stub2).to.have.been.calledOnce;
+			expect(stub3).to.have.not.been.called;
+		});
+
+		it('focusCardDropdown should return false if the org does not match and cards', () => {
```
AllanKerr, line 99: I think theres a typo here:
> if the org does not match ~and~ any cards

src/d2l-my-courses-content.js
svanherk: Hmm I think I just missed `_enrollmentRefetchResponse`, will add it there!  `_handleEnrollmentsRefetch` sets it to false so resizing there will just measure `containerWidth` of 0 and bail.  The flash currently does exist, I just want it to happen a bit faster... when the loading spinner is there, `display: none` is on the card grid and we can't measure it at all.  We have to wait until it's measurable to calculate.  We maybe could do something fancy with opacity instead in the future... 🤔

src/d2l-my-courses-content.js
svanherk: So I couldn't easily access the all courses `simple-overlay` without reaching into its `shadowRoot`, and only after its "created" because of the delayed creation... but seeing what this function turned into, you're right that's way better and `d2l-all-courses` should expose a function to return its overlay`.
My plan is to move _both_ the image selector overlay and the all courses overlay out of here and into the `d2l-my-courses-container`, so there is only one (not one per tab).  Are you good with this for now and I'll do that as I move them, or would you like to see this hooked up properly before merge?

AllanKerr: Yep, I'm fine with it as is!

src/d2l-my-courses-content.js
AllanKerr: Ah, yeah, my mistake, I think I meant `_populateEnrollments` instead of `_handleEnrollmentsRefetch` but its not immediately clear if it actually needs it. It does set `this._showContent` to `true` but is also called from a few other places that already set it to `true` and call resize.

I'm okay with a bit of a flash if its a hard fix and it isn't a new regression.

PR #807, opened 2020-05-01
Update d2l-all-courses.js

klhuff:
DE38875 - Force default sort to "Current" and search to blank when opening or re-opening the all courses overlay.


src/d2l-all-courses.js
```diff
@@ -446,7 +446,7 @@ class AllCourses extends mixinBehaviors([
 				autoPinCourses: false,
 				orgUnitTypeId: this.orgUnitTypeIds,
 				embedDepth: 0,
-				sort: this._sortParameter || 'Current'
+				sort: 'Current'
```
klhuff, line 449: In load: always force the default sort param which is "Default".

src/d2l-all-courses.js
```diff
@@ -596,6 +596,15 @@ class AllCourses extends mixinBehaviors([
 	}
 
 	_onSimpleOverlayOpening() {
+		if (this._enrollmentsSearchAction) {
```
klhuff, line 599: This is probably the most weird/interesting part of this change.

I wanted to put both of these field changes into their respective "clear" methods. But the problem is that each clear (whether it be the filters, or the search, or the sort dropdown) spawn their own events which cause the search url to  change. Because these happen in event handlers, you can't predict which order they're going to happen in. Before hand, this "clear" code existed, but it didn't always work, because this on overlay load method was just one giant race condition. So, here, I'm opting to set the defaults before anything else happens.

klhuff:
I clearly have some things to fix, so maybe don't look at this right away.

klhuff:
It turns out, I had to do some mucking around with installing the polymer CLI globally in order to get the tests running locally. But, it seems I've sorted out my environment now, and this is ready for review.

src/d2l-all-courses.js
```diff
@@ -784,6 +793,7 @@ class AllCourses extends mixinBehaviors([
 
 	_resetSortDropdown() {
 		this._selectSortOption(this._defaultSortValue);
+		this._sortParameter = 'Current';
```
AllanKerr, line 796: Do we need to update the `_searchUrl` since the `_sortParameter` is changing?

src/d2l-all-courses.js
svanherk: Do you know why this only runs if there's no tabs?  Maybe because it's handled in `_onTabSelected`?

klhuff: That's correct. When the overlay first loads, it by default selects the "All" tab, and thus hits the handler.

src/d2l-all-courses.js
svanherk: I'm also wondering how this is working... 😕

src/d2l-all-courses.js
```diff
@@ -605,6 +605,15 @@ class AllCourses extends mixinBehaviors([
 	}
 
 	_onSimpleOverlayOpening() {
+		if (this._enrollmentsSearchAction && this._enrollmentsSearchAction.hasFieldByName) {
```
svanherk, line 608: Could we just clear `_enrollmentsSearchAction` completely here?  Or if load has already run, maybe in `load`?  Does that retrigger everything properly?

src/d2l-all-courses.js
svanherk: @njostonehouse The different sorts for > 50 enrollments versus < 50 enrollments... are both of those called "current" and differentiated on the back end?  Like if I have more than 50 enrollments and choose to sort by current, does that really just give me last accessed with pins promoted?

src/d2l-all-courses.js
svanherk: Hmm maybe not, if that gets set by the `EnrollmentsEntity` changes... just trying to think of a simple way to reset things and let the flow happening similar to the first time things open, but may not be simple

src/d2l-all-courses.js
klhuff: So the reason this works, is because the search action is being updated immediately in the Overlay opening. Handlers for both of the events generated by clearing the filters (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/search-filter/d2l-filter-menu.js#L250), and by clearing the search control (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/search-filter/d2l-search-widget-custom.js#L131) wind up updating the search url. By either, causing the enrollment entity to change (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-all-courses.js#L693) or, by changing the search results (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-all-courses.js#L596) which causes the entity to change. Note that, when computing the sort here, it doesn't look at _sortParameter, but looks at what is set on the action already.

I could move this line into the if on line 608 to make it more clear.

klhuff: It might also help to know, that this line wasn't needed in the case where a user closes, and then re-opens the overlay. It was needed to cover the following case:

1) Open overlay.
2) Switch sort order.
3) Switch tabs.
4) Close overlay.
5) Open overlay (at this point, it was reset to default as you'd expect).
6) Change tabs back to the original tab.
At this point, things get messed up if _sortParameter isn't set, because it gets remembered when the tab changes.

klhuff: On top of the above, there's also an event that fires when the Sort Option is changed:
https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-all-courses.js#L542

Which uses the event details to determine the sort order, and not the _sortParameter.

src/d2l-all-courses.js
klhuff: Yeah, I definitely experimented with a bunch of things here. Unfortunately though, un-setting the action completely results in a bunch of javascript errors because the underlying components (the search control, and the filter mostly) are relying on the action retaining some data after they've been intialized.

I did have a solution at one point that simply set the `_enrollmentSearchAction` to either the default `enrollmentSearchAction` or the correct `tabSearchAction` here (depending on if the view had tabs or not), but it ran into problems because the default `enrollmentSearchAction` itself wants to sort by `OrgUnitId` and nothing else, and required the sort to be set manually anyway.

src/d2l-all-courses.js
svanherk: Thanks Kevin, this helps a lot!  I think moving it up to the line where you do the sort clearing would make it a bit more clear for me

svanherk: It's unfortunate that it seems to be in a state where so many different things are looked at, depending on the scenario... it can look at the event, or `_sortParameter`, or the action itself. 😞 Not sure how easy it would be to clean that up, but maybe we can get to a point where the `_sortParameter` isn't needed and the `action` is the source of truth, with the event requesting updates.

src/d2l-all-courses.js
njostonehouse: IIRC if you request `current` with > 50 Enrollments, the backend will override your decision, choose a better (ie: more performant) sort order, and return the sort order it actually used in the `self` ref of the response.

svanherk: Thanks Noel!

PR #809, opened 2020-05-05
Khuff/us113521

klhuff:
This pull request does a few things:
1) Remove tracking of the sort from a member variable (ie. remove _sortParameter). Instead, let the action be the source of truth in all cases.
2) Refactor the sort code such that there is only one method for mapping all of the sort properties (langterm, sort action, promotepins).

3) (This is the functional change): set promotePins == true for the default sort, so that it is consistent with the widget itself.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -489,7 +489,8 @@ class AllCourses extends mixinBehaviors([
 				autoPinCourses: false,
 				orgUnitTypeId: this.orgUnitTypeIds,
 				embedDepth: 0,
-				sort: 'Current'
+				sort: this._sortMap[0].action,
+				promotePins: this._sortMap[0].promotePins
 			})
 		);
 	}
@@ -583,9 +584,6 @@ class AllCourses extends mixinBehaviors([
 	}
 
 	_onSortOrderChanged(e) {
-		let sortParameter, langterm;
-		let promotePins = false;
-
 		var sortData = this._mapSortOption(e.detail.value, 'name');
 
 		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.js
```diff
@@ -441,12 +479,16 @@ class AllCourses extends mixinBehaviors([
 		if (!this.enrollmentsSearchAction) {
 			return;
 		}
+
+		const sortData = this._mapSortOption(this._sortMap[0].name, 'name');
```
klhuff, line 483: I did this in a few spots: use the sort map directly. I'm not sure if I should have bothered to use the map function here or if I should have just plugged values from `_sortMap[0]` down below. Let me know what you think/what seems cleaner.

src/d2l-all-courses.js
```diff
@@ -792,8 +776,20 @@ class AllCourses extends mixinBehaviors([
...
+				return this._sortMap[i];
+			}
+		}
+
+		return this._sortMap[0];
+	}
+
 	_resetSortDropdown() {
-		this._selectSortOption(this._defaultSortValue);
+		const sortData = this._mapSortOption(this._sortMap[0].name, 'name');
```
klhuff, line 791: In hindsight this one in particular seems really silly, because I'm only using the name. Definitely thinking about changing this, and the one in the load method to use just use `_sortMap[0]` directly instead of doing this weird call.

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -365,13 +365,11 @@ describe('d2l-all-courses', function() {
...
 
 		it('should set the _searchUrl based on the selected tab\'s action', function() {
-			widget._sortParameter = 'SortOrder';
 
 			widget.dispatchEvent(new CustomEvent(
 				'd2l-tab-panel-selected', { bubbles: true, composed: true }
 			));
-
-			expect(widget._searchUrl.indexOf('/example/foo?autoPinCourses=false&embedDepth=0&sort=SortOrder&search=&bustCache=') !== -1).to.be.true;
+			expect(widget._searchUrl.indexOf('/example/foo?autoPinCourses=false&embedDepth=0&sort=Current&search=&bustCache=') !== -1).to.be.true;
```
klhuff, line 372: This test needs to change because we no longer pay attention to _sortParameter, and the default is always Current.

src/d2l-all-courses.js
svanherk: Yeah, I think using it in both places directly is fine, especially since we do that in a couple other places.

PR #811, opened 2020-05-08
US116105 - Prep moving image selector to container component

svanherk:
This PR includes a collection of changes to make the next PR a little cleaner.  The next PR will be moving the image selector out from the `d2l-my-courses-content` component to the `d2l-my-courses-container` component.  This is to make `d2l-my-courses-content` a bit simpler, stop having one image selector per tab which isn't needed, and facilitate eventually moving to having one all-courses overlay as well (instead of one per tab).

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -281,7 +281,8 @@ class AllCourses extends mixinBehaviors([
 				id="all-courses"
 				title-name="[[localize('allCourses')]]"
 				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]"
-				with-backdrop>
+				with-backdrop
+				restore-focus-on-close>
 
 				<div hidden$="[[!_showContent]]">
 					<iron-scroll-threshold id="all-courses-scroll-threshold" on-lower-threshold="_onAllCoursesLowerThreshold">
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -449,7 +449,7 @@ class AllCourses extends mixinBehaviors([
 	}
 
 	focusCardDropdown(organization) {
-		this._getCardGrid().focusCardDropdown(organization);
+		return this._getCardGrid().focusCardDropdown(organization);
 	}
 
 	_getCardGrid() {
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -280,7 +280,8 @@ class AllCourses extends mixinBehaviors([
 			<d2l-simple-overlay
 				id="all-courses"
 				title-name="[[localize('allCourses')]]"
-				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]">
+				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]"
+				with-backdrop>
 
 				<div hidden$="[[!_showContent]]">
 					<iron-scroll-threshold id="all-courses-scroll-threshold" on-lower-threshold="_onAllCoursesLowerThreshold">
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -280,9 +280,7 @@ class AllCourses extends mixinBehaviors([
 			<d2l-simple-overlay
 				id="all-courses"
 				title-name="[[localize('allCourses')]]"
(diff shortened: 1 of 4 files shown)


src/d2l-all-courses.js
```diff
@@ -280,9 +280,7 @@ class AllCourses extends mixinBehaviors([
 			<d2l-simple-overlay
 				id="all-courses"
 				title-name="[[localize('allCourses')]]"
-				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]"
-				with-backdrop=""
```
svanherk, line 284: `d2l-simple-overlay` doesn't support `with-backdrop` or `restore-focus-on-close` (maybe it used to?)

src/d2l-my-courses-content.js
```diff
@@ -302,6 +301,22 @@ class MyCoursesContent extends mixinBehaviors([
 	/*
 	* Public API functions
 	*/
+
+	// After the image selector is closed, this is called to set focus back to the correct card
+	focusCardDropdown(imageOrg) {
```
svanherk, line 306: Moving this out to a public function that will be used by the `d2l-my-courses-container` in the next PR

src/d2l-my-courses-content.js
```diff
@@ -315,7 +330,35 @@ class MyCoursesContent extends mixinBehaviors([
 		if (!this._setImageOrg.links) {
 			return;
 		}
-		return this._getOrgUnitIdFromHref(this.getEntityIdentifier(this._setImageOrg));
+		return this.getOrgUnitIdFromHref(this.getEntityIdentifier(this._setImageOrg));
+	}
+
+	/*
+	* Changing Course Image Functions
```
svanherk, line 337: Just moving these functions to live altogether.  They'll be moving to the other component next PR

src/d2l-my-courses-content.js
```diff
@@ -583,9 +620,9 @@ class MyCoursesContent extends mixinBehaviors([
 		});
 	}
 	_onSimpleOverlayClosed(e) {
-		this._removeAlert('setCourseImageFailure');
 
 		if (e.composedPath()[0].id === 'all-courses') {
+			this._removeAlert('setCourseImageFailure');
```
svanherk, line 625: This only needs to happen if this is the all-courses overlay, since the image selector case handles this here: https://github.com/Brightspace/d2l-my-courses-ui/pull/811/files#diff-525f94b42429acc4c3aec0d06d7b7e6aR354

src/d2l-my-courses-content.js
```diff
@@ -710,14 +718,6 @@ class MyCoursesContent extends mixinBehaviors([
 
 		this._enrollmentsRootResponse(enrollmentsEntity);
 	}
-	_getOrgUnitIdFromHref(organizationHref) {
```
svanherk, line 713: Moved this to the utility behavior because both `d2l-my-courses-content` and `d2l-my-courses-container` will be using it soon

src/d2l-utility-behavior.js
```diff
@@ -82,12 +91,6 @@ D2L.MyCourses.UtilityBehavior = {
 			}))
 			.then(this.responseToSirenEntity.bind(this));
 	},
-	sirenEntityStoreFetch: function(url, token, clearCache) {
```
svanherk, line 85: This is only used in `legacy` (which has its own copy of the utility behavior).

src/d2l-utility-behavior.js
```diff
@@ -108,20 +111,7 @@ D2L.MyCourses.UtilityBehavior = {
 			}
 		}
 	},
-	submitForm: function(url, formParameters) {
```
svanherk, line 111: This is only used in `legacy` (which has its own copy of the utility behavior).

src/d2l-all-courses.js
svanherk: So turns out its the `iron-overlay-behavior` that understands these values.  I added `with-backdrop` back because without it, the behaviour trying to focus on the overlay after opening is broken.  I'm still going to leave `restore-focus-on-close` off because this functionality wasn't working for us and we now handle this ourselves.

svanherk: Turns out `restore-focus-on-close` was helpful too, I was testing the wrong thing 🤦 .  Added that back as well, sorry about all the churn

PR #812, opened 2020-05-12
DE38876

klhuff:
Add an observer for `token`, and only send the first enrollments and user settings requests when all three are set of:
-token
-enrollment url
-settings url

Before this change, the initial enrollments call, and user settings call are made without an authorization token.

### DIFF
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -46,7 +46,10 @@ class MyCoursesContainer extends mixinBehaviors([
 			// URL to fetch a user's settings (e.g. default tab to select)
 			userSettingsUrl: String,
 			// Token JWT Token for brightspace | a function that returns a JWT token for brightspace
-			token: String,
+			token: {
+				type: String,
+				observer: '_tokenChanged'
+			},
 			// URL to fetch widget settings
 			_presentationUrl: String,
 			_currentTabId: String,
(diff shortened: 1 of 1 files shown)


src/d2l-my-courses-container.js
```diff
@@ -242,6 +238,12 @@ class MyCoursesContainer extends mixinBehaviors([
 	_tabSelectedChanged(e) {
 		this._currentTabId = `panel-${e.detail.tabId}`;
 	}
+	_tokenChanged(token) {
+		if (token && this.enrollmentsUrl && this.userSettingsUrl) {
```
svanherk, line 242: I don't think we'll ever actually run into an issue because `token` should always be slower than `enrollmentsUrl` and `userSettingsUrl`, but I'm wondering if we should use an `observers` array instead (like https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-content.js#L170) and watch for changes on all three values?

klhuff: That seems reasonable to me. Will fix this.

klhuff: Fixed.

src/d2l-my-courses-container.js
```diff
@@ -46,7 +46,10 @@ class MyCoursesContainer extends mixinBehaviors([
 			// URL to fetch a user's settings (e.g. default tab to select)
 			userSettingsUrl: String,
 			// Token JWT Token for brightspace | a function that returns a JWT token for brightspace
-			token: String,
+			token: {
```
klhuff, line 49: Oops, need to get rid of this change.

PR #813, opened 2020-05-12
US116105 - Cleanup course image error alerts

svanherk:
Before moving the image selector, it's easier to simplify the alerts it causes now.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -17,7 +17,6 @@ import 'd2l-alert/d2l-alert.js';
 import 'd2l-organization-hm-behavior/d2l-organization-hm-behavior.js';
 import 'd2l-simple-overlay/d2l-simple-overlay.js';
 import 'd2l-tabs/d2l-tabs.js';
-import './d2l-alert-behavior.js';
 import './d2l-my-courses-card-grid.js';
 import './d2l-utility-behavior.js';
 import './search-filter/d2l-filter-menu.js';
@@ -33,7 +32,6 @@ import SirenParse from 'siren-parser';
 
 class AllCourses extends mixinBehaviors([
 	D2L.PolymerBehaviors.Hypermedia.OrganizationHMBehavior,
-	D2L.MyCourses.AlertBehavior,
 	D2L.MyCourses.UtilityBehavior
 ], MyCoursesLocalizeBehavior(PolymerElement)) {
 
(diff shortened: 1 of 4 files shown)


src/d2l-all-courses.js
```diff
@@ -458,12 +459,12 @@ class AllCourses extends mixinBehaviors([
 	}
 
 	_onSetCourseImage(details) {
```
svanherk, line 461: This function and the almost-identical one in `d2l-my-courses-content` will be moving to `d2l-my-courses-container` with the image selector.  Once that happens, we'll control the hiding/showing of this alert by setting a property the child components, so I'm moving to the property now.  `showImageError` is only used privately right now, but will be used publicly in the next PR so setting it to public now.

src/d2l-all-courses.js
```diff
@@ -17,7 +17,6 @@ import 'd2l-alert/d2l-alert.js';
 import 'd2l-organization-hm-behavior/d2l-organization-hm-behavior.js';
 import 'd2l-simple-overlay/d2l-simple-overlay.js';
 import 'd2l-tabs/d2l-tabs.js';
-import './d2l-alert-behavior.js';
```
svanherk, line 20: Removing the alerts behaviour because this is the only alert in all courses anyways.  In a future PR, I'm going to refactor the other alerts in `d2l-my-courses-content` as well - after the split and cleanup from the legacy code this behaviour isn't really need anymore.

src/d2l-my-courses-content.js
```diff
@@ -38,6 +38,11 @@ class MyCoursesContent extends mixinBehaviors([
 			* Public Polymer properties
 			*/
 			enrollmentsSearchAction: Object,
+			// Set by the image selector when it experiences an error trying to set a new course image
+			showImageError: {
```
svanherk, line 42: These changes basically match the other file.  The `_onSetCourseImage` function will move to `d2l-my-courses-container`, and eventually I want to move the alert itself inside `d2l-my-courses-card-grid`.  I can't do it until tabs have been moved inside that component as well, since this alert needs to go above the tabs in the all courses overlay (and we're going to update the widget view to match).

test/d2l-all-courses/d2l-all-courses.js
```diff
@@ -87,44 +87,47 @@ describe('d2l-all-courses', function() {
 	});
 
 	describe('Alerts', function() {
```
svanherk, line 89: A lot of the test duplication between the two test files will also be cleaned up next PR

PR #815, opened 2020-05-14
US116105 - Move image selector to container

svanherk:
Ok, here's the main PR for this story.  Mostly a lot of re-arranging, moving the `d2l-basic-image-selector` to `d2l-my-courses-container` so that only one exists, instead of one per tab.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -387,7 +387,6 @@ class AllCourses extends mixinBehaviors([
 		this._onFilterDropdownClose = this._onFilterDropdownClose.bind(this);
 		this._onFilterChanged = this._onFilterChanged.bind(this);
 		this._onSearchResultsChanged = this._onSearchResultsChanged.bind(this);
-		this._onSetCourseImage = this._onSetCourseImage.bind(this);
 	}
 
 	connectedCallback() {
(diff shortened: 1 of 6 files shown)


src/d2l-all-courses.js
```diff
@@ -565,7 +549,6 @@ class AllCourses extends mixinBehaviors([
 			}
 		}
 
-		this.showImageError = false;
```
svanherk, line 568: Instead of clearing this here when the all courses overlay is opening, I moved this call into `d2l-my-courses-content` which clears it before it opens the overlay.  This will just make future cleanup easier when the all courses overlay is moved into `container` as well.

src/d2l-my-courses-content.js
```diff
@@ -739,9 +671,18 @@ class MyCoursesContent extends mixinBehaviors([
 
 		allCourses.open();
 
+		this.showImageError = false; // Clear image error when opening and closing the all courses overlay
+
 		e.preventDefault();
 		e.stopPropagation();
 	}
+	_setAllCoursesImageError(newValue) {
```
svanherk, line 679: This hookup is temporary.  Once `d2l-all-courses` is moved from `d2l-my-courses-content` to `d2l-my-courses-container` (like the image selector), then the container will just set the error on both the widget and the overlay without the widget having to pass it through.

test/d2l-my-courses-container/d2l-my-courses-container.js
```diff
@@ -197,52 +201,201 @@ describe('d2l-my-courses', () => {
 	});
 
 	describe('Public API', () => {
-		it('should call d2l-my-courses-content.courseImageUploadCompleted when not grouped by tab', done => {
-			component.updatedSortLogic = true;
+		it('If image setting was not a success, do nothing in courseImageUploadCompleted', done => {
```
AllanKerr, line 204: nit: `it('should ...'` reads a lot better

svanherk: Good call, that organization was also sloppy.  Updated!

PR #820, opened 2020-05-26
US114862 - Do not dynamically create all courses

svanherk:
This PR adds `d2l-all-courses` to the template, instead of dynamically creating it on mouseover of the link.  We think this was done for performance reasons, but the overlay is already setup to wait until opening to start loading things.

The next PR will move `d2l-all-courses` to `d2l-my-courses-container`, but the diff is nicer if we do this here first.


src/d2l-all-courses.js
```diff
@@ -616,9 +616,9 @@ class AllCourses extends mixinBehaviors([
 		if (e.detail.orgUnitId) {
 			orgUnitId = e.detail.orgUnitId;
 		} else if (e.detail.organization) {
-			orgUnitId = this._getOrgUnitIdFromHref(this.getEntityIdentifier(this.parseEntity(e.detail.organization)));
+			orgUnitId = this.getOrgUnitIdFromHref(this.getEntityIdentifier(this.parseEntity(e.detail.organization)));
```
svanherk, line 619: Some unrelated cleanup here to use the shared function from the utility behaviour, which is already being included in this component.

src/d2l-my-courses-content.js
```diff
@@ -278,9 +285,7 @@ class MyCoursesContent extends mixinBehaviors([
 		this._getCardGrid().refreshCardGridImages(imageOrg);
 
 		const allCourses = this.shadowRoot.querySelector('d2l-all-courses');
-		if (allCourses) {
```
svanherk, line 281: No longer need to check if this exists.  It will from the start.

src/d2l-my-courses-content.js
```diff
@@ -653,22 +651,8 @@ class MyCoursesContent extends mixinBehaviors([
 		return enrollmentsLength > 0 ? `${viewAllCourses} (${count})` : viewAllCourses;
 	}
 	_openAllCoursesView(e) {
-		this._createAllCourses();
-
 		const allCourses = this.shadowRoot.querySelector('d2l-all-courses');
-
-		allCourses.enrollmentsSearchAction = this.enrollmentsSearchAction;
 		allCourses.tabSearchActions = this.tabSearchActions;
```
svanherk, line 655: I wait to set this because it helps with the timing of the first load, and makes it so things aren't updating needlessly in the background after first opening.

PR #823, opened 2020-06-03
Use annotated event listeners in All Courses

svanherk:
This is closer to the format Lit uses, and seems to lessen some of the race conditions (because of Polymer magic) until I can get the full fix in.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -282,9 +282,10 @@ class AllCourses extends mixinBehaviors([
 
 			<d2l-simple-overlay
 				id="all-courses"
-				title-name="[[localize('allCourses')]]"
+				on-d2l-simple-overlay-opening="_onSimpleOverlayOpening"
 				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]"
 				restore-focus-on-close
+				title-name="[[localize('allCourses')]]"
 				with-backdrop>
 
 				<div hidden$="[[!_showContent]]">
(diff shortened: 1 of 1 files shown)


src/d2l-all-courses.js
```diff
@@ -282,9 +282,10 @@ class AllCourses extends mixinBehaviors([
 
 			<d2l-simple-overlay
 				id="all-courses"
-				title-name="[[localize('allCourses')]]"
+				on-d2l-simple-overlay-opening="_onSimpleOverlayOpening"
```
klhuff, line 285: Cool, I didn't know things could be wired up this way.

src/d2l-all-courses.js
```diff
@@ -306,9 +308,17 @@ class AllCourses extends mixinBehaviors([
...
 										<d2l-icon icon="tier1:chevron-down" aria-hidden="true"></d2l-icon>
 									</button>
-									<d2l-dropdown-content id="filterDropdownContent" no-padding="" min-width="350" render-content="">
+									<d2l-dropdown-content
+										id="filterDropdownContent"
+										on-d2l-dropdown-open="_onFilterDropdownOpen"
+										on-d2l-dropdown-close="_onFilterDropdownClose"
+										no-padding
+										min-width="350"
+										render-content>
```
klhuff, line 317: I'm assuming this is a better way to use the property than setting it to blank, as was happening before?

svanherk: Technically either are allowed (https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes), but I'm not sure why we were only doing this way here.  Updated for consistency and because I much prefer this version 😛

PR #824, opened 2020-06-08
US117309 - Move the pinning syncing to the container to control

svanherk:
Before moving `d2l-all-courses` to the `d2l-my-courses-container`, I needed to cleanup the pinning workflow.  The `d2l-my-courses-container` now controls the pinning event (from the main widget view, the all courses view, and the course selector) and passes that info down to the other components.

I tried to keep the existing behavior mostly the same for now (the all courses overlay does not instantly refresh like the widget view when something is pinned), but this makes it a lot easier to improve things later.


src/d2l-all-courses.js
```diff
@@ -596,32 +605,16 @@ class AllCourses extends mixinBehaviors([
...
+			if (!newValue.isPinned && actionName === Actions.enrollments.searchMyPinnedEnrollments) {
 				this._searchUrl = this._appendOrUpdateBustCacheQueryString(this._searchUrl);
 			}
 		}
 
-		let orgUnitId;
-		if (e.detail.orgUnitId) {
-			orgUnitId = e.detail.orgUnitId;
-		} else if (e.detail.organization) {
-			orgUnitId = this.getOrgUnitIdFromHref(this.getEntityIdentifier(this.parseEntity(e.detail.organization)));
```
svanherk, line 612: Through code search, I never found any situation that was setting `e.detail.organization` in the `d2l-course-pinned-change` event, so I didn't move that to the `d2l-my-courses-container`'s handling of this.

src/d2l-all-courses.js
```diff
@@ -546,13 +552,16 @@ class AllCourses extends mixinBehaviors([
...
-		if (this.hasEnrollmentsChanged) {
+		if (this._hasEnrollmentsChanged && this._searchUrl) {
+			this._hasEnrollmentsChanged = false;
 			this._bustCacheToken = Math.random();
 			this._searchUrl = this._appendOrUpdateBustCacheQueryString(this._searchUrl);
 		}
 	}
 
 	_onTabSelected(e) {
+		e.stopPropagation();
```
svanherk, line 563: Since the all courses tabs are nested in the main widget tabs (for now), not stopping this event was making the `eventListener` in `d2l-my-courses-content` try to handle these events as well, which is unnecessary.

src/d2l-all-courses.js
```diff
@@ -546,13 +552,16 @@ class AllCourses extends mixinBehaviors([
 		this.$.filterMenu.clearFilters();
 		this._filterText = this.localize('filtering.filter');
 		this._resetSortDropdown();
-		if (this.hasEnrollmentsChanged) {
+		if (this._hasEnrollmentsChanged && this._searchUrl) {
```
svanherk, line 555: There's no point in busting the cache and resetting the `_searchUrl` if it is null, meaning this is the first time we're opening the overlay.

src/d2l-all-courses.js
```diff
@@ -596,32 +605,16 @@ class AllCourses extends mixinBehaviors([
 		);
 	}
 
-	_onEnrollmentPinned(e) {
-		if (this._showGroupByTabs) {
+	_onCourseEnrollmentChange(newValue) {
+		if (this._showGroupByTabs && this._searchUrl) {
```
svanherk, line 609: There's no point in doing this `_searchUrl` is null, meaning we've never opened the overlay yet.

src/d2l-my-courses-container.js
```diff
@@ -163,20 +164,31 @@ class MyCoursesContainer extends mixinBehaviors([
...
 		afterNextRender(this, () => {
-			this.addEventListener('d2l-course-enrollment-change', this._onCourseEnrollmentChange);
 			this.addEventListener('d2l-tab-changed', this._tabSelectedChanged);
 		});
 
 		this.addEventListener('open-change-image-view', this._onOpenChangeImageView);
 		this.addEventListener('set-course-image', this._onSetCourseImage);
 		this.addEventListener('clear-image-scroll-threshold', this._onClearImageScrollThreshold);
 
+		document.body.addEventListener('d2l-course-pinned-change', this._onCourseEnrollmentChange);
```
svanherk, line 182: The container now listens for this event at the body level (instead of `d2l-my-courses-content` doing it), because it needs to catch pinning events from the courses selector waffle as well.

src/d2l-my-courses-content.js
```diff
@@ -441,78 +438,14 @@ class MyCoursesContent extends mixinBehaviors([
...
-		} else {
-			orgUnitId = this.getOrgUnitIdFromHref(e.detail.enrollment.organizationHref());
-		}
-		// Only want to move pinned/unpinned enrollment if it exists in the panel
-		const changedEnrollmentId = orgUnitId && this._orgUnitIdMap[orgUnitId];
-		if (!changedEnrollmentId) {
-			this._refetchEnrollments();
-		}
-
-		this.dispatchEvent(new CustomEvent('d2l-course-enrollment-change', {
```
svanherk, line 465: This event is no longer needed to tell the container something has changed - it knows, and is the one that told this component about it.

src/d2l-my-courses-content.js
```diff
@@ -565,9 +495,20 @@ class MyCoursesContent extends mixinBehaviors([
...
+		const changedEnrollmentId = newValue.orgUnitId && this._orgUnitIdMap[newValue.orgUnitId];
+		if (changedEnrollmentId) {
+			updateEntity(changedEnrollmentId, this.token);
+		}
+
+		if (this.tabSearchActions.length === 0 || this._thisTabSelected) {
+			// Only want to move pinned/unpinned enrollment if it exists in the panel
+			if (!changedEnrollmentId) {
+				this._refetchEnrollments();
+				return;
```
svanherk, line 507: This wasn't in `_onEnrollmentPinnedMessage` before, but I think that was a mistake.  It used to be and was lost in a re-org of promise chains: https://github.com/Brightspace/d2l-my-courses-ui/commit/b950db4f91b0ec2a1ae8497bc6c1c362f6c65e62#diff-8e9d13c03e9ba0329fc66a3fdc4941e9R408

src/d2l-my-courses-content.js
```diff
@@ -576,6 +517,41 @@ class MyCoursesContent extends mixinBehaviors([
 	_computeIsPinnedTab(actionName) {
 		return actionName === Actions.enrollments.searchMyPinnedEnrollments;
 	}
+	_rearrangeAfterPinning(changedEnrollmentId, isPinned) {
```
svanherk, line 520: This is a straight copy, just separated into its own function to keep `_onCourseEnrollmentChange` cleaner.

src/d2l-all-courses.js
```diff
@@ -45,6 +45,7 @@ class AllCourses extends mixinBehaviors([
 
 			// URL that directs to the advanced search page
 			advancedSearchUrl: String,
+			changedCourseEnrollment: Object,
```
AllanKerr, line 48: It feels a bit weird to passing up an object to notify components of changes. It feels like this would be more idiomatic to have a public `updateCourseEnrollment` method?

svanherk: Hmm, good point - I think it felt easier to pass through this way while `d2l-all-courses` still contained `d2l-my-courses-content`, but changing to a public method now would still be fairly readable and easy to update.

src/d2l-my-courses-container.js
```diff
@@ -318,10 +330,20 @@ class MyCoursesContainer extends mixinBehaviors([
...
-		};
+		let orgUnitId;
+		if (e.detail.orgUnitId) { // Pinning was done in the Course Selector
+			orgUnitId = e.detail.orgUnitId;
+		} else { // Pinning was done in the My Courses widget
+			orgUnitId = this.getOrgUnitIdFromHref(e.detail.enrollment.organizationHref());
+		}
+
+		// Only update the property if something has changed
+		if (!this._changedCourseEnrollment || this._changedCourseEnrollment.orgUnitId !== orgUnitId || this._changedCourseEnrollment.isPinned !== e.detail.isPinned) {
```
AllanKerr, line 341: Is this check necessary? It'd be nice to just be able to trust the `_onCourseEnrollmentChange` to only be fired if something actually changed.

svanherk: Agreed, it is currently very necessary and I haven't figured out why yet.  I see this event listener run multiple times for different events with the same details, different timestamps.  I think it might be an issue lower down... are you good if I add a comment and make a story to investigate this?

src/d2l-my-courses-content.js
```diff
@@ -441,78 +438,14 @@ class MyCoursesContent extends mixinBehaviors([
...
-		if (e.composedPath()[0] === this) return;
-
-		const isPinned = e.detail.isPinned;
-		let orgUnitId;
-
-		if (e.detail.orgUnitId) {
-			orgUnitId = e.detail.orgUnitId;
-			if (this._orgUnitIdMap[orgUnitId]) {
-				const enrollmentHref = this._orgUnitIdMap[orgUnitId];
-				updateEntity(enrollmentHref, this.token);
```
AllanKerr, line 454: Do you know why `updateEntity` was only called if the `orgUnitId` was attached directly rather than inferred from the enrollment? I noticed now we always call it if there is an `orgUnitId`. Is it possible this will result in double fetches?

svanherk: Nice catch, I meant to make a note about this!  So from what I can tell, it was added to handle changes from the course selector: https://github.com/Brightspace/d2l-my-courses-ui/commit/3218122dbc4c6b4ae55cf8f12624d528b7144bd2

But, I found that doing it regardless fixes an issue where other tabs and the overview may not be showing the pinned button when a course is in fact pinned, and vice versa, if it was pinned in the other view.  So I found this helps make sure the card's state is up-to-date everywhere you see the card, not jsut the view in which you changed it

src/d2l-my-courses-container.js
AllanKerr: Yep, that sounds good to me!

src/d2l-all-courses.js
svanherk: Ok, updated

src/d2l-my-courses-content.js
```diff
@@ -272,14 +267,39 @@ class MyCoursesContent extends mixinBehaviors([
...
+
+		const changedEnrollmentId = newValue.orgUnitId && this._orgUnitIdMap[newValue.orgUnitId];
+		if (changedEnrollmentId) {
+			updateEntity(changedEnrollmentId, this.token);
+		}
+
+		if (this.tabSearchActions.length === 0 || this._thisTabSelected) {
+			// Only want to move pinned/unpinned enrollment if it exists in the panel
+			if (!changedEnrollmentId) {
+				this._refetchEnrollments();
```
klhuff, line 291: What happens if your current search options don't promote pins, and then the pin is done through the hamburger menu? I assume nothing bad happens, but we refresh the widget/enrollments anyway?

svanherk: In the All Courses overlay, or the main widget view?  Does the main widget view have a sort option that doesn't promote pins?  I _thought_ it did for both < 50 and >50 enrollments...

klhuff: Good point. I wasn't really thinking about how you can't see/use the waffle when you have the all courses overlay open.

klhuff: But, I was originally thinking about the overlay, and what happens when you have the enrollments sorted by Course Name/ Code.

svanherk: Yeah, so if we theoretically could change the selector when the overlay was opened we should still be good for now... the overlay doesn't refetch enrollments, it just sets a flag to refetch next time it reloads unless its the pinned tab (https://github.com/Brightspace/d2l-my-courses-ui/pull/824/files#diff-7d9e8e373cdb89087fc53450617b7257R412).

But I'd been experimenting with it acting a bit more like the widget view.  If we went to that, then I think we'd need to be careful about the promoted pinned stuff.  Maybe another reason to leave as is.

PR #825, opened 2020-06-08
Setup auto-versioning

svanherk:
Auto-updating the version via `frau-ci` so that the version in `package.json` is correct. This should let BSI pull it in with a new version of NPM.


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-- "11"
+- '11'
```
svanherk, line 3: Going to leave this for now, and try updating in a separate PR

PR #827, opened 2020-06-09
US114862 - Move All Courses Overlay out of Tabs and into the Container

svanherk:
Moving `d2l-all-courses` out of `d2l-my-courses-content` and into `d2l-my-courses-container`.  This means that instead of each tab getting their own all-courses overlay, there will only be one.

This is needed to make sure the overlay doesn't gracelessly force-close (messing up focus and still having a ghost version covering the screen) when the pinned tab is removed (therefore removing its all courses overlay).  It also makes things a little cleaner and will help cut down on the number of duplicate calls asking for the same thing with a different cache parameter.

### DIFF
--- a/src/d2l-my-courses-content.js
+++ b/src/d2l-my-courses-content.js
@@ -8,7 +8,6 @@ import '@brightspace-ui/core/components/link/link.js';
 import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import './d2l-alert-behavior.js';
-import './d2l-all-courses.js';
 import './d2l-my-courses-card-grid.js';
 import './d2l-utility-behavior.js';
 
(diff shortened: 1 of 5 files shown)


src/d2l-my-courses-content.js
```diff
@@ -629,22 +591,15 @@ class MyCoursesContent extends mixinBehaviors([
 		return enrollmentsLength > 0 ? `${viewAllCourses} (${count})` : viewAllCourses;
 	}
 	_openAllCoursesView(e) {
-		const allCourses = this.shadowRoot.querySelector('d2l-all-courses');
```
AllanKerr, line 632: I know it would likely require substantial refactoring but thoughts on also moving the link up? It feels a bit off having the link to open it be in the content but not the actual overlay that gets opened.

src/d2l-my-courses-content.js
```diff
@@ -629,22 +591,15 @@ class MyCoursesContent extends mixinBehaviors([
...
-		allCourses.tabSearchActions = this.tabSearchActions;
-		allCourses.open();
-
-		this.showImageError = false; // Clear image error when opening and closing the all courses overlay
```
AllanKerr, line 636: I assume we decided we don't want to clear the error when we open all courses or am I missing something here?

PR #829, opened 2020-06-11
US114858 - More prep for hiding pinned tab

svanherk:
I'll add some commentary below, but this PR contains a bunch of "fixes" that needed to be made once I tried actually hiding the pinned tab.  They could all be done proactively and I'll add a little context to each.  `dom-repeat` is making adding/removing tabs a nightmare, but _hopefully_ this is everything.

Actual pinned tab PR coming shortly!

### DIFF
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -395,12 +395,9 @@ class MyCoursesContainer extends mixinBehaviors([
 		// Whenever the selected tab changes, update tabSearchActions so
 		// All Courses will have the same tab selected when it opens
 		this._tabSearchActions = this._tabSearchActions.map((action) => {
-			return {
-				name: action.name,
-				title: action.title,
-				selected: action.name === e.detail.tabId,
-				enrollmentsSearchAction: action.enrollmentsSearchAction
-			};
+			return Object.assign({}, action, {
+				selected: action.name === e.detail.tabId
+			});
 		});
 
 	}
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -124,7 +124,6 @@ class MyCoursesContainer extends mixinBehaviors([
 								token="[[token]]"
 								enrollments-search-action="[[item.enrollmentsSearchAction]]"
 								org-unit-type-ids="[[orgUnitTypeIds]]"
-								tab-search-actions="[[_tabSearchActions]]"
 								tab-search-type="[[_tabSearchType]]"
 								update-user-settings-action="[[_updateUserSettingsAction]]">
 							</d2l-my-courses-content>
@@ -134,6 +133,7 @@ class MyCoursesContainer extends mixinBehaviors([
 			</template>
 			<template is="dom-if" if="[[!_showGroupByTabs]]">
 				<d2l-my-courses-content
+					no-tabs
 					on-d2l-my-courses-content-open-all-courses="_openAllCoursesOverlay"
 					presentation-url="[[_presentationUrl]]"
 					show-image-error="[[_showImageError]]"
@@ -262,9 +262,12 @@ class MyCoursesContainer extends mixinBehaviors([
 		return this.shadowRoot.querySelector('d2l-all-courses');
 	}
 
-	_openAllCoursesOverlay(e) {
+	_openAllCoursesOverlay() {
 		const allCourses = this._getAllCoursesComponent();
-		allCourses.tabSearchActions = e.detail.tabSearchActions;
+		// Make a deep copy of the tabSearchActions so we can modify it to remember All Courses's last selected tab
+		// without affecting the selected tab on the main widget view
+		const TabSearchActionsCopy = this._tabSearchActions.map(action => Object.assign({}, action));
+		allCourses.tabSearchActions = TabSearchActionsCopy;
 		allCourses.open();
 
 		this._showImageError = false; // Clear image error when opening and closing the all courses overlay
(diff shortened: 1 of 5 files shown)


src/d2l-my-courses-container.js
```diff
@@ -262,9 +262,12 @@ class MyCoursesContainer extends mixinBehaviors([
 		return this.shadowRoot.querySelector('d2l-all-courses');
 	}
 
-	_openAllCoursesOverlay(e) {
+	_openAllCoursesOverlay() {
 		const allCourses = this._getAllCoursesComponent();
-		allCourses.tabSearchActions = e.detail.tabSearchActions;
+		// Make a deep copy of the tabSearchActions so we can modify it to remember All Courses's last selected tab
+		// without affecting the selected tab on the main widget view
+		const TabSearchActionsCopy = this._tabSearchActions.map(action => Object.assign({}, action));
```
svanherk, line 269: This being its own copy that we can manipulate isn't important yet, but will be in the next PR.  If we modify the objects in AllCourses's `tabSearchActions`, we _do not_ want those changes coming back to the main widget view.

src/d2l-my-courses-container.js
```diff
@@ -389,6 +392,17 @@ class MyCoursesContainer extends mixinBehaviors([
 	}
 	_tabSelectedChanged(e) {
 		this._currentTabId = `panel-${e.detail.tabId}`;
+		// Whenever the selected tab changes, update tabSearchActions so
```
svanherk, line 395: Moved from `d2l-my-courses-content` to here

src/d2l-my-courses-content.js
```diff
@@ -591,11 +581,7 @@ class MyCoursesContent extends mixinBehaviors([
 		return enrollmentsLength > 0 ? `${viewAllCourses} (${count})` : viewAllCourses;
 	}
 	_openAllCoursesView(e) {
-		this.dispatchEvent(new CustomEvent('d2l-my-courses-content-open-all-courses', {
-			detail: {
-				tabSearchActions: this.tabSearchActions
```
svanherk, line 596: No longer needed, the container will just use the copy it has, which was updated every time the tab was changed to know which tab should be selected on opening

src/d2l-my-courses-content.js
```diff
@@ -727,6 +713,7 @@ class MyCoursesContent extends mixinBehaviors([
 		this._existingEnrollmentsMap = {};
 		this._enrollments = [];
 		this._numberOfEnrollments = 0;
+		this._nextEnrollmentEntityUrl = null;
```
svanherk, line 716: This is an issue that only seems to show itself once we're adding/removing tabs.  A race condition causes the refetchEnrollments code to happen at the same time as the code to determine if we should be removing closed courses.  If the latter uses the old `this._nextEnrollmentEntityUrl` from a previous tab-panel (because dom-repeat re-uses them) they can get the next set of courses from the previous panel instead.  If we're resetting enrollments, we should clear this and then reset it once we get the results back.

test/d2l-my-courses-container/d2l-my-courses-container.js
```diff
@@ -196,17 +196,6 @@ describe('d2l-my-courses', () => {
 		});
 	});
 
-	it('should have updated _currentTabId property based on the event', () => {
```
svanherk, line 199: Moved to `Events` and updated

test/d2l-my-courses-container/d2l-my-courses-container.js
```diff
@@ -442,7 +431,7 @@ describe('d2l-my-courses', () => {
 			});
 		});
 
-		describe('d2l-my-courses-content-open-all-courses', () => {
+		describe('d2l-simple-overlay-closed', () => {
```
svanherk, line 434: I somehow just totally messed up these two titles lol

src/d2l-my-courses-container.js
```diff
@@ -389,6 +392,17 @@ class MyCoursesContainer extends mixinBehaviors([
 	}
 	_tabSelectedChanged(e) {
 		this._currentTabId = `panel-${e.detail.tabId}`;
+		// Whenever the selected tab changes, update tabSearchActions so
+		// All Courses will have the same tab selected when it opens
+		this._tabSearchActions = this._tabSearchActions.map((action) => {
+			return {
+				name: action.name,
+				title: action.title,
+				selected: action.name === e.detail.tabId,
```
AllanKerr, line 401: It feels a bit dangerous to be modifying actions client-side like this; although, it likely does simplify it a bit. If we are going to do it this way we should at least use `Object.assign` or a spread to ensure all properties get copied over in case someone adds new properties to the action in the future.

svanherk: Good point, I can do something closer to https://github.com/Brightspace/d2l-my-courses-ui/pull/829/files#diff-79c7c139ccd6d5d799b4998ccaba34a9R269

src/d2l-my-courses-container.js
AllanKerr, line 269: nit: Should `TabSearchActionsCopy` be capitalized?

src/d2l-my-courses-container.js
svanherk: So looking at this again, what it's actually doing is pulling out some key values it wants (`name`, `title`), adding `selected`, and then storing the whole action in the `enrollmentsSearchAction` value.  It first creates this here: https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-my-courses-container.js#L330-L339

So I think we're actually good, we'll get all the action properties in `enrollmentsSearchAction` (what is actually used to perform actions), and we're just wrapping that with a few key characteristics.  If I use `Object.assign`, I basically just make a duplicate of the action which contains the action.

svanherk: Discussed offline, there's still value in doing it here if we ever change the way we define tabSearchActions

PR #830, opened 2020-06-11
US114858 - Hide and Show the Pinned Tab

svanherk:
A few things going on in this PR:
- Determining whether the pinned tab should be shown or hidden on load
  - We try to get the previous state from local browser cache
  - If we can't get this, we won't add the pinned tab yet
  - After making this choice, we asynchronously check that we were correct and add/remove if we were not
- Determining whether to hide/show the pinned tab as the user pins and unpins courses
- Keeping the overlay and main widget tabs in sync
- If there are no pinned or semester/department/role tabs, hide the all tab (no work here to do this, that is done for us in the tabs component)

I need to write new tests, but wanted to get this up early for feedback since we're down to the wire.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -579,8 +579,10 @@ class AllCourses extends mixinBehaviors([
 		let tabAction;
 		for (let i = 0; i < this.tabSearchActions.length; i++) {
 			if (this.tabSearchActions[i].name === actionName) {
+				this.tabSearchActions[i].selected = true;
 				tabAction = this.tabSearchActions[i];
-				break;
+			} else {
+				this.tabSearchActions[i].selected = false;
 			}
 		}
 
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -471,7 +471,7 @@ class MyCoursesContainer extends mixinBehaviors([
 	}
 	_verifyPinnedTab(pinnedTabAction) {
 		let pinnedSearchUrl = this.createActionUrl(pinnedTabAction);
-		if (pinnedSearchUrl.indexOf('?') > -1) {
+		if (pinnedSearchUrl && pinnedSearchUrl.indexOf('?') > -1) {
 			// pinnedSearchUrl already has some query params, append ours
 			pinnedSearchUrl += `&bustCache=${Math.random()}`;
 		} else {
--- a/src/d2l-my-courses-container.js
+++ b/src/d2l-my-courses-container.js
@@ -388,6 +388,16 @@ class MyCoursesContainer extends mixinBehaviors([
 			});
 
 			this._getAllCoursesComponent().courseEnrollmentChanged(this._changedCourseEnrollment);
+
+			if (this._pinnedTabAction) {
+				if (e.detail.isPinned) {
+					// We need to check if we need to add the pinned tab
+					this._addPinnedTab();
+				} else {
+					// We need to check if we've removed the last pinned course and if so, remove the pinned tab
+					this._verifyPinnedTab(this._pinnedTabAction);
+				}
+			}
 		}
 	}
 	_tabSelectedChanged(e) {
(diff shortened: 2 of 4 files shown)


src/d2l-all-courses.js
```diff
@@ -579,8 +579,10 @@ class AllCourses extends mixinBehaviors([
 		let tabAction;
 		for (let i = 0; i < this.tabSearchActions.length; i++) {
 			if (this.tabSearchActions[i].name === actionName) {
```
svanherk, line 581: We need to update All Courses's copy of `tabSearchActions` with the correct `selected` values because adding/removing the pinned tab cause a full re-render.  If we don't have the correct value set, the selected tab will jump to whatever one was selected on opening.

src/d2l-my-courses-container.js
```diff
@@ -444,6 +469,73 @@ class MyCoursesContainer extends mixinBehaviors([
...
+		});
+	}
+	_addPinnedTab() {
+		const pinnedTabIndex = this._tabSearchActions.findIndex(action => action.name === this._pinnedTabAction.name);
+		if (pinnedTabIndex === -1) {
+			this.splice('_tabSearchActions', 1, 0, this._getPinnedTabAction());
+
+			const allCourses = this._getAllCoursesComponent();
+			if (allCourses.tabSearchActions && allCourses.tabSearchActions.length > 0) {
+				allCourses.splice('tabSearchActions', 1, 0, this._getPinnedTabAction());
```
svanherk, line 496: Needed to use `splice` here to notify Polymer; just setting `allCourses`'s actions directly did not cause a re-render.

src/d2l-my-courses-container.js
```diff
@@ -444,6 +469,73 @@ class MyCoursesContainer extends mixinBehaviors([
...
+			}
+
+			const contents = this.shadowRoot.querySelectorAll('d2l-my-courses-content');
+			contents.forEach(content => {
+				content.requestRefresh();
+			});
+		}
+		this._trySetItemLocalStorage('myCourses.pinnedTab', {'previouslyShown': false});
+	}
+	_tryGetItemLocalStorage(itemName) {
```
svanherk, line 523: Took these two functions directly from how we did this in org tabs (except there we used session storage)

src/d2l-my-courses-content.js
```diff
@@ -444,7 +450,7 @@ class MyCoursesContent extends mixinBehaviors([
 
 		if (this._isRefetchNeeded) {
 			this._handleEnrollmentsRefetch();
-		} else if (this._numberOfEnrollments === 0) {
+		} else if (this._numberOfEnrollments === 0 && !this._rootTabSelected) {
```
svanherk, line 453: This is a slightly unrelated fix - when the pinned tab is empty but has already been loaded (`this._rootTabSelected = true;`), we don't want re-fetch for no reason.  This doesn't matter now that the tab gets removed, but other empty tabs would also benefit from this.

src/d2l-my-courses-content.js
```diff
@@ -521,7 +527,7 @@ class MyCoursesContent extends mixinBehaviors([
 		};
 		let enrollmentsSearchUrl = this.createActionUrl(this.enrollmentsSearchAction, query);
 
-		if (bustCache) {
+		if (bustCache || this._isPinnedTab) {
```
svanherk, line 530: Because the pinned tab is added and removed, it goes through the "initial load" code path multiple times.  If we don't bust the cache, we get old data.  We always want up-to-date data for the pinned tab

src/d2l-my-courses-container.js
```diff
@@ -444,6 +469,73 @@ class MyCoursesContainer extends mixinBehaviors([
...
+			this.splice('_tabSearchActions', 1, 0, this._getPinnedTabAction());
+
+			const allCourses = this._getAllCoursesComponent();
+			if (allCourses.tabSearchActions && allCourses.tabSearchActions.length > 0) {
+				allCourses.splice('tabSearchActions', 1, 0, this._getPinnedTabAction());
+			}
+
+			const contents = this.shadowRoot.querySelectorAll('d2l-my-courses-content');
+			contents.forEach(content => {
+				content.requestRefresh();
```
AllanKerr, line 501: Why do we need to refresh all the contents? Is this to ensure pinned items get brought to the top?

src/d2l-my-courses-container.js
```diff
@@ -444,6 +469,73 @@ class MyCoursesContainer extends mixinBehaviors([
 
 		return this._setPromotedSearchEntity(this.promotedSearches);
 	}
+	_verifyPinnedTab(pinnedTabAction) {
+		let enrollmentsSearchUrl = this.createActionUrl(pinnedTabAction);
+		enrollmentsSearchUrl += `&bustCache=${Math.random()}`;
```
AllanKerr, line 474: I assume we're always guaranteed to have other query params here?

src/d2l-my-courses-container.js
svanherk: No, it's because the Polymer `dom-repeat` is going to try to reuse DOM nodes and the tab-panels will have the content of the tab before/after it (depending which way we shifted).  This is definitely not great, and I think we can make a better experience in 20.20.8.  People won't run into this a lot unless they are constantly pinning and unpinning their one pinned course and then traversing the rest of the widget, which they've already traversed through.

src/d2l-my-courses-container.js
svanherk: Hmmm... I'll make this smarter.  Pretty sure I copied this months ago from another area that I have since made smarter after you pointed it out 😛

PR #832, opened 2020-06-23
[20.20.7] DE39576 - DE39577 - Fix Missing Filter Terms

svanherk:
Because the All Courses component now exists from the get-go (instead of being dynamically created later) some langterms were showing up as blank because they weren't initialized, or because they were computed too early (before we'd even determined the language).

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -395,6 +395,7 @@ class AllCourses extends mixinBehaviors([
 		super.connectedCallback();
 
 		this.addEventListener('d2l-tab-panel-selected', this._onTabSelected);
+		this._filterText = this.localize('filtering.filter');
 	}
 
 	/*
--- a/src/search-filter/d2l-filter-menu.js
+++ b/src/search-filter/d2l-filter-menu.js
@@ -66,26 +66,6 @@ class FilterMenu extends mixinBehaviors([
 				value: false,
 				computed: '_computeHasFilters(_departmentFilters.length, _semesterFilters.length, _roleFiltersCount)'
 			},
-			_semestersTabText: {
-				type: String,
-				computed: '_computeTabText(filterStandardSemesterName, _semesterFilters.length)'
-			},
-			_departmentsTabText: {
-				type: String,
-				computed: '_computeTabText(filterStandardDepartmentName, _departmentFilters.length)'
-			},
-			_rolesTabText: {
-				type: String,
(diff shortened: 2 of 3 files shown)


src/d2l-all-courses.js
```diff
@@ -395,6 +395,7 @@ class AllCourses extends mixinBehaviors([
 		super.connectedCallback();
 
 		this.addEventListener('d2l-tab-panel-selected', this._onTabSelected);
+		this._filterText = this.localize('filtering.filter');
```
svanherk, line 398: Setting an initial value

src/search-filter/d2l-filter-menu.js
```diff
@@ -169,27 +149,56 @@ class FilterMenu extends mixinBehaviors([
...
 
 		<iron-pages attr-for-selected="data-tab-name" selected="semesters" fallback-selection="semesters">
-			<d2l-filter-menu-tab id="semestersTab" data-tab-name="semesters" aria-labelledby="semestersTabButton" menu-label-text="[[filterStandardSemesterName]]" no-filters-text="[[localize('filtering.noSemesters', 'semester', filterStandardSemesterName)]]" search-action="[[_searchSemestersAction]]" search-placeholder-text="[[_semestersSearchPlaceholderText]]" selected-filters="{{_semesterFilters}}" hidden$="[[_semestersTabHidden]]">
+			<d2l-filter-menu-tab
+				id="semestersTab"
+				data-tab-name="semesters"
+				aria-labelledby="semestersTabButton"
+				menu-label-text="[[filterStandardSemesterName]]"
+				no-filters-text="[[localize('filtering.noSemesters', 'semester', filterStandardSemesterName)]]" search-action="[[_searchSemestersAction]]"
+				search-placeholder-text="[[localize('filtering.searchBy', 'filter', filterStandardSemesterName)]]"
```
klhuff, line 178: Nice. It's not clear to me why some of these were inline and some were using properties before.

svanherk: You'll almost always see them inline, which means they are re-calculated every time the component is re-rendered.  Having them as computed properties would cut down on some of that work, but will almost never work because they're calculated too early.  It was working before because we only created the component after hovering on the link to open it, which was late enough to calculate it.

We could fix the computed properties by adding `language` as a parameter to all of them (telling them to recalculate if that changes), but that's a properties of `app-localize-behavior` and it feels bad to use that in this code.  It's existence is behind the scenes and shouldn't matter to us

PR #837, opened 2020-07-22
US118784 - Cleanup alert behavior

svanherk:
The Polymer `d2l-alert-behavior` was only being used in one place after all the refactoring, so I removed it and tried to combine and simplify the alert logic.


src/d2l-my-courses-content.js
```diff
@@ -195,18 +196,16 @@ class MyCoursesContent extends mixinBehaviors([
 				presentation-url="[[presentationUrl]]"
 				widget-view>
 
-				<d2l-alert hidden$="[[!_hasOnlyPastCourses]]" type="call-to-action">
-					[[localize('onlyPastCoursesMessage')]]
-				</d2l-alert>
 				<d2l-alert id="imageErrorAlert" hidden$="[[!showImageError]]" type="warning">
 					[[localize('error.settingImage')]]
 				</d2l-alert>
+				<d2l-alert id="courseInfoAlert" hidden$="[[!_courseInfoAlertText]]" type="call-to-action">
```
svanherk, line 202: From my testing, I couldn't get these two alerts to show up at the same time.  Once the new courses you've been enrolled in are `closed`, the `d2l-enrollment-new` event is not sent for those courses.  But the old code allowed for them to appear at the same time, so I've kept them separate here for now

src/d2l-my-courses-content.js
```diff
@@ -195,18 +196,16 @@ class MyCoursesContent extends mixinBehaviors([
 				presentation-url="[[presentationUrl]]"
 				widget-view>
 
-				<d2l-alert hidden$="[[!_hasOnlyPastCourses]]" type="call-to-action">
-					[[localize('onlyPastCoursesMessage')]]
-				</d2l-alert>
 				<d2l-alert id="imageErrorAlert" hidden$="[[!showImageError]]" type="warning">
```
svanherk, line 199: This alert will eventually live in the `d2l-my-courses-card-grid` component, but I need to move the tabs into that component first.

PR #842, opened 2020-08-07
Add a d2l-all-courses-close event

svanherk:
Because of https://github.com/Brightspace/simple-overlay/blob/master/d2l-simple-overlay.js#L109, listening to the `d2l-simple-overlay-closed` event on the `d2l-all-courses` component meant `_onAllCoursesOverlayClosed` was being called over ten times for one overlay close.  To stop this, I'm now sending a custom `d2l-all-courses-close` event, which feels better anyways - `d2l-my-courses-container` shouldn't know that `d2l-all-courses` contains a `d2l-simple-overlay` anyways.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -553,7 +553,7 @@ class AllCourses extends mixinBehaviors([
 		}
 	}
 
-	_onSimpleOverlayClosed() {
+	_onSimpleOverlayClosed(e) {
 		if (this._enrollmentsSearchAction && this._enrollmentsSearchAction.hasFieldByName) {
 			if (this._enrollmentsSearchAction.hasFieldByName('search')) {
 				this._enrollmentsSearchAction.getFieldByName('search').value = '';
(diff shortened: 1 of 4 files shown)


src/d2l-my-courses-container.js
```diff
@@ -144,7 +144,7 @@ class MyCoursesContainer extends mixinBehaviors([
 			</template>
 
 			<d2l-all-courses
-				on-d2l-simple-overlay-closed="_onAllCoursesOverlayClosed"
+				on-d2l-all-courses-close="_onAllCoursesOverlayClosed"
```
klhuff, line 147: NIT: should we also change the handler method name?

klhuff: On second thought. This seems perfectly fine. Ignore my comment.

svanherk: Good call, I like when they match so you it's obvious

src/d2l-all-courses.js
```diff
@@ -570,6 +570,9 @@ class AllCourses extends mixinBehaviors([
 		this.$.filterMenu.clearFilters();
 		this._filterText = this.localize('filtering.filter');
 		this._resetSortDropdown();
+
+		e.stopPropagation();
```
AllanKerr, line 574: Do we need to stop propagation here since we're now listening for `d2l-all-courses-close`? Just not sure how far it bubbles down and who else might have decided they want to use it given that its composed.

svanherk: You're right that we don't have to - I did just because the event is so noisy and it didn't seem like anything else needed to use it.  But that's from my testing, not from checking every component, so probably is safer to leave it propagating.

PR #843, opened 2020-08-07
Us114855 - Use shared d2l-sort-by-dropdown in d2l-all-courses

svanherk:
This is blocked on https://github.com/BrightspaceHypermediaComponents/common/pull/36 and https://github.com/BrightspaceHypermediaComponents/activities/pull/1027 - otherwise BSI will break if someone tries to take the latest my-courses version (I plan on just doing a minor release).

This swaps out our homemade sort dropdown with the shared one in `facet-filter-sort`.  I'll be doing the same for the filter dropdown next.  Until then the openers will look a bit odd being different, but hopefully for only a few days.

image

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -8,14 +8,13 @@ import '@brightspace-ui/core/components/alert/alert.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/dropdown/dropdown.js';
 import '@brightspace-ui/core/components/dropdown/dropdown-content.js';
-import '@brightspace-ui/core/components/dropdown/dropdown-menu.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/link/link.js';
 import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
-import '@brightspace-ui/core/components/menu/menu.js';
-import '@brightspace-ui/core/components/menu/menu-item-radio.js';
 import '@brightspace-ui/core/components/tabs/tabs.js';
 import '@brightspace-ui/core/components/tabs/tab-panel.js';
+import 'd2l-facet-filter-sort/components/d2l-sort-by-dropdown/d2l-sort-by-dropdown.js';
+import 'd2l-facet-filter-sort/components/d2l-sort-by-dropdown/d2l-sort-by-dropdown-option.js';
 import 'd2l-organization-hm-behavior/d2l-organization-hm-behavior.js';
 import 'd2l-simple-overlay/d2l-simple-overlay.js';
 import './d2l-my-courses-card-grid.js';
@@ -234,6 +233,13 @@ class AllCourses extends mixinBehaviors([
 						margin-top: 5px;
 					}
 				}
+				d2l-sort-by-dropdown {
+					margin-left: 1rem;
+				}
+				:host(:dir(rtl)) d2l-sort-by-dropdown {
+					margin-left: 0;
+					margin-right: 1rem;
+				}
 				.dropdown-opener-text {
 					font-size: 0.95rem;
 					font-family: Lato;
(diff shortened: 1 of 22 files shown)


src/d2l-all-courses.js
```diff
@@ -142,37 +141,31 @@ class AllCourses extends mixinBehaviors([
 						{
 							name: 'Default',
 							action: 'Current',
-							langterm: 'sorting.sortDefault',
```
svanherk, line 145: We used to have to set the opener text ourselves and pulled the value from here - the shared component does this for us.

src/d2l-all-courses.js
```diff
@@ -569,7 +544,7 @@ class AllCourses extends mixinBehaviors([
 		this._clearSearchWidget();
 		this.$.filterMenu.clearFilters();
 		this._filterText = this.localize('filtering.filter');
-		this._resetSortDropdown();
```
svanherk, line 572: Removed this function because it was turning into one line and only used in one place anyways.

src/d2l-all-courses.js
```diff
@@ -635,18 +610,6 @@ class AllCourses extends mixinBehaviors([
 
 		const searchAction = myEnrollmentsEntity.getActionByName(Actions.enrollments.searchMyEnrollments);
 		this._enrollmentsSearchAction = searchAction;
-
-		if (searchAction && searchAction.hasFieldByName('sort')) {
```
svanherk, line 639: Discussed this in the my courses slack channel - we don't think this is actually needed for anything.  If we get to this code path because the user made a sort dropdown selection, then it will already match.  There isn't any other way to change the sort because it's cleared on overlay closing (so it's not like we would need to reset it) and kept across tab changes.

src/lang/en.js
```diff
@@ -33,11 +33,11 @@ export default {
 	"recentSearches": "Recent Searches",
 	"search.searchCourses": "Search Courses",
 	"sorting.sortBy": "Sort By",
-	"sorting.sortCourseCode": "Sort: Course Code",
```
svanherk, line 36: The shared component already adds the "Sort: " to the opener, so if we left it like this we'd get "Sort: Sort: Default".  I tried to cleanup all the languages with my best guess, but serge should trigger an update to these langterms and the translators can fix them if there's an issue.

klhuff: Yeah. I would not be surprised at all if the concatenation did not work well in other languages.

src/d2l-all-courses.js
```diff
@@ -569,7 +544,7 @@ class AllCourses extends mixinBehaviors([
 		this._clearSearchWidget();
 		this.$.filterMenu.clearFilters();
 		this._filterText = this.localize('filtering.filter');
-		this._resetSortDropdown();
+		this.shadowRoot.querySelector(`d2l-sort-by-dropdown-option[value=${this._sortMap[0].name}]`).click();
```
AllanKerr, line 547: It feels a bit weird to handle reset by faking a UI interaction. Is it worth adding a `reset` method to `d2l-sort-by-dropdown`?

svanherk: Yeah it feels odd, but I think I'd have to make `d2l-sort-by-dropdown` do the same thing, or duplicate the code here: https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item-radio-mixin.js#L16

The problem, from what I can tell, is that the code above only runs if the `d2l-menu-item-select` is fired: https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item-radio-mixin.js#L13

Which only happens on a click or key event: https://github.com/BrightspaceUI/core/blob/master/components/menu/menu-item-mixin.js#L89 (only called in `_onClick` and `onKeyDown`.

This is what I was asking about in the thread in #gaudi-dev... is it best to simulate a UI interaction, fake it by raising my own menu-select event, or just have the dropdown implement its own clearing logic (either duplicate, or call its mixin's private method - something I jsut undid somewhere else).

I don't like any of them, really.

AllanKerr: It seems a bit odd that setting `selected` doesn't correctly trigger the value to be updated but I don't see a good fix so I'm okay leaving as is.

PR #844, opened 2020-08-13
Use subtle button opener for the filter in the All Courses overlay

svanherk:
Swapping the custom filter for `d2l-filter-dropdown` won't be making it in this release, so I'm just updating the opener to be a subtle button like the new `d2l-sort-by-dropdown`, to keep things looking nice for the September release.  Also lets me do some nice css cleanup earlier.

Before:
image
image

After:
image
image

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -5,7 +5,6 @@ Polymer-based web component for the all courses overlay.
 
 import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
 import '@brightspace-ui/core/components/alert/alert.js';
-import '@brightspace-ui/core/components/colors/colors.js';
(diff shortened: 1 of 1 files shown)


src/d2l-all-courses.js
```diff
@@ -227,42 +221,15 @@ class AllCourses extends mixinBehaviors([
 						margin-top: 5px;
 					}
 				}
+				d2l-dropdown-button-subtle,
 				d2l-sort-by-dropdown {
-					margin-left: 1rem;
+					margin-left: 0.5rem;
```
svanherk, line 226: This might change, still waiting to hear back from Jeff G

svanherk: Jeff is good with this

PR #850, opened 2020-09-08
Some more filter prep

svanherk:
These are some changes that came out of the filter work, but can happen ahead of time.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -426,7 +426,7 @@ class AllCourses extends mixinBehaviors([
 	_onFilterDropdownClose() {
 		let text;
 		const totalFilterCount = this._filterCounts.departments + this._filterCounts.semesters + this._filterCounts.roles;
-	
+
 		if (totalFilterCount === 0) {
 			text = this.localize('filtering.filter');
 		} else if (totalFilterCount === 1) {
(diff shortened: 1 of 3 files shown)


src/d2l-all-courses.js
```diff
@@ -109,28 +121,6 @@ class AllCourses extends mixinBehaviors([
 				type: Boolean,
 				value: false
 			},
-			// Number of filters currently selected; used to change opener text when menu closes
-			_totalFilterCount: {
-				type: Number,
-				value: 0
-			},
```
svanherk, line 116: All the properties changes here are just rearranging the order, except for this one which is being deleted.

src/d2l-all-courses.js
```diff
@@ -433,17 +421,18 @@ class AllCourses extends mixinBehaviors([
 	_onFilterChanged(e) {
 		this._searchUrl = this._appendOrUpdateBustCacheQueryString(e.detail.url);
 		this._filterCounts = e.detail.filterCounts;
-		this._totalFilterCount = this._filterCounts.departments + this._filterCounts.semesters + this._filterCounts.roles;
```
svanherk, line 436: Only one function is going to need this count, so we'll just calculate it when we need it.  For now there are two, but `_onFilterDropdownClose` completely goes away with the new filter.

PR #852, opened 2020-09-10
US114856 - DE36186 - Use New Filter

svanherk:
Hooking up the new filter in `d2l-all-courses` and removing the old one.  I'll do the large cleanup of old filter files, langterms and tests in a follow-up PR.

Still to finish:
- [x] Tests

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -5,8 +5,6 @@ Polymer-based web component for the all courses overlay.
 
 import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
 import '@brightspace-ui/core/components/alert/alert.js';
-import '@brightspace-ui/core/components/dropdown/dropdown-button-subtle.js';
-import '@brightspace-ui/core/components/dropdown/dropdown-content.js';
 import '@brightspace-ui/core/components/link/link.js';
 import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 import '@brightspace-ui/core/components/tabs/tabs.js';
@@ -16,7 +14,6 @@ import 'd2l-facet-filter-sort/components/d2l-sort-by-dropdown/d2l-sort-by-dropdo
 import 'd2l-organization-hm-behavior/d2l-organization-hm-behavior.js';
 import 'd2l-simple-overlay/d2l-simple-overlay.js';
 import './d2l-my-courses-card-grid.js';
-import './search-filter/d2l-filter-menu.js';
 import './search-filter/d2l-my-courses-filter.js';
 import './search-filter/d2l-search-widget-custom.js';
 import { Actions, Classes } from 'd2l-hypermedia-constants';
@@ -82,8 +79,6 @@ class AllCourses extends mixinBehaviors([
 					};
 				}
 			},
-			// Filter dropdown opener text
-			_filterText: String,
 			_hasEnrollmentsChanged: {
 				type: Boolean,
 				value: false
@@ -215,11 +210,11 @@ class AllCourses extends mixinBehaviors([
 						margin-top: 5px;
 					}
 				}
-				d2l-dropdown-button-subtle,
+				d2l-my-courses-filter,
 				d2l-sort-by-dropdown {
 					margin-left: 0.5rem;
 				}
-				:host(:dir(rtl)) d2l-dropdown-button-subtle,
+				:host(:dir(rtl)) d2l-my-courses-filter,
 				:host(:dir(rtl)) d2l-sort-by-dropdown {
 					margin-left: 0;
 					margin-right: 0.5rem;
@@ -253,26 +248,6 @@ class AllCourses extends mixinBehaviors([
 							</d2l-search-widget-custom>
 
 							<div id="filterAndSort">
-								<d2l-dropdown-button-subtle text="[[_filterText]]">
-									<d2l-dropdown-content
-										id="filterDropdownContent"
-										on-d2l-dropdown-open="_onFilterDropdownOpen"
-										on-d2l-dropdown-close="_onFilterDropdownClose"
-										no-padding
-										min-width="350"
-										render-content>
-
-										<d2l-filter-menu
-											id="filterMenu"
-											on-d2l-filter-menu-change="_onFilterChanged"
-											tab-search-type="[[tabSearchType]]"
-											org-unit-type-ids="[[orgUnitTypeIds]]"
-											my-enrollments-entity="[[_myEnrollmentsEntity]]"
-											filter-standard-semester-name="[[filterStandardSemesterName]]"
-											filter-standard-department-name="[[filterStandardDepartmentName]]">
-										</d2l-filter-menu>
-									</d2l-dropdown-content>
-								</d2l-dropdown-button-subtle>
 								<d2l-my-courses-filter
 									on-d2l-my-courses-filter-change="_onFilterChange"
 									on-d2l-my-courses-filter-clear="_onFilterClear"
@@ -329,11 +304,6 @@ class AllCourses extends mixinBehaviors([
 				<d2l-loading-spinner hidden$="[[_showContent]]" size="100">
 				</d2l-loading-spinner>
 			</d2l-simple-overlay>`;
-	}
-
-	connectedCallback() {
-		super.connectedCallback();
-		this._filterText = this.localize('filtering.filter');
 	}
 
 	/*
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.js
```diff
@@ -418,30 +399,6 @@ class AllCourses extends mixinBehaviors([
...
-	_onFilterDropdownClose() {
-		let text;
-		const totalFilterCount = this._filterCounts.departments + this._filterCounts.semesters + this._filterCounts.roles;
-
-		if (totalFilterCount === 0) {
-			text = this.localize('filtering.filter');
-		} else if (totalFilterCount === 1) {
-			text = this.localize('filtering.filterSingle');
-		} else {
-			text = this.localize('filtering.filterMultiple', 'num', totalFilterCount);
```
svanherk, line 435: Going to cleanup these langterms in the next PR with a bunch more

src/d2l-all-courses.js
```diff
@@ -468,6 +425,95 @@ class AllCourses extends mixinBehaviors([
...
+
+		this._filterCounts = {
+			departments: departmentFilters ? departmentFilters.selectedOptions.length : 0,
+			semesters: semesterFilters ? semesterFilters.selectedOptions.length : 0,
+			roles: roleFilters ? roleFilters.selectedOptions.length : 0
+		};
+
+		if (roleFilters && e.detail.categoryChanged === 'roles') {
+			this._handleRolesFilterChange(roleFilters.selectedOptions);
+		} else if (semesterFilters && departmentFilters) {
```
AllanKerr, line 441: What if we have only `semesterFilters` or `departmentFilters` but not both? I assume we'd still want this to run?

svanherk: Hmmm... so we always have both, _unless_ something went wrong and a search action was missing.  But this check does seem overly worried, and I can check for what filter changed

src/d2l-all-courses.js
```diff
@@ -468,6 +425,95 @@ class AllCourses extends mixinBehaviors([
...
+					request = changeFilter(roleFilters[i], Actions.enrollments.roleFilters.addFilter);
+				}
+			} else if (!isSelected && roleFilters[i].hasActionByName(Actions.enrollments.roleFilters.removeFilter)) {
+				if (request) {
+					request = request.then((newFiltersEntity) => changeFilter(newFiltersEntity.entities[i], Actions.enrollments.roleFilters.removeFilter));
+				} else {
+					request = changeFilter(roleFilters[i], Actions.enrollments.roleFilters.removeFilter);
+				}
+			}
+		}
```
AllanKerr, line 480: Can we initialize request to `Promise.resolve(this._roleFiltersEntity.entities)`? Then we can get rid of all the `undefined` request checks.

```
let request = Promise.resolve(this._roleFiltersEntity.entities);
const roleFilters = this._roleFiltersEntity.entities;
for (let i = 0; i < roleFilters.length; i++) {
    const isSelected = selectedRoles.find(role => role === roleFilters[i].title);
    if (isSelected && roleFilters[i].hasActionByName(Actions.enrollments.roleFilters.addFilter)) {
        request = request.then((newFiltersEntity) => changeFilter(newFiltersEntity.entities[i], Actions.enrollments.roleFilters.addFilter));
    } else if (!isSelected && roleFilters[i].hasActionByName(Actions.enrollments.roleFilters.removeFilter)) {
        request = request.then((newFiltersEntity) => changeFilter(newFiltersEntity.entities[i], Actions.enrollments.roleFilters.removeFilter));
    }
}
```

Also, is there any reason we're avoiding `async` / `await`? I think we could simplify this block even further.

src/d2l-all-courses.js
svanherk: Ok, pushed a change for this.  Adds more checking but I think makes more logical sense?

src/d2l-all-courses.js
svanherk: You're right, this is way cleaner with async/await.  Will push that shortly

src/d2l-all-courses.js
```diff
@@ -573,7 +678,7 @@ class AllCourses extends mixinBehaviors([
 		const bustCacheStr = 'bustCache=';
 		let index = url.indexOf(bustCacheStr);
 		if (index === -1) {
-			return `${url}${(url.indexOf('?') !== -1 ? '&' : '?')}bustCache=${this._bustCacheToken}`;
+			return `${url}${(url.indexOf('?') !== -1 ? '&' : '?')}bustCache=${this._bustCacheToken || ''}`;
```
svanherk, line 681: I broke this when updating the linting.  It could still be smarter (don't add the `bustCache` parameter at all if it's undefined) but for now just putting it back to how it was.

src/d2l-all-courses.js
```diff
@@ -489,11 +523,18 @@ class AllCourses extends mixinBehaviors([
...
 				this._enrollmentsSearchAction.getFieldByName('promotePins').value = this._sortMap[0].promotePins;
 			}
+			// Only clear if My Courses is not grouped by semesters/departments
+			if (this.tabSearchType !== 'BySemester' && this.tabSearchType !== 'ByDepartment' && this._enrollmentsSearchAction.hasFieldByName('parentOrganizations')) {
+				this._enrollmentsSearchAction.getFieldByName('parentOrganizations').value = '';
+			}
+			// Only clear if My Courses is not grouped by role
+			if (this.tabSearchType !== 'ByRoleAlias' && this._enrollmentsSearchAction.hasFieldByName('roles')) {
+				this._enrollmentsSearchAction.getFieldByName('roles').value = '';
+			}
```
AllanKerr, line 533: This seems to be duplicated, we merge those two uses together into something like `_clearFilterParams`?

src/d2l-all-courses.js
```diff
@@ -468,6 +425,83 @@ class AllCourses extends mixinBehaviors([
...
+		if (this.tabSearchType !== 'BySemester' && this.tabSearchType !== 'ByDepartment') {
+			params.parentOrganizations = '';
+		}
+		// Only clear if My Courses is not grouped by role
+		if (this.tabSearchType !== 'ByRoleAlias') {
+			params.roles = '';
+		}
+
+		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
+			createActionUrl(this._enrollmentsSearchAction, params)
```
AllanKerr, line 501: Should we also be clearing these params for `_enrollmentsSearchAction`?

src/d2l-all-courses.js
```diff
@@ -559,6 +600,70 @@ class AllCourses extends mixinBehaviors([
 
 		const searchAction = myEnrollmentsEntity.getActionByName(Actions.enrollments.searchMyEnrollments);
 		this._enrollmentsSearchAction = searchAction;
+
```
AllanKerr, line 603: Are we still okay with returning early if there isn't a `searchMyEnrollmentsAction` since it'll mean the role filters and filter categories don't get set up?

src/d2l-all-courses.js
svanherk: So it does, in a very roundabout way.  Setting the `_searchUrl` causes the search widget to do a search, triggering `_onSearchResultsChanged`.  Here, we set `_myEnrollmentsEntity` to the result, which triggers `_myEnrollmentsEntityChanged`, which sets `_enrollmentsSearchAction` to the new search action.

The reason we change it directly in `_onSimpleOverlayClosed` is because clearing filters, sort and search one after causes crazy race conditions.  Filter wasn't needed before because of the way it handled it's own changes, but now it acts like sort.

This data flow drives me crazy.  This is the thing I want to fix next, but we'll see if it makes the cut in the meeting this afternoon.

src/d2l-all-courses.js
svanherk: Yeah, if the `searchMyEnrollments` action is missing it means the overlay hasn't been opened yet and `myEnrollmentsEntity` is an empty siren object.  I don't want to run the role setup until we have a good enrollment entity

src/d2l-all-courses.js
```diff
@@ -559,6 +600,70 @@ class AllCourses extends mixinBehaviors([
...
+		const filterCategories = [];
+
+		// If My Courses is grouped by semesters/departments, don't show either of these tabs
+		if (this.tabSearchType !== 'BySemester' && this.tabSearchType !== 'ByDepartment' && searchSemestersAction) {
+			filterCategories.push({
+				key: 'semesters',
+				name: this.filterStandardSemesterName,
+				noOptionsText: this.localize('filtering.noSemesters', 'semester', this.filterStandardSemesterName),
+				filterAction: searchSemestersAction
+			});
```
AllanKerr, line 645: nit: I'd love to see some helpers or an entity around `FilterCategory`. It makes the code really difficult to read passing around objects that are accessed via raw properties without having any reference as to what the structure of the object actually is.

src/d2l-all-courses.js
svanherk: Yeah still trying to think or a way to do this... ideally this part shouldn't really exist like this.  I'd rather follow the pattern in other places and make a `params` array  and update `_searchUrl`, then apply the changes below. But that'll just cause the same race conditions.

I think ideally, I rip the fetching of the new enrollment state out of search and do it here.  And instead of setting a searchUrl, you call a function with a url that you can wait on (or not).

src/d2l-all-courses.js
svanherk: I like this a lot.  We talked about offline, will do this in a follow-up PR

src/d2l-all-courses.js
AllanKerr: It might not hurt to have a comment saying the `Actions.enrollments.searchMyEnrollments` check is actually to check if we have a valid Siren entity.

AllanKerr (approved):

PR #855, opened 2020-09-28
US120538 - Cleanup pre-pinned tab code

svanherk:
Now that the pinned tab flag is removed from code and we can always expect to be sent the pinned tab action, we can remove the code that used the " no tabs" path.  The tests required a lot of cleanup because they were mostly testing the old path ( where `_showGroupByTabs` was `false`).

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -327,7 +327,10 @@ class AllCourses extends mixinBehaviors([
 
 	// After a user-uploaded image is set, this is called to try to update the image
 	refreshCardGridImages(organization) {
-		this._getCardGrid().refreshCardGridImages(organization);
+		const cardGrid = this._getCardGrid();
+		if (cardGrid) {
+			cardGrid.refreshCardGridImages(organization);
+		}
 	}
 
 	_getCardGrid() {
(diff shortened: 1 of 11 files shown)


src/d2l-all-courses.js
```diff
@@ -311,7 +297,7 @@ class AllCourses extends mixinBehaviors([
 	*/
 
 	courseEnrollmentChanged(newValue) {
-		if (this._showGroupByTabs) {
+		if (this.tabSearchActions.length > 0) {
```
svanherk, line 300: This is checking now that the All Courses overlay has been opened and loaded (otherwise no point in attempting the below)

svanherk: It's necessary in that `this._selectedTabId.replace` will fail below because we don't have any tabs yet, but rather than check for that specifically we can cut it off earlier here (like was being done before).  I'll add a comment!

src/d2l-my-courses-container.js
```diff
@@ -251,9 +234,7 @@ class MyCoursesContainer extends MyCoursesLocalizeBehavior(PolymerElement) {
 	}
 
 	_getContentComponent() {
-		return this._showGroupByTabs === false || !this._currentTabId
```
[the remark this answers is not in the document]
svanherk: It _should_ always run first, especially since we use the content component when all courses is closed and when a user uploads a new course image.  But for all these cases, the workflow can't be critical if the tab hasn't been set to anything, so I'm thinking I'll just add checks to be safe

AllanKerr (approved):

PR #856, opened 2020-09-30
US120556 - Mobile Filter and Sort

svanherk:
Currently, at small screen sizes the filter and sort components are hidden.  Now that we are using the shared components which are more responsive, we're going to show them.
- Removed the media query that ides the filter and sort components
- Rearranged where the Advanced Search link was so that it stays underneath the search bar
- Added some more flex logic so things wrap nicely

my-courses-mobile-filter-sort

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -172,37 +172,36 @@ class AllCourses extends mixinBehaviors([
 					width: 100%;
 				}
 				#search-and-filter {
-					margin-bottom: 50px;
-				}
-				.search-and-filter-row {
 					display: flex;
+					flex-wrap: wrap;
 					justify-content: space-between;
+					margin-bottom: 1rem;
+					margin-top: -0.5rem;
 				}
 				.advanced-search-link {
 					font-size: 0.8rem;
 					margin-top: 3px;
-					flex: 1;
 				}
 				.advanced-search-link[hidden] {
 					display: none;
 				}
 				d2l-search-widget-custom {
+					min-width: 300px;
+					width: 100%;
+				}
+				#searchAndLink,
+				#filterAndSort {
+					display: flex;
+					margin-top: 0.5rem;
+				}
+				#searchAndLink {
 					flex: 1;
+					flex-direction: row;
+					flex-wrap: wrap;
 				}
 				#filterAndSort {
 					flex: 1.4;
-					display: flex;
 					justify-content: flex-end;
-					align-items: center;
-				}
-				@media screen and (max-width: 767px) {
-					#filterAndSort {
-						display: none;
-					}
-					.advanced-search-link {
-						text-align: right;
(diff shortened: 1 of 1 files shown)


src/d2l-all-courses.js
```diff
@@ -232,35 +231,32 @@ class AllCourses extends mixinBehaviors([
...
-									<d2l-sort-by-dropdown-option value="Default" text="[[localize('sorting.sortDefault')]]"></d2l-sort-by-dropdown-option>
-									<d2l-sort-by-dropdown-option value="OrgUnitName" text="[[localize('sorting.sortCourseName')]]"></d2l-sort-by-dropdown-option>
-									<d2l-sort-by-dropdown-option value="OrgUnitCode" text="[[localize('sorting.sortCourseCode')]]"></d2l-sort-by-dropdown-option>
-									<d2l-sort-by-dropdown-option value="PinDate" text="[[localize('sorting.sortDatePinned')]]"></d2l-sort-by-dropdown-option>
-									<d2l-sort-by-dropdown-option value="LastAccessed" text="[[localize('sorting.sortLastAccessed')]]"></d2l-sort-by-dropdown-option>
-									<d2l-sort-by-dropdown-option value="EnrollmentDate" text="[[localize('sorting.sortEnrollmentDate')]]"></d2l-sort-by-dropdown-option>
-								</d2l-sort-by-dropdown>
-
-							</div>
+							<d2l-link class="advanced-search-link" hidden$="[[!_showAdvancedSearchLink]]" href$="[[advancedSearchUrl]]">[[localize('advancedSearch')]]</d2l-link>
```
klhuff, line 242: Does this change the relative order of the Advanced search link and the filters? I'm assuming not because the filters appear on the right?

svanherk: No, you're right in that it WILL change it.  Someone tabbing through will now hit search bar -> advanced search link -> filter -> sort -> tabs below.  I'll confirm quickly with Jeff that that's ok, I never explicitly called that out to him

svanherk: Yup Jeff is good with this, but good call!

klhuff (approved):

src/d2l-all-courses.js
```diff
@@ -232,35 +231,32 @@ class AllCourses extends mixinBehaviors([
 					</iron-scroll-threshold>
 
 					<div id="search-and-filter">
-						<div class="search-and-filter-row">
+						<div id="searchAndLink">
```
AllanKerr, line 234: nit: `id`s should be kebab-case.

AllanKerr (approved):

src/d2l-all-courses.js
svanherk: Haha I was matching the syntax of the other flex container below - I'll make a note to update them both in a future PR

PR #857, opened 2020-10-05
US121196 - Move card grid to Lit

svanherk:
~Moving `d2l-my-courses-card-grid` to Lit and using `repeat` removed some rendering issues we were seeing with `dom-repeat`.~

Turns out there's a much simpler solution for this.  But we might as well still move this to Lit since the work is already done.

### DIFF
--- a/src/d2l-my-courses-card-grid.js
+++ b/src/d2l-my-courses-card-grid.js
@@ -1,90 +1,45 @@
 /*
 `d2l-my-courses-card-grid`
-Polymer-based web component for the grid of enrollment cards.
+Lit web component for the grid of enrollment cards.
 */
 
 import 'd2l-enrollments/components/d2l-enrollment-card/d2l-enrollment-card.js';
-import { html, PolymerElement } from '@polymer/polymer/polymer-element.js';
-import { afterNextRender } from '@polymer/polymer/lib/utils/render-status.js';
+import { css, html, LitElement } from 'lit-element';
 import { entityFactory } from 'siren-sdk/src/es6/EntityFactory.js';
 import { PresentationEntity } from 'siren-sdk/src/presentation/PresentationEntity.js';
+import { repeat } from 'lit-html/directives/repeat.js';
+import { styleMap } from 'lit-html/directives/style-map';
 
-class MyCoursesCardGrid extends PolymerElement {
-
-	static get is() { return 'd2l-my-courses-card-grid'; }
+class MyCoursesCardGrid extends LitElement {
 
 	static get properties() {
 		return {
 			// Array of courses to show
-			filteredEnrollments: Array,
+			filteredEnrollments: { type: Array },
 			// If true, will hide courses that are "closed" (only needed if a closed course was just unpinned,
 			// since we remove closed courses from the filteredEnrollments array on load)
-			hidePastCourses: {
-				type: Boolean,
-				value: false,
-				reflectToAttribute: true
-			},
-			// Token JWT Token for brightspace | a function that returns a JWT token for brightspace
-			token: String,
+			hidePastCourses: { attribute: 'hide-past-courses', reflect: true, type: Boolean },
 			// URL to fetch widget settings
-			presentationUrl: String,
+			presentationUrl: { type: String },
+			// Token JWT Token for brightspace | a function that returns a JWT token for brightspace
+			token: { type: Object },
 			// Limit the number of cards shown to 12 (unless more than 12 are pinned)
-			widgetView: {
-				type: Boolean,
-				value: false,
-				reflectToAttribute: true
-			},
-			/*
-			* Presentation Attributes
-			*/
-			_hideCourseStartDate: {
-				type: Boolean,
-				value: false
-			},
-			_hideCourseEndDate: {
-				type: Boolean,
-				value: false
-			},
-			_showOrganizationCode: {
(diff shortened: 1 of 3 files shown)


src/d2l-my-courses-card-grid.js
```diff
@@ -185,46 +182,11 @@ class MyCoursesCardGrid extends PolymerElement {
...
-		ie11retryCount = ie11retryCount || 0;
-		if (
-			ie11retryCount < 20
-			&& courseCardDivs.length === 0
-		) {
-			// If course cards haven't yet rendered, try again for up to one second
-			// (only happens sometimes, only in IE)
-			setTimeout(this.onResize.bind(this, ++ie11retryCount), 250);
-			return;
-		}
```
svanherk, line 217: From my testing in Edge 15 in Sauce, I wasn't apply to reproduce this issue, so I believe this was only needed for IE11.  I'm not super concerned if it does happen every once in a while for very old versions of Edge, which will be gone soon.

src/d2l-my-courses-card-grid.js
```diff
@@ -185,46 +182,11 @@ class MyCoursesCardGrid extends PolymerElement {
...
-
-			// The (* 2 - 1) accounts for the grid-gap-esque columns
-			const column = (position % numColumns + 1) * 2 - 1;
-			const row = Math.floor(position / numColumns) + 1;
-			div.style['-ms-grid-column'] = column;
-			div.style['-ms-grid-row'] = row;
-		}
+		// For old version of Legacy Edge
+		if (window.ShadyCSS) window.ShadyCSS.styleSubtree(this, { '--course-image-card-height': `${containerWidth / this._numColumns * 0.43}px` });
+		else this.style.setProperty('--course-image-card-height', `${containerWidth / this._numColumns * 0.43}px`);
```
svanherk, line 189: I wasn't sure if this should still live in the `onResize` or not.  It means `render` isn't really a function of the component's data, but would I start storing `_containerWidth` so a change always causes a re-render?  I think this will cause a lot more re-renders than before.

src/d2l-my-courses-content.js
```diff
@@ -305,18 +305,20 @@ class MyCoursesContent extends StatusMixin(MyCoursesLocalizeBehavior(PolymerElem
...
 		}
 
+		const cardGrid = this._getCardGrid();
 		const hide = this._hidePastCourses && (enrollmentCardStatusDetails.status.closed);
 		const index = this._enrollments.indexOf(enrollmentCardStatusDetails.enrollmentUrl);
 
 		if (hide && index !== -1 && index > this._lastPinnedIndex) {
-			this.splice('_enrollments', index, 1);
+			cardGrid.spliceEnrollments(index, 1);
+			this.notifyPath('_enrollments.length');
```
svanherk, line 314: @AllanKerr @dbatiste This is based on that slack convo we had a week or two ago.  Since there is an observer on `_enrollments.length`, I still need to let Polymer know that has changed.

src/d2l-my-courses-card-grid.js
AllanKerr: This seems like a reasonable place to set it. In other places we've been setting it in `updated`

src/d2l-my-courses-card-grid.js
```diff
@@ -129,35 +84,39 @@ class MyCoursesCardGrid extends PolymerElement {
...
+		if (changedProperties.has('presentationUrl')) {
+			this._onPresentationEntityChange();
+		}
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
-		afterNextRender(this, () => {
+		requestAnimationFrame(() => {
 			window.addEventListener('resize', this.onResize);
```
AllanKerr, line 120: It'd be nice to move this out of the `requestAnimationFrame` since it should be available immediately in the `connectedCallback`

AllanKerr (approved):

src/d2l-my-courses-card-grid.js
```diff
@@ -129,35 +84,39 @@ class MyCoursesCardGrid extends PolymerElement {
...
+
+		if (changedProperties.has('presentationUrl')) {
+			this._onPresentationEntityChange();
+		}
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
-		afterNextRender(this, () => {
+		requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
svanherk: Yeah I'm not concerned about the timing change, from what I can tell I think the purpose was just to delay non-urgent work.  But like you mentioned below, seems cleaner to just remove the `requestAnimationFrame` altogether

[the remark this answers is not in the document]
svanherk: Or I guess, move out the addEventListener... I think I'll keep the initial resize call delayed.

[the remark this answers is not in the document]
AllanKerr: The resize call will need it since `this.shadowRoot.querySelector` won't find anything without a delay.

src/d2l-my-courses-card-grid.js
dbatiste: Yeah I don't have any major objection here but perhaps `render` is the most appropriate place.  I know we didn't elsewhere though.

dbatiste (approved):

PR #858, opened 2020-10-08
[1] US117308 - Fix loading spinner in all courses

svanherk:
Looks like the "load more" spinner in the View All Courses overlay has been broken for a long long time.  It will work the first time we load more, but all the times after that it won't appear because the entity is in a different form and we need to dig in and get `_entity`.  Rearranged the logic below to do so, and cleanup some duplication at the same time.

This will be the first of a number of PRs focused on simplifying the data flow in All Courses.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -340,27 +340,15 @@ class AllCourses extends mixinBehaviors([
 	_onAllCoursesLowerThreshold() {
 		if (this.$['all-courses'].opened && this._lastEnrollmentsSearchResponse) {
 			const lastResponseEntity = this._lastEnrollmentsSearchResponse;
-			if (!lastResponseEntity._entity) {
-				if (lastResponseEntity && lastResponseEntity.hasLinkByRel('next')) {
-					const url = lastResponseEntity.getLinkByRel('next').href;
-					this.$.lazyLoadSpinner.scrollIntoView();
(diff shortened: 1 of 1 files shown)


src/d2l-all-courses.js
```diff
@@ -340,27 +340,15 @@ class AllCourses extends mixinBehaviors([
 	_onAllCoursesLowerThreshold() {
 		if (this.$['all-courses'].opened && this._lastEnrollmentsSearchResponse) {
 			const lastResponseEntity = this._lastEnrollmentsSearchResponse;
```
svanherk, line 342: We no longer have to check this because down below, we set `_lastEnrollmentsSearchResponse` to the entity we need.

src/d2l-all-courses.js
```diff
@@ -666,7 +654,6 @@ class AllCourses extends mixinBehaviors([
 			return false;
 		}
 
-		lastResponse = SirenParse(lastResponse);
```
svanherk, line 669: We don't need to parse this - we can expect a Siren entity.  And if we can't I want to know about it, because we should be able to.

svanherk: Although, I am wondering if I should do a null check on `lastResponse`.  It never should be `null`, because if `showTabContent` is true that means we _have_ content and have set the response.  But it doesn't hurt anything to do one...

AllanKerr (approved):

PR #860, opened 2020-10-09
[3] US117308 - Remove _myEnrollmentsEntity

svanherk:
The `_myEnrollmentsEntity` was used by the old filter component.  Now, instead of storing it, we can just set everything accordingly and skip the "observer" step.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -93,11 +93,6 @@ class AllCourses extends mixinBehaviors([
 			_isSearched: Boolean,
 			// Object containing the last response from an enrollments search request
 			_lastEnrollmentsSearchResponse: Object,
-			// Entity returned from my-enrollments Link from the enrollments root
-			_myEnrollmentsEntity: {
-				type: Object,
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.js
```diff
@@ -303,20 +298,17 @@ class AllCourses extends mixinBehaviors([
 		this._hasEnrollmentsChanged = true;
 	}
 
-	load() {
```
svanherk, line 306: `load` used to do a lot more, prior to turning on the pinned tab on permanently.  Now that it's only this, it doesn't need it's own public function (only tests were using it).

src/d2l-all-courses.js
```diff
@@ -365,16 +357,8 @@ class AllCourses extends mixinBehaviors([
 
 	_onSearchResultsChanged(e) {
```
svanherk, line 358: Soon, the search widget will not be in charge of data retrieval, and will work more like the sort and filter components.  This move sets things up better for that, while cleaning up the setting of `_myEnrollmentsEntity `.

AllanKerr (approved):

PR #863, opened 2020-10-14
US117308 - DE39401 - Stop using d2l-search-widget and instead let All Courses fetch enrollments data

svanherk:
Instead of having the data flow go through searching, and manually setting the `_searchUrl` of the search widget to trick it into doing a fetch, we're now going to let the All Courses overlay do the fetching.  It can then request data the same way the "load more" workflow does, and filter/sort/search all work the same.  Since all search is now responsible for is firing an event (and all the saved searches stuff), it can just use `d2l-input-search` from `core` directly.

The next PR is going to cleanup the `_searchUrl` part, to keep track of the current state of the parameters instead of passing around the full url.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -525,7 +525,7 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 		);
 	}
 
-	async _fetchEnrollments(url) {
+	_fetchEnrollments(url) {
 		entityFactory(EnrollmentCollectionEntity, url, this.token, entity => {
 			if (entity) {
 				this._handleNewEnrollmentsEntity(entity);
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -89,8 +89,11 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			_isSearched: Boolean,
 			// Object containing the last response from an enrollments fetch
 			_lastEnrollmentCollectionResponse: Object,
-			// URL passed to search widget, called for searching
-			_searchUrl: String,
+			// URL to fetch enrollments, set by filtering, sorting, searching, and selecting a tab
+			_searchUrl: {
+				type: String,
+				observer: '_fetchEnrollments'
+			},
 			_selectedTabId: String,
 			_showAdvancedSearchLink: {
 				type: Boolean,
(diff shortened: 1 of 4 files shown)


src/d2l-all-courses.js
```diff
@@ -506,7 +510,8 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			autoPinCourses: false,
 			sort: sortData.action,
 			embedDepth: 0,
-			promotePins: sortData.promotePins
+			promotePins: sortData.promotePins,
+			pageSize: 20
```
svanherk, line 514: This was being set by `search` before.  Now that we don't go through search, we can just set it here from the get-go.

src/d2l-all-courses.js
```diff
@@ -350,10 +350,14 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
...
 
 	_onSearchChange(e) {
-		this._isSearched = !!e.detail.searchValue;
+		this._isSearched = !!e.detail.value;
 
-		const enrollmentsEntity = new EnrollmentCollectionEntity(e.detail.searchResponse, this.token);
-		this._handleNewEnrollmentsEntity(enrollmentsEntity);
+		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
+			createActionUrl(this._enrollmentsSearchAction, {
+				orgUnitTypeId: this.orgUnitTypeIds,
```
svanherk, line 357: I don't think we need to keep setting this in search, filter, and sort.  It shouldn't ever change, so I'll investigate that as part of the next step.

src/search-filter/d2l-my-courses-search.js
```diff
@@ -119,46 +96,29 @@ class MyCoursesSearch extends MyCoursesLocalizeBehavior(PolymerElement) {
...
-	search() {
-		this._getSearchWidget().search();
-	}
+
 	clear() {
-		this._getSearchWidget().clear();
-	}
-	_computeSearchQuery(orgUnitTypeIds) {
-		return {
-			page: 1,
```
svanherk, line 130: From what I can tell, `page` was doing nothing.  Since it didn't have a corresponding `field` it was never even added to the url (https://github.com/Brightspace/d2l-my-courses-ui/blob/master/src/d2l-utility-helpers.js#L18), and the "load more" workflow uses `bookmark`.

src/search-filter/d2l-my-courses-search.js
```diff
@@ -69,29 +52,25 @@ class MyCoursesSearch extends MyCoursesLocalizeBehavior(PolymerElement) {
...
-					on-blur="_onSearchInputBlur">
-				</d2l-search-widget>
+					on-keydown="_onSearchInputKeyPressed"
+					label="[[localize('search.searchCourses')]]"
+					placeholder="[[localize('courseSearchInputPlaceholder')]]">
+				</d2l-input-search>
 			</div>
-			<d2l-dropdown-content id="dropdown" min-width="[[_dropdownWidth]]" max-width="[[_dropdownWidth]]" no-pointer no-auto-close no-auto-focus no-padding>
-
-				<div class="dropdown-content">
```
AllanKerr, line 85: Is there some context for why this was removed? Was it not doing anything?

svanherk: Yeah from what I can tell it was doing nothing.  It was applying the transparent background which didn't seem to do anything toggling it on and off.  I retested the saved searches behaviour and it looks the same as before

AllanKerr (approved):

PR #864, opened 2020-10-15
US117308 - DE39401 - Store Action Params

svanherk:
To make things simpler when switching tabs, and to cleanup the logic to remove the race conditions when closing the overlay, we can store the current state of the params and use those each time we fetch enrollments.  For now I've kept the rest of the fetching workflow in place, but an upcoming PR will wrap common code in a function and call that instead of using an observer.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -35,7 +35,10 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			// Standard Semester OU Type name to be displayed in the filter dropdown
 			filterStandardSemesterName: String,
 			// Configuration value passed in to toggle Learning Paths code
-			orgUnitTypeIds: Array,
+			orgUnitTypeIds: {
+				type: Array,
+				observer: '_onOrgUnitTypeIdsChange'
+			},
 			// URL to fetch widget settings
 			presentationUrl: String,
 			// Set by the image selector when it experiences an error trying to set a new course image
@@ -278,8 +281,8 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			</d2l-simple-overlay>`;
 	}
 
-	connectedCallback() {
-		super.connectedCallback();
+	ready() {
+		super.ready();
 		this._actionParams = {
 			autoPinCourses: false,
 			embedDepth: 0,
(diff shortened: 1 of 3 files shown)


src/d2l-all-courses.js
```diff
@@ -495,33 +496,10 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			return;
 		}
```
svanherk, line 498: We now no longer have to do anything when switching tabs, since we've been keeping track and updating `_actionParams` whenever we sort/filter/search.

src/d2l-my-courses-container.js
```diff
@@ -172,14 +174,14 @@ class MyCoursesContainer extends MyCoursesLocalizeBehavior(PolymerElement) {
...
-		let ouTypeIds = []; //default value
+		let ouTypeIds = []; // default value
 		try {
 			ouTypeIds = JSON.parse(this.orgUnitTypeIds).value;
 		} catch (e) {
 			// default value used
 		}
 
-		this.orgUnitTypeIds = ouTypeIds;
+		this._orgUnitTypeIds = ouTypeIds;
```
svanherk, line 184: The changes in this file are to cleanup the format of `orgUnitTypeIds` _before_ `d2l-my-courses-content` and `d2l-all-courses` ever get it.  They are expecting an array, and at first they were receiving `{ type = "Array", value = m_orgUnitTypeIds }` as a string (https://search.d2l.dev/xref/lms/le/manageCourses/D2L.LE.ManageCourses/Web/Desktop/Views/Widget/MyCoursesWidgetFactory.cs?r=31de5c0b#299-302).

src/d2l-all-courses.js
```diff
@@ -276,6 +278,19 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			</d2l-simple-overlay>`;
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		this._actionParams = {
+			autoPinCourses: false,
+			embedDepth: 0,
+			orgUnitTypeId: this.orgUnitTypeIds,
```
svanherk, line 286: Wondering if I should add an observer for this just in case... from what I can tell, it _should_ be ready on time, and I'm confident it is never supposed to change after the fact because of that reformat in the `connectedCallback` of `d2l-my-courses-container` that would only happen once.

AllanKerr: I think this should probably be in `ready` instead. Polymer docs say:
> Called after property values are set and local DOM is initialized.
Use for one-time configuration of your component after local DOM is initialized. (For configuration based on property values, it may be preferable to use an observer.)

AllanKerr (approved):

PR #866, opened 2020-10-15
US117308 - Remove _searchUrl and cleanup duplicate code

svanherk:
Trying to combine duplicate code and create a data flow that's more intuitive.  The url construction is now done once and the url isn't saved

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -91,11 +91,6 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			_isSearched: Boolean,
 			// Object containing the last response from an enrollments fetch
 			_lastEnrollmentCollectionResponse: Object,
-			// URL to fetch enrollments, set by filtering, sorting, searching, and selecting a tab
-			_searchUrl: {
-				type: String,
-				observer: '_fetchEnrollments'
-			},
 			_selectedTabId: String,
 			_showAdvancedSearchLink: {
 				type: Boolean,
@@ -294,8 +289,8 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 		if (this.tabSearchActions.length > 0) {
 			this._bustCacheToken = Math.random();
 			const actionName = this._selectedTabId.replace('all-courses-tab-', '');
-			if (!newValue.isPinned && actionName === Actions.enrollments.searchMyPinnedEnrollments && this._searchUrl) {
-				this._searchUrl = this._appendOrUpdateBustCacheQueryString(this._searchUrl);
+			if (!newValue.isPinned && actionName === Actions.enrollments.searchMyPinnedEnrollments && this._enrollmentsSearchAction) {
+				this._updateEnrollments(this._enrollmentsSearchAction);
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


src/d2l-all-courses.js
```diff
@@ -491,10 +476,14 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 		}
 
 		this._showTabContent = false;
+		this._updateEnrollments(tabAction.enrollmentsSearchAction);
+	}
 
-		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
-			createActionUrl(tabAction.enrollmentsSearchAction, this._actionParams)
+	async _updateEnrollments(searchAction) {
```
svanherk, line 482: Will remove this `async` and the one below as part of fixing merge conflicts

src/d2l-all-courses.js
```diff
@@ -352,18 +347,14 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 		this._actionParams.sort = sortData.action;
 		this._actionParams.promotePins = sortData.promotePins;
 
-		this._searchUrl = this._appendOrUpdateBustCacheQueryString(
-			createActionUrl(this._enrollmentsSearchAction, this._actionParams)
-		);
+		this._updateEnrollments(this._enrollmentsSearchAction);
```
AllanKerr, line 350: What are your thoughts on making `_actionParams` immutable and copying / reassigning it whenever anything changes?
```
this._actionParams = {...{ sort: sortData.actions, promotePins: sortData.promotePins}, ...this._actionParams}
```
Then we could use an observer on `_actionParams` to call `_updateEnrollments`?

Its a bit of a hyrbid between how it was done before and how its done now. My main concern with this approach is that it relies on manually keeping the enrollments in sync with `_actionParams` when they should never diverge. I also like the idea of keeping `_actionParams` immutable and coping it rather than modifying it directly.

svanherk: Yeah I considered that, but there's a few places where we need to _`updateEnrollments` without changing any `_actionParams`... I didn't know if it was weird to call it manually in those cases but have an observer for the other cases.  But might be worth it to keep the enrollments in sync, like you said.  Thoughts?

AllanKerr: I do think it'd be more maintainable to automatically keep it in sync and have a few manual calls in the places that don't change the params.

src/d2l-all-courses.js
```diff
@@ -491,13 +476,17 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 		}
 
 		this._showTabContent = false;
+		this._updateEnrollments(tabAction.enrollmentsSearchAction);
```
AllanKerr, line 479: Should we be assigning `this._enrollmentsSearchAction` here? I know its assigned indirectly via `_updateEnrollments` but it means theres a brief period where it might be `null` or reference different tab depending on how long the enrollments take to fetch.

svanherk:
After chatting offline, the possible readability and code organization benefits here don't seem to outweigh the risk of introducing a different timing structure. Closing PR!

PR #867, opened 2020-10-15
US117308 - Rename _searchUrl and trim down duplicate code

svanherk:
This is an alternative to https://github.com/Brightspace/d2l-my-courses-ui/pull/866.  It keeps the observer on the `_searchUrl`, and just renames it and creates a helper function for constructing the url


svanherk:
After chatting offline, the possible readability and code organization benefits here don't seem to outweigh the risk of introducing a different timing structure. Closing PR!

PR #870, opened 2020-10-16
DE409652 - Stop listening for tab events on the body

svanherk:
The event listener for when a new tab is selected is connected at the `body` level so the `d2l-my-courses-content` components can get them (as they are _inside_ the tabs).  This is fragile though, causing weird syncing with other my courses widgets on the homepage and unknown effects if another component with tabs is added.  Instead, we'll let the container catch the and let the content components know.


src/d2l-my-courses-container.js
```diff
@@ -381,15 +377,22 @@ class MyCoursesContainer extends MyCoursesLocalizeBehavior(PolymerElement) {
 		}
 	}
 	_tabSelectedChanged(e) {
-		this._currentTabId = `panel-${e.detail.tabId}`;
+		e.stopPropagation();
+		this._currentTabId = e.composedPath()[0].id;
+		const actionName = this._currentTabId.substring(6); // remove "panel-"
```
AllanKerr, line 382: This feels a little fragile. What are your thoughts on passing the `id` up to `newTabSelected` and returning the `actionName` rather than trying to build it ourselves?

svanherk: Sure, I'll push that change and we'll see how it feels!

svanherk: Ok, pushed.  Is that what you had in mind?

AllanKerr: That's exactly what I had in mind!

src/d2l-my-courses-container.js
```diff
@@ -381,15 +377,25 @@ class MyCoursesContainer extends MyCoursesLocalizeBehavior(PolymerElement) {
...
 	}
 	_tabSelectedChanged(e) {
-		this._currentTabId = `panel-${e.detail.tabId}`;
+		e.stopPropagation();
+		this._currentTabId = e.composedPath()[0].id;
+
+		let actionName;
+		const contents = this.shadowRoot.querySelectorAll('d2l-my-courses-content');
+		contents.forEach(content => {
+			const name = content.newTabSelected(e.composedPath()[0].id);
```
AllanKerr, line 386: nit: reuse `this._currentTabId`

src/d2l-my-courses-container.js
```diff
@@ -381,15 +377,25 @@ class MyCoursesContainer extends MyCoursesLocalizeBehavior(PolymerElement) {
...
+			if (name) {
+				actionName = name;
```
AllanKerr, line 388: nit: we can break out of the loop as soon as we get an action back.

svanherk: I need to tell all the contents about the tab change though, so they can set their `_thisTabSelected` property to `false`.  I think we could design this flow to be nicer and have the container set a public property and handle that change on this side, but I'm out of time lol

AllanKerr (approved):

PR #876, opened 2020-12-16
Migrating from Travis CI to GitHub Actions

dlockhart:
The major change here is the switch to use semantic-release for releasing new versions -- @klhuff let me know if you have concerns here that aren't explained by the README changes. We do have an incremental release action that's more similar to how this used to work (i.e. `[increment patch]`), but we prefer semantic-release for its other nice properties.


wct.conf.sauce.json
```diff
@@ -5,13 +5,18 @@
       "idleTimeout": 360,
       "browsers": [
```
dlockhart, line 6: I merged these configs into a single file, as I couldn't think of an advantage to having them separate and didn't remember the original rationale. Previously, the tests would have run sequentially, whereas now Sauce will run them in parallel.

wct.conf.sauce.json
```diff
@@ -5,13 +5,18 @@
...
         },
         {
           "browserName": "firefox",
-          "platform": "Windows 10",
+          "platform": "OS X 10.15",
+          "version": ""
+        },
+        {
+          "browserName": "safari",
+          "platform": "OS X 10.13",
```
dlockhart, line 19: I had to downgrade this to `10.13` as a test was failing on `10.15` -- likely something to investigate outside of this PR.

wct.conf.sauce.json
svanherk: So I _think_ at one point the tests were so flakey and slow that we would randomly pick a config to run.  Or maybe we were _going_ to do that.  I can't remember, but I agree there's no need for them to be separate.

svanherk (approved):
Looks good to me!  I agree we should move this to `semantic-release`.  It'll be less confusing to keep the experience the same as all the component repos it uses (`core`, `organizations`, `enrollments`, `facet-filter-sort`, etc.)

PR #892, opened 2021-05-21
Replace simple-overlay with dialog-fullscreen when showing all courses

annabelsegalld2l:
Updating the show all courses widget to use the d2l-dialog-fullscreen component rather than d2l-simple-overlay.

This is part of a set of PRs all needed to make the image-selector component work with d2l-dialog-fullscreen: 

- https://github.com/Brightspace/d2l-my-courses-ui/pull/892
- https://github.com/Brightspace/d2l-image-banner-overlay/pull/112
- https://github.com/Brightspace/image-selector/pull/48

This can only be merged after the changes in the image-selector which expose public functions.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -5,13 +5,13 @@ Polymer-based web component for the all courses overlay.
 
 import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
 import '@brightspace-ui/core/components/alert/alert.js';
+import '@brightspace-ui/core/components/dialog/dialog-fullscreen.js';
 import '@brightspace-ui/core/components/link/link.js';
 import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
 import '@brightspace-ui/core/components/tabs/tabs.js';
 import '@brightspace-ui/core/components/tabs/tab-panel.js';
 import '@brightspace-ui-labs/facet-filter-sort/components/sort-by-dropdown/sort-by-dropdown.js';
 import '@brightspace-ui-labs/facet-filter-sort/components/sort-by-dropdown/sort-by-dropdown-option.js';
-import 'd2l-simple-overlay/d2l-simple-overlay.js';
 import './d2l-my-courses-card-grid.js';
 import './search-filter/d2l-my-courses-search.js';
 import { createActionUrl, fetchSirenEntity } from './d2l-utility-helpers.js';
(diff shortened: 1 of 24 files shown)


src/d2l-all-courses.js
```diff
@@ -208,14 +208,10 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 				}
 			</style>
 
-			<d2l-simple-overlay
+			<d2l-dialog-fullscreen
 				id="all-courses"
-				on-d2l-simple-overlay-opening="_onSimpleOverlayOpening"
-				on-d2l-simple-overlay-closed="_onSimpleOverlayClosed"
-				close-simple-overlay-alt-text="[[localize('closeSimpleOverlayAltText')]]"
```
annabelsegalld2l, line 215: The dialog component has a built-in alt text for the close button, which says "Close this dialog" rather than "Close dialog"

dbatiste:
Is the plan to switch the other usages within my-courses over in a separate PR?

annabelsegalld2l:
> Is the plan to switch the other usages within my-courses over in a separate PR?

There is one more usage of the overlay in my-courses, but it looks like I need to replace both of them at the same time (since one spawns the other).

src/d2l-all-courses.js
```diff
@@ -208,14 +208,10 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
...
 				restore-focus-on-close
```
svanherk, line 213: Doesn't look like dialog has this property but I expect it will do it by default - after closing the dialog (either by picking and image or hitting enter on the close button) does focus return to the proper menu item?

src/d2l-all-courses.js
svanherk: You can cleanup the langterm here as well - the legacy code has its own copy of the langterms in that folder so we don't need to worry about it still needing it.

src/d2l-all-courses.js
```diff
@@ -282,6 +278,8 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
 			search: '',
 			sort: this._sortMap.Default.action
 		};
+		this.addEventListener('d2l-dialog-open', this._onSimpleOverlayOpening);
+		this.addEventListener('d2l-dialog-close', this._onSimpleOverlayClosed);
```
svanherk, line 282: I think you should be able to keep these defined right on the component using:
```
<d2l-dialog-fullscreen
...
     on-d2l-dialog-open="_onDialogOpening"
     on-d2l-dialog-close="_onDialogClosed"
...

```
This is Polymer's version of Lit's `@event-name=""` syntax, and I _think_ it should still work on a Lit component inside a Polymer one.

Also, I'd rename the functions to remove "SimpleOverlay" to cutdown on future confusion.

src/d2l-all-courses.js
annabelsegalld2l: It seems to be working without the restore-focus tag, so I can remove it.

svanherk (approved):

PR #894, opened 2021-06-16
fix: Add IntersectionObserver to detect bottom of scroll instead of iron-scroll-threshold

annabelsegalld2l:
The change from simple-overlay to d2l-dialog-fullscreen means that iron-scroll-threshold does not work, as the on-lower-threshold event does not get fired. We can replace it with an IntersectionObserver that detects when we scroll to the bottom of the course images, using a fake "scrollThreshold" div.

### DIFF
--- a/src/d2l-all-courses.js
+++ b/src/d2l-all-courses.js
@@ -3,7 +3,6 @@
 Polymer-based web component for the all courses overlay.
 */
 
-import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
 import '@brightspace-ui/core/components/alert/alert.js';
 import '@brightspace-ui/core/components/dialog/dialog-fullscreen.js';
 import '@brightspace-ui/core/components/link/link.js';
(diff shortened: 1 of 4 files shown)


annabelsegalld2l:
The CI seems to be broken on the lint step... I ran it locally and it worked. Doesn't look like any recent changes would have caused it to break. Any thoughts?

src/d2l-all-courses.js
```diff
@@ -3,7 +3,6 @@
 Polymer-based web component for the all courses overlay.
 */
 
-import '@polymer/iron-scroll-threshold/iron-scroll-threshold.js';
```
annabelsegalld2l, line 6: We can't delete the scroll threshold from our imports in package.json because it is still being used by the legacy myCourses view.

svanherk:
> The CI seems to be broken on the lint step... I ran it locally and it worked. Doesn't look like any recent changes would have caused it to break. Any thoughts?

Maybe try bumping to Node 12 or 14?  I think we needed to do that in `core` for linting issues, but not sure if it's the same error.

annabelsegalld2l:
> > The CI seems to be broken on the lint step... I ran it locally and it worked. Doesn't look like any recent changes would have caused it to break. Any thoughts?
> 
> Maybe try bumping to Node 12 or 14? I think we needed to do that in `core` for linting issues, but not sure if it's the same error.

That worked!

dlockhart (approved):

dbatiste (approved):

PR #909, opened 2021-12-13
feat: US134551 - Lit 2 Upgrade - backwards-compatible changes

annabelsegalld2l:
This PR makes backwards-compatible changes to prepare all repos for a Lit 2 upgrade. Since the changes are backwards compatible, it can be merged right away.

**Non-breaking change checklist:** 

- [x] Add missing `.js` extensions for lit-html or lit-element imports
- [x] `await requestUpdate` -> `await elem.updateComplete`
- [x] add non-null checks to shadowRoot accesses (not including test files)
- [x] ensure `undefined` is not passed into unsafeHTML calls
- [x] remove dynamic code from inside  `<template>` tags (Lit files only)
- [x] remove any dependencies on old Lit versions
- [x] audit `ifDefined` usages for null parameters
  - Usages audited - actually makes things better 🎉 

For more information on the Lit 2 upgrade, visit https://lit.dev/docs/releases/upgrade/ or contact team Gaudi.


svanherk:
Other than `d2l-my-courses-card-grid.js`, the other files here are actually still Polymer components.  I don't expect we need to add the `shadowRoot` checks to them?  I know it doesn't hurt anything, but wonder if we should remove them just to make an eventual migration cleaner

annabelsegalld2l:
> Other than `d2l-my-courses-card-grid.js`, the other files here are actually still Polymer components. I don't expect we need to add the `shadowRoot` checks to them? I know it doesn't hurt anything, but wonder if we should remove them just to make an eventual migration cleaner

Yes Stacey, we can definitely remove them. That's totally my bad, I didn't even notice.

src/d2l-my-courses-card-grid.js
```diff
@@ -197,12 +197,11 @@ class MyCoursesCardGrid extends LitElement {
 
 		this._numColumns = Math.min(Math.floor(containerWidth / 350), 4) + 1;
 
-		// For old version of Legacy Edge
```
svanherk, line 200: 🎉

svanherk (approved):

PR #916, opened 2022-01-18
Migrate to es-dev-server and @web/test-runner

svanherk:
Instead of sifting through the 100s of console errors to find the ones that Safari was actually upset with, I decided (for better or worse) to migrate to the latest test framework, removing all the errors as I went.

The only remaining errors are due to `dialog` and `tabs` trying to do things after they've been disconnected from the tests, and these only appear occasionally on a local Firefox.  They don't cause test failures, but are annoying.  I think they may go away on their own if we migrated everything from `Polymer` to `Lit`, or we can add some code to `dialog` and `tabs` themselves to handle missing pieces.


svanherk:
This PR doesn't cleanup useless tests, it just migrates them all and adjusts ones that were flaky/broken.  Might update linting in a follow-up PR if it's simple

package.json
```diff
@@ -12,10 +13,13 @@
...
   "name": "d2l-my-courses",
   "scripts": {
-    "lint": "npm run lint:wc && npm run lint:js",
-    "lint:js": "eslint . --ext .js,.html src/**/*.js test/**/*.js test/**/*.html demo/**/*.html",
-    "lint:wc": "polymer lint",
-    "test": "npm run lint && polymer test --config-file wct.conf.local.json"
+    "lint": "npm run lint:eslint && npm run lint:polymer",
+    "lint:eslint": "eslint . --ext .js,.html src/**/*.js test/**/*.js demo/**/*.html",
+    "lint:polymer": "polymer lint",
+    "start": "es-dev-server --app-index demo/index.html --node-resolve --dedupe --open --watch",
```
dlockhart, line 19: I think this should be `web-dev-server` -- maybe they kept the old one around for backwards-compat.

test/d2l-my-courses-filter.test.js
```diff
@@ -1,27 +1,19 @@
-import { MyCoursesFilterCategory } from '../../src/search-filter/d2l-my-courses-filter.js';
+//import '../src/search-filter/d2l-my-courses-filter.js';
```
dlockhart, line 1: Commented out line

package.json
svanherk: Oh whoops!  Wonder where I stole that from...

dlockhart (approved):

PR #923, opened 2022-03-09
Remove polymer linting

svanherk:
I already moved this repo to the web server and test runner, but left Polymer linting.  But we broke that with a recent `core` change and are trying to remove `polymer-cli` everywhere, so removing it now.


dlockhart (approved):
💯

PR #929, opened 2022-03-21
fix: removing forced attribute binding

dlockhart:
I've addressed the [underlying core issue](https://github.com/BrightspaceUI/core/pull/2244), so this temporary hack can be removed.


svanherk (approved):
Thanks!

PR #937, opened 2022-05-03
feat: US134779 - Use new core filter

svanherk:
This switches the filter in My Courses from the labs component to the new `d2l-filter` from `BrightspaceUI/core`:
image


src/d2l-all-courses.js
```diff
@@ -383,28 +383,37 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
...
+			semesters: selectedSemesters.length,
+			roles: selectedRoles.length
 		};
 
-		if (roleFilters && e.detail.categoryChanged === 'roles') {
-			this._handleRolesFilterChange(roleFilters.selectedOptions);
-		} else if ((semesterFilters && e.detail.categoryChanged === 'semesters') || (departmentFilters && e.detail.categoryChanged === 'departments')) {
-			const selectedSemesters = semesterFilters ? semesterFilters.selectedOptions : [];
-			const selectedDepartments = departmentFilters ? departmentFilters.selectedOptions : [];
+		// If semester and/or department values were changed, load those first and then apply the role filters
```
svanherk, line 397: From what I can tell this was an existing bug/race condition where a user selecting a semester and role filter in quick succession will cause one or the other to undo.  This is _a tad_ more evident now that we batch change events, but it would still be very difficult to have happen in practice because a user cannot switch filters that quickly.  I found it artificially by turning off the special "clear all" functionality to test it.

But in case it ever does happen, forcing the semester/department filters to load and update the `_roleFiltersEntity` first should help.

src/d2l-all-courses.js
```diff
@@ -522,17 +531,15 @@ class AllCourses extends MyCoursesLocalizeBehavior(PolymerElement) {
...
 
 		this._enrollmentsSearchAction = enrollmentsEntity.getSearchEnrollmentsActions();
 
 		const entity = enrollmentsEntity._entity;
 		if (entity.hasActionByName(Actions.enrollments.setRoleFilters)) {
 			const href = createActionUrl(entity.getActionByName(Actions.enrollments.setRoleFilters));
-			fetchSirenEntity(href).then(entity => {
-				this._roleFiltersEntity = entity;
-			});
+			this._roleFiltersEntity = await fetchSirenEntity(href);
```
svanherk, line 542: This is probably the riskiest part of the PR, making this wait on getting the `_roleFiltersEntity` back, but I need to know this has happened to know when it's safe to call `_handleRolesFilterChange`.  This call should be quick, but a general regression test should be done

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -3,47 +3,30 @@
 Lit web component for the my courses filter.
 */
 
-import '@brightspace-ui/core/components/loading-spinner/loading-spinner.js';
-import '@brightspace-ui-labs/facet-filter-sort/components/filter-dropdown/filter-dropdown.js';
```
svanherk, line 7: I can't remove `@brightspace-ui-labs/facet-filter-sort` from `package.json` yet because this still uses the labs sort dropdown.

grant-cleary (approved):

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -192,75 +162,60 @@ class MyCoursesFilter extends MyCoursesLocalizeMixin(LitElement) {
 		this.requestUpdate();
 	}
 
-	_onDropdownOpen() {
-		this._hasBeenOpened = true;
```
svanherk, line 196: Because the filter no longer uses tabs (which throws a selected event regardless of whether its containing dropdown has been opened), we don't need to keep track of whether the filter has been opened anymore.

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -192,75 +162,60 @@ class MyCoursesFilter extends MyCoursesLocalizeMixin(LitElement) {
...
 
 		this._fetchSirenEntity(searchActionHref).then((resultingEntity) => {
-			category.optionsEntity = resultingEntity;
-			// We don't need to check the filters class here, since if the filters class is present search is disabled
-			this._fetchSearchOptions(category);
+			let optionsToDisplay = [];
+			if (resultingEntity.entities) {
+				optionsToDisplay = resultingEntity.entities.map(option => option.href);
+			}
+			e.detail.searchCompleteCallback(optionsToDisplay);
```
svanherk, line 182: Instead of physically removing options (which messes with the filter counts), the search event provides a callback to pass in the options you want to hide.

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -192,75 +162,60 @@ class MyCoursesFilter extends MyCoursesLocalizeMixin(LitElement) {
...
-
-		this.dispatchEvent(new CustomEvent('d2l-my-courses-filter-clear'));
-	}
 
-	_onFilterDropdownOptionChange(e) {
-		const category = this.filterCategories.find(category => category.key === e.detail.categoryKey);
-		const optionIndex = category.selectedOptions.findIndex(optionKey => optionKey === e.detail.menuItemKey);
-
-		if (e.detail.selected) {
-			this._totalSelectedCount++;
```
svanherk, line 241: The new filter handles all the counts, we don't need to keep track of this and pass it in anymore

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -271,13 +226,12 @@ class MyCoursesFilter extends MyCoursesLocalizeMixin(LitElement) {
...
 export class MyCoursesFilterCategory {
-	constructor(key, name, noOptionsText, filterAction) {
+	constructor(key, name, filterAction, apiSearch) {
 		this.key = key;
 		this.name = name;
-		this.noOptionsText = noOptionsText;
 		this.filterAction = filterAction;
+		this.apiSearch = apiSearch;
 
-		this.isSearched = false;
```
svanherk, line 280: The filter handles disabling the search based on option count, we don't need to keep track of this anymore.

src/search-filter/d2l-my-courses-filter.js
```diff
@@ -271,13 +226,12 @@ class MyCoursesFilter extends MyCoursesLocalizeMixin(LitElement) {
 window.customElements.define('d2l-my-courses-filter', MyCoursesFilter);
 
 export class MyCoursesFilterCategory {
-	constructor(key, name, noOptionsText, filterAction) {
+	constructor(key, name, filterAction, apiSearch) {
 		this.key = key;
 		this.name = name;
-		this.noOptionsText = noOptionsText;
```
svanherk, line 277: The filter doesn't let you customize the message when a dimension has no options, so we can remove all those langterms.

test/d2l-all-courses.test.js
```diff
@@ -225,71 +222,42 @@ describe('d2l-all-courses', () => {
 					expect(widget._filterCounts.roles).to.equal(0);
 				});
 
-				[
-					{ categoryChanged: 'departments' },
-					{ categoryChanged: 'semesters' },
-					{ categoryChanged: 'roles' }
-				].forEach(testCase => {
-					it(`should not set _searchUrl if the categoryChanged is "${testCase.categoryChanged}" but its selected filters info is missing`, async() => {
```
svanherk, line 233: In theory this should never come up as an issue, and we don't guard against this anymore.

test/d2l-my-courses-filter.test.js
```diff
@@ -189,21 +190,11 @@ describe('d2l-my-courses-filter', () => {
 			expect(component.filterCategories[1].options).to.be.empty;
 		});
 
-		it('should show a loading spinner if the options are not loaded', async() => {
```
svanherk, line 192: Filter handles its own loading states

test/d2l-my-courses-filter.test.js
```diff
@@ -399,283 +352,204 @@ describe('d2l-my-courses-filter', () => {
...
 			}));
 
 			await timeout(0);
 			expect(component.filterCategories[0].selectedOptions.length).to.equal(1);
 			expect(component.filterCategories[0].selectedOptions[0]).to.equal('/semester/2');
-			expect(component.filterCategories[0].options.length).to.equal(1);
-			expect(component.filterCategories[0].options[0].key).to.equal('/semester/1');
-		});
-
-		it('should show the correct text and not disable the search input if all options are filtered out', async() => {
```
svanherk, line 424: Filter handles this

src/d2l-all-courses.js
klhuff: Right I guess the other risk plays a part here, since in the case where you've changed both, we'll also be waiting for the fetch request to finish.

klhuff (approved):

src/d2l-all-courses.js
svanherk: Yeah exactly.  Which is honestly very unlikely to happen because we handle "clear all" specially, and we batch every 200 ms.  So a user would need to select a filter, change to the role filter, and select one there in less than 200ms.  I wanted to fix the bug in case the filter ever increased that batching time, but if we're more concerned with the fix or see concerning regression testing results, we could "put this bug back".

